PR #94, opened 2015-04-28
Fix issue 91

mtoma:


sandoche2k:
@mtoma can you check and merge

PR #110, opened 2015-06-02
Idn problem

cdybedahl:


sandoche2k:
@mtoma can you check and merge?

PR #125, opened 2015-09-24
Fix issue 89

mtoma:


mtoma:
Fixes dotse/zonemaster-gui#89

PR #180, opened 2016-06-26
Fix issue 159

mtoma:


mtoma:
I merged both pull requests.

----- Original Message -----

PR #187, opened 2016-10-04
Fixes #185, Fixes #153

### LINKED ISSUE #153: Improve the batch API

### LINKED ISSUE #185: Document "add_api_user" and "add_batch_job" completely
Please update the API documentation to explain that "add_api_user" only works from "localhost".

Please also document how to use "add_batch_job".

mtoma:


matsduf:
The document is incomplete. It is not specified what responses are returned when add_job_result or get_batch_job_result fail.

mtoma:
Well, as for all the APIs it is more or less specified by the JSON-RPC protocol itself. A failure results in a standard JSON-RPC failure this is probably not optimal as it exposes the raw perl die message but this is all we have now.
If of course by "fail" you mean crash.

matsduf:
It does not have to crash to fail. If "add_job" fails because the user is not allowed to create the batch. If "get_batch_job_result" fails because the ID is invalid.

PR #189, opened 2016-10-04
s/cutom/custom/

aabdnn:
Fix typo

### DIFF
--- a/lib/Zonemaster/WebBackend/Config.pm
+++ b/lib/Zonemaster/WebBackend/Config.pm
@@ -142,7 +142,9 @@ sub force_hash_id_use_in_API_starting_from_id {
 sub CustomProfilesPath {
     my $cfg = _load_config();
 
-    return $cfg->val( 'ZONEMASTER', 'custom_profiles_path' );
+    my $value  = $cfg->val( 'ZONEMASTER', 'cutom_profiles_path' );
+    $value  = $cfg->val( 'ZONEMASTER', 'custom_profiles_path' ) unless ($value);
+    return $value;
 }
 
 sub GetCustomConfigParameter {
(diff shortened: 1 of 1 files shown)


mtoma:
This won't work.
Can you complete the fix like this please:
```
my $value  = $cfg->val( 'ZONEMASTER', 'cutom_profiles_path' );
$value  = $cfg->val( 'ZONEMASTER', 'custom_profiles_path' ) unless ($value);
return $value;
````

aabdnn:
I'm just submitting fixes for spelling errors. Are you trying to keep the code backwards compatible?

mtoma:
Yes I'm trying to make the code backward compatible with all parts that are not overridden by an upgrade. So the code needs to be backward compatible with the historic database contents and I'd say config files as well. We have at least one user of the backend I'm aware of using the config files with typos and I'd wouldn't like to relay on them fixing the typos in their config files in order to keep a working setup if they upgrade at some point.

aabdnn:
That's kind of sad. Zonemaster isn't so widely deployed, and it should be okay to make some incompatible changes early in its life, instead of carrying around baggage like this.

mtoma:
Yes, but it is only one more line of code per typo. I think it is worth the effort.

matsduf:
Even though Zonemaster is early in its deployment, it is important to have control over changes that are backward incompatible. At the same time, @aabdnn, you have a point that we should try to get rid of bad baggage. We should try to do this in a controlled upgrade where the incompatibility is pointed out.

aabdnn:
Alright, there you go. It looks really ugly to me, but I've made the change.

mtoma:
Yes, I agree Anand, it is really ugly. For now this seems to me as the only solution.
Thx for the fix.

PR #232, opened 2016-11-24
Various improvements

mattias-p:
* The installation instruction has been made easier to follow by introducing bullet lists where the linear flow is temporarily branched off into separate paths for different database engines or operating systems.
* Placeholder sections for SQLite has been added.
* Notes about unsupported stuff have been replaced by notes about missing instructions.
* Notes have been updated to use proper formatting.
* The fact that sudo is only needed for administrative tasks has been clarified.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,6 +21,7 @@ This installation guide assumes that the following softwares are already install
 
 * one of CentOS, Debian, FreeBSD or Ubuntu
 * sudo (only for installation and other administrative tasks)
+* cpanm (only for installation)
 * curl (only for post-installation sanity check)
 
 It also assumes that you've chosen (though not necessarily installed) one of the following database engines:
@@ -40,7 +41,7 @@ It also assumes that you've chosen (though not necessarily installed) one of the
 
 ```sh
 sudo yum install perl-Module-Install perl-IO-CaptureOutput perl-String-ShellQuote
-sudo cpan -i Config::IniFiles Daemon::Control JSON::RPC::Dispatch Parallel::ForkManager Plack::Builder Plack::Middleware::Debug Router::Simple::Declare Starman
+sudo cpanm -i Config::IniFiles Daemon::Control JSON::RPC::Dispatch Parallel::ForkManager Plack::Builder Plack::Middleware::Debug Router::Simple::Declare Starman
 ```
 
 Install the chosen database engine and related dependencies:
@@ -73,7 +74,7 @@ Install the chosen database engine and related dependencies:
 ```sh
 sudo apt-get update
 sudo apt-get install git libmodule-install-perl libconfig-inifiles-perl libdbd-sqlite3-perl starman libio-captureoutput-perl libproc-processtable-perl libstring-shellquote-perl librouter-simple-perl libclass-method-modifiers-perl libtext-microtemplate-perl libdaemon-control-perl 
-sudo cpan -i Plack::Middleware::Debug Parallel::ForkManager JSON::RPC
+sudo cpanm -i Plack::Middleware::Debug Parallel::ForkManager JSON::RPC
 ```
 
 Note: The Perl modules `Parallel::ForkManager` and `JSON::RPC` exist as Debian packages, but with versions too old to be useful for us.
@@ -128,37 +129,19 @@ Install the chosen database engine and related dependencies:
 
 ## Installation
 
-### Fetch the source code
-
 ```sh
-git clone https://github.com/dotse/zonemaster-backend.git
-cd zonemaster-backend
+sudo cpanm Zonemaster::WebBackend
 ```
-
-
-### Build source code
-
-```sh
-perl Makefile.PL
-make
-make test
-```
-
-These steps produce quite a bit of output. As long as it ends by
-printing `Result: PASS`, everything is OK.
-
-
-### Install 
-
-```sh
-sudo make install
-```
-
-This too produces some output. The `sudo` command may not be necessary,
-if you normally have write permissions to your Perl installation.
 
 
 ## Configuration
+
+The Zonemaster::WebBackend module installs a number of configuration files in a shared data directory.
+This section refers to the shared data directory as the current directory, so locate it and go there:
+
+```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-WebBackend")'`
+```
 
 Create directories for configuration and log files.
 
@@ -170,7 +153,7 @@ mkdir "$HOME/logs"
 Copy the `backend_config.ini` file to `/etc/zonemaster`.
 
 ```sh
-sudo cp share/backend_config.ini /etc/zonemaster/
+sudo cp ./backend_config.ini /etc/zonemaster/
 ```
 
 Install service scripts for the relevant operating system:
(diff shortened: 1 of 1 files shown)


mattias-p:
* A prompt about choosing a database engine has been added to the prerequisites section

mattias-p:
@mtoma Please don't just merge everything. Instead leave a comment with your approval. Even if you don't have any issues with something, maybe someone else does.

matsduf:
The instructions is based on installation of backend by cloning the repo and doing make install. We should use the CPAN installation method here so that they really get a release version.

```
Fetch the source code

git clone https://github.com/dotse/zonemaster-backend.git
cd zonemaster-backend

Build source code

perl Makefile.PL
make
make test

These steps produce quite a bit of output. As long as it ends by printing Result: PASS, everything is OK.
Install

sudo make install
```

mattias-p:
@matsduf Good point. I'll have to figur out a new way to refer to the `share` directory though.

sandoche2k:
@matsduf Good point. This should be the objective.

mattias-p:
CPAN itself recommends using `cpanm` over `cpan`. I'll update the guide to use `cpanm` at the same time as I update the "Installation" section.

mtoma:
The travis automatic test files use cpanm as well.

----- Original Message -----

mattias-p:
The CentOS instruction starts the database engine.

The FreeBSD instruction does not configure and start the database engine. I'll add instructions and notes about configuring and starting databases on FreeBSD.

How do Debian and Ubuntu behave in this respect? Does their instruction also need updating?

mattias-p:
I'm not sure database engine configuration and startup belong in this document. Maybe the the database engine should be a prerequisite instead? In that case the database engine configuration and startup instructions should be moved to the distrib preparation instructions. In any case that's out of scope for this PR - it deserves an issue of its own.

mattias-p:
@mtoma When you install this from CPAN, where do the sql files for setting up the database end up? I can't find them.

mtoma:
If you don't have the latest release they end up nowhere.
With the latest release they should end up in the share folder.

----- Original Message -----

mattias-p:
@mtoma I see, so 1.0.7-A is broken in this respect, but the problem has been fixed since then?

mtoma:
Yes, this was done after the release to help CZnic to package the backend.

----- Original Message -----

mattias-p:
I've tested the instruction for FreeBSD and PostgreSQL and fixed all issues I encountered. The PR has seen a decent amount of eyeballs. I think this should be ready to merge.

mtoma:
Ok, merging

----- Original Message -----

docs/installation.md
```diff
@@ -8,7 +8,6 @@ This document begins with a number of sections concerning native installation:
 * [Configuration](#configuration)
 * [Startup](#startup)
 * [Post-installation sanity check](#post-installation-sanity-check)
-* [What's next](#whats-next)
```
sandoche2k: I think the sub section "Whats next" is useful

docs/installation.md
```diff
@@ -35,49 +41,63 @@ This installation guide assumes that the following softwares are already install
```
sandoche2k: Add 
```
#### Installing dependencies on CentOS


```

docs/installation.md
```diff
@@ -21,9 +20,16 @@ This document ends with these appendices:
...
+* sudo (only for installation and other administrative tasks)
+* cpanm (only for installation)
 * curl (only for post-installation sanity check)
 
+It also assumes that you've chosen (though not necessarily installed) one of the following database engines:
+
+* MySQL
+* PostgreSQL
+* SQLite
+
```
sandoche2k: Add the following:
```
The process is as follows:
1. Install the dependencies for your OS
2. Choose the database and install its dependecies
3. Install the Zonemaster backend
4. Configure the backend to be used with its respective database
5. Start the backend
6. Verify that your backend has started correctly
```

docs/installation.md
matsduf: This is installation guide for the backend. The "What's next" section pointed at the installation of the GUI. I see no point to point at that. Under Zonemaster we have the following text that bind the different parts together. There is no point of extending the installation guide for the backend with other information.

> In order to install and run the different components, have a look at the installation documentation under each of the repositories below:
>
>    zonemaster-engine - which contains the test framework.
    zonemaster-cli - a Command Line Interface for the engine.
    zonemaster-backend - which interfaces between the GUI and an API with the engine.
    zonemaster-gui - A graphical user interface for the engine.

docs/installation.md
matsduf: Yes, that is a good checklist to have early in the document.

docs/installation.md
```diff
@@ -21,9 +20,16 @@ This document ends with these appendices:
...
 
 ## Dependencies
```
sandoche2k: Remove ## Dependencies and add the following :
```

## Instructions for CentOS

Comment (matsduf) [inline docs/installation.md]: Not only CentOS is covered in the Dependencies section.

**Comment by sandoche2k** [comment]
@mattias-p I think there is a modification needed in the structure of this document. It is cumbersome and time consming to comment one by one. I prefer you merge and then i do the modification.

**Comment by mattias-p** [comment]
@sandoche2k Please point out any structural issues and I will address them. If you have trouble expressing your concerns and would rather make a PR, then make that PR to `mattias-p:installation-tweaks`. It is terrible process to merge things into master halfway through review.

**Comment by sandoche2k** [comment]
@mattias :  As you know, I have already lost a day of work wherein all my
comments were lost in the review modifications. I prefer to work in the
same manner as we have done for the last two years. If there are short
comments i will provide directly to the document.

File: docs/installation.md (hunk 2 above)
Comment (mattias-p) [inline docs/installation.md]: Are you saying that the listed prerequisites should be marked as dependencies? I believe they're not. Can you elaborate?

**Comment by sandoche2k** [comment]
No. I wanted "Dependecies" section to be added just before #installation
instructions for dependencies.

Now, i think we will have email discussion on line numbers :-).

File: docs/installation.md (hunk 3 above)
Comment (mattias-p) [inline docs/installation.md]: The outline of the document differs slightly from your listing, so I can't simply add it without reorganizing the text.

What I believe you're asking for is a more concise explanation of the procedure. To that end I've reworked the beginning of the document. I had trouble describing the "Prerequisites" section. Turns out this section was complex so I broke it into two simple sections.

File: docs/installation.md [hunk 5]
```diff
@@ -1,32 +1,64 @@
...
+6. [Startup](#startup) shows you how to start the *Backend*.
+7. [Post-installation sanity check](#post-installation-sanity-check) lets you verify that your *Backend* has started correctly.
+
+This document ends with appendices:
 
 * [Quick installation using a Docker container](#quick-installation-using-a-docker-container)
 * [Administrative tasks](#administrative-tasks)
 
 
+## Choices
```
mattias-p: I'll add that.

mattias-p: Oh crap. Now I remember why I didn't do that before. It's because including numbers in the titles affects the section anchors. This way you can't add another section at the beginning or in between two other sections without messing with incoming links from other documents.

So which way hurts the least? Cemented section numbering or section titles without numbers?

docs/installation.md
mattias-p: @sandoche2k When replying to a sub-thread like this one, do you think you could add your comment to the actual sub-thread rather than the main thread of the PR? That way we can keep our parallel discussions separated.

mattias-p:
I may have made a mistake addressing @matsduf's comment about section numbering, and now Github is hiding the sub-thread because of the change.

> Oh crap. Now I remember why I didn't [add numbers to section titles] before. It's because including numbers in the titles affects the section anchors. This way you can't add another section at the beginning or in between two other sections without messing with incoming links from other documents.
>
> So which way hurts the least? Cemented section numbering or section titles without numbers?
—https://github.com/dotse/zonemaster-backend/pull/232#discussion_r89758109

docs/installation.md
mattias-p: > No. I wanted "Dependecies" section to be added just before #installation instructions for dependencies.
—@sandoche2k

I believe the section for installation instruction for dependencies is already titled "Dependencies".

matsduf:
In general, we should probably avoid linking into a section of a document, since there is always a risk that the headlines are changed. I think the numbering makes it clearer. To make the link work again I guess you have to add the number in the link too.

mattias-p:
@matsduf I specifically want to link to the Dependencies, Configuration, Startup and Post-Installation Sanity Check sections of this document from the SystemTesting document. The Choices, Prerequisites and Installation sections don't apply to the SystemTesting document.

The way I do documentation, you need to refer to sections of other documents (whether it's with anchored links or not) so the section titles can't be changed without care. I'm not saying that my way is the only way, but if I'm going to do it some other way, I'll need some guidance.

mattias-p:
I believe I've addressed all issues that came up during the review of this PR. If nobody disagrees, this can be merged.

Issue #176, opened 2016-06-10
Backend API needs further specification


matsduf:
The documentation for the API specifies the query format and the response format when successful. There are multiple error situations that will result in a response that is not documented. The API document must cover the total range of replies that can be returned.

matsduf:
Merge https://github.com/dotse/zonemaster-backend/blob/master/docs/API.md and https://github.com/dotse/zonemaster-backend/blob/master/docs/notes_api.md into one document, which is updated with as complete information as possible. If information is missing, then a placeholder should be put in the doc.

matsduf:
@mtoma , send Jan's document on API template.

matsduf:
Merging of #233 has resolved enough of this issue to close it.

PR #233, opened 2016-11-24
API documentation

mattias-p:
This integrates notes_api.md into API.md, adds lots of details, and splits it all up into a number of different documents and links to all of them from the documentation section of README.

I'd say this is an improvement over the previous situation, but it's far from complete. For instance there are still almost 50 "TODO" notes in API.md.


docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+
+* `"client_id"`: A free-form string, optional.
+* `"domain"`: A *domain name*, required.
+* `"profile"`: A *profile name*, optional.
+* `"client_version"`: A free-form string, optional.
+* `"nameservers"`: A list of *name server* objects, optional.
+* `"ds_info"`: A list of *DS info* objects, optional.
+* `"advanced"`: A boolean, optional.
+* `"ipv6"`: A boolean, optional.
+* `"ipv4"`: A boolean, optional.
```
matsduf: Add that default of both ipv4 and ipv6 is "false".

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
-  "level": "NOTICE",
-}
-```
+An object with the following properties:
 
-Or
+* `"id"`: A *test id*, required.
+* `"language"`: A string, required. Must be at least two characters long. The
+  two first characters are used to look up the *translation language* to be
+  used. If the lookup fails, the choice defaults to English.
```
matsduf: List supported language tags. If the lookup the messages can be broken since the even for English "translation" is needed.

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
-### *JSON-RPC Call*: `add_api_user`
+#### `"params"`
 
-**Request**:
-```
+An object with the following properties:
+
+* `"offset"`: An integer, optional. (default: 0).
+* `"limit"`: An integer, optional. (default: 200).
+* `"frontend_params"`: As described below.
```
matsduf: Is "offset" and "limit" really implemented?

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+
+>
+> TODO: Method description.
+>
+
+All the domains will be tested using identical parameters.
+
+If an identical *test* for a domain was already enqueued and hasn't been started or was enqueued less than 10 minutes earlier,
+no new *test* is enqueued for this domain.
+
```
matsduf: A "batch user" cannot start a new batch before the first batch has completed. If a second "batch user" is created, then a second batch can be started in parallel with the first batch.

docs/Architecture.md
```diff
@@ -0,0 +1,87 @@
...
+
+
+### Worker
+
+A Zonemaster *Worker* is a daemon that picks up *test* requests from the
+*Database*, runs them using the *Engine* library, and records the results back
+to the *Database*. A single *Worker* may handle several requests concurrently.
+The *Backend* architecture supports a single *Workers* interacting with a single
+*Database*.
+
```
matsduf: If the workers picks the tests from different queues.

README.md
```diff
@@ -35,14 +35,26 @@ instructions](docs/installation.md).
```
mattias-p: Isn't this section supposed to be about listing and documenting the individual configuration options rather than the simplest possible steps to the application up and running (which is what's described in the installation guide)?

docs/API.md
mattias-p: The set of translation languages is configurable, right? And it's the Zonemaster Engine configuration that's the authority? I think the Backend documentation should link into the Engine documentation where it describes how you can inspect the set of translation languages.

docs/API.md
mattias-p: This is from the current documentation. If it's not implemented that's a bug, right?

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+2. If the string is a single character, that character must be `.`.
+
+3. The length of the string must not be greater than 254 characters.
+
+4. When the string is split at `.` characters (after IDNA conversion,
+   if needed), each component part must be at most 63 characters long.
+
+5. Each such component part must also consist only of the characters
+   `0` to `9`, `a` to `z`, `A` to `Z` and `-`.
```
matsduf: If we have the classical limitation (as BASIC00 requires) then the label may not start nor end with a "-".

But as it is implemented, there are no visible limitations on the character set. The domain can contains any character. This is in conflict with BASIC00.

docs/API.md
mattias-p: What's the behavior when an api user tries to add a second batch while his first batch is running? I haven't succeeded in adding any api users, and I haven't had time to investigate it further, so I can't check this myself right now.

docs/API.md
mattias-p: I understand it that BASIC00 is the matter of some controversy at the moment, is that correct? If so, shouldn't we defer things that depend on it until its issues have been straightened out? I figure we'll need to go over things after the specification is settled anyhow.

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+The HTTP request must contain the header `Content-Type: application/json`.
+
+All JSON-RPC request and response objects have the keys `"jsonrpc"`, `"id"` and `"method"`.
+For details on these, refer to the JSON-RPC 2.0 specification.
+
+
+### Deviations from JSON-RPC 2.0
+
+* The `"jsonrpc"` property is not checked.
+* The error code -32603 is used for invalid arguments, as opposed to -32602.
```
mattias-p: Fixing with a slightly different wording.

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+* When standard error codes are used, the accompanying messages are not the standard ones.
+
+
+### Notes on the JSON-RPC 2.0 implementation
+
+* Extra top-level properties in request objects are allowed but ignored.
+* Extra properties in the `"params"` object are allowed for some methods but ignored for others.
+* Error messages from the API should be considered sensitive sas they leak details about the internals of the application and the system.
```
mattias-p: Fixing with a slightly different wording.

docs/API.md
```diff
@@ -1,681 +1,1085 @@
...
+* The error code -32601 is used when the `"method"` property is missing, as opposed to -32600.
```
mattias-p: Fixing with a slightly different wording.

mtoma:
It is implemented but not tested. Will require under 1h of work to fix if necessary.

----- Original Message -----

README.md
```diff
@@ -35,14 +35,26 @@ instructions](docs/installation.md).
 
 ### Documentation
 
-There is a fully documented [API](docs/API.md), which is the primay way
-to use the backend. The [docs](docs/) directory also contains the SQL commands
-for manipulating the database. 
+The Zonemaster Backend documentation is split up into several documents:
```
sandoche2k: Test

README.md
matsduf: My comment refer to line 33, which is one line before what we see. It was not clear.

docs/API.md
matsduf: * Bullet 1 gives a correct picture of the behavior of the backend. It does impose this features, but too generously (even invalid U-labels are converted to A-labels).
* Bullet 2-4 gives a correct picture of the behavior of the backend. It does impose those restrictions.
* Bullet 5 does not give a correct picture of the behavior of the backend. It does not impose those restrictions.

BASIC00 is a different, even though related, question.

docs/API.md
matsduf: No, it is hardcoded in the backend as far as I can see:

```
    if ( $browser_lang eq 'fr' ) {
        setlocale( LC_ALL, "fr_FR.UTF-8" );
    }
    elsif ( $browser_lang eq 'sv' ) {
        setlocale( LC_ALL, "sv_SE.UTF-8" );
    }
    else {
        setlocale( LC_ALL, "en_US.UTF-8" );
    }
```

From Zonemaster/WebBackend/Translator.pm

docs/API.md
matsduf: @mtoma, what is the status of "offset" and "limit"? I can see code for Mysql and PostgreSQL, but is it working?

docs/API.md
matsduf: You get an error back saying that you already run a batch:

```json
{
  "error": {
    "message": "You can't create a new batch job, job:[7] started on:[2016-11-09 10:39:02] still running  at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/DB/MySQL.pm line 89.\n",
    "code": -32603
  },
  "id": "MD4712",
  "jsonrpc": "2.0"
}
```

mtoma:
The API behavior when it finds one or more tests with progress < 100 for a batch job from a user that tries to add another batch job is to refuse to add the new batch job.

----- Original Message -----

docs/API.md
mattias-p: So I will remove item 5 as this requirement is not imposed.

What should I do about item 1? Remove it? Update its wording? To what?

matsduf: "Currently there are no restrictions on what characters that are allowed."

docs/API.md
mattias-p: Documented this as an error state.

docs/API.md
mattias-p: I can limit results by twiddling with the "offset" and "limit" parameters.

docs/Architecture.md
```diff
@@ -0,0 +1,87 @@
+# Architecture
+
+The Zonemaster *Backend* is a system for performing domain health checks and
+keeping records of performed domain health checks.
+
+A Zonemaster *Backend* system consists of at least three components: a
+single *Database*, one or more *Workers* and one or more *Web backends*.
```
matsduf: Only one Worker unless we also document the queues.

docs/Architecture.md
matsduf: `single *Workers* interacting` --> `single *Worker* interacting`

mattias-p:
This PR resolves some but not all items in #176.
 * API and Notes_API have been fully merged.
 * Placeholders for missing information are very prevalent, but probably not complete.
 * The request format is explained to some degree, but it is not fully specified.
 * The response format is mostly descibed by example and still needs much improvement.
 * Error responses are hardly described at all.

In my view:
 * I have addressed all issues that came up during the review of this PR.
 * There is consensus within the team that the current state of this PR is an improvement over the current state of the master branch.

If nobody disagrees with the above, this can be merged.

matsduf:
Does anyone have any problem if I merge this?

Issue #270, opened 2017-02-24
"default_profile" would not be loaded


matsduf:
The code below will load a profile (policy) if the parameter has been set in the API and it has been set to "default_profile". The other legal value, "default_profile", will be ignored. (The third legal value, "default_profile", will be replace by "default_profile" earlier.)

```
$ cat -n /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm | grep -A8 123
   123	    if ( $params->{profile} eq 'test_profile_1' ) {
   124			if (Zonemaster::WebBackend::Config->CustomProfilesPath()) {
   125				Zonemaster->config->load_policy_file( Zonemaster::WebBackend::Config->CustomProfilesPath() . '/iana-profile.json' );
   126			}
   127			else {
   128				Zonemaster->config->load_policy_file( 'iana-profile.json' );
   129			}
   130		}
   131	
```

matsduf:
Fixed by https://github.com/dotse/zonemaster-backend/pull/271

Issue #206, opened 2016-11-08
Perl warning when creating a batch


matsduf:
I had the following JSON file:

```
{
   "method" : "add_batch_job",
   "params" : {
      "domains" : [
          "00000.se.",
          "0001.se.",
          "00046.se.",
          "0006.se.",
          "000kr.se.",
          "000px.se.",
          "0-0-0.se.",
          "000.se.",
          "0011.se.",
          "001masc.se."
      ],
      "username" : "xxxxx",
      "test_params" : {
          "ipv6" : true,
          "ipv4" : true
      },
      "api_key" : "xxxxxxxxxxx"
   },
   "jsonrpc" : "2.0",
   "id" : "MD4712"
}
```

I ran and got warnings from perl:

```
$ curl -H "Content-Type: application/json" -d @add_batch_job.json http://localhost:8080/ ; echo; echo ; echo
{"result":"5","jsonrpc":"2.0","id":"MD4712"}

$ Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
Use of uninitialized value in string eq at /usr/local/share/perl/5.18.2/Zonemaster/WebBackend/Runner.pm line 117.
```

The code says

```
[...]

PR #271, opened 2017-02-24
Updated loading of profile (policy) so undefined value is not compare…

matsduf:
Updated loading of profile (policy) so undefined value is not compared to a string causing warning (resolving issue #206). Updating so that if "default_profile" is selected, that should be loaded (resolving issue #270).

### DIFF
--- a/lib/Zonemaster/WebBackend/Runner.pm
+++ b/lib/Zonemaster/WebBackend/Runner.pm
@@ -121,15 +121,16 @@ sub run {
     }
     
 
-    # If the profile parameter has been set in the API, then load a profile, else Engine configuration will decide what profile (policy) to load.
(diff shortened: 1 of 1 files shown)


mattias-p:
The comments explain things that are actually easier to understand from the code, and the first one assumes things about the calling code. Comments like this are not helpful to the reader. They act as time sinks for the reader to every time verify that the code and comments match up. In the end they teach readers to ignore comments entirely.

A helpful comment would be a very short one simply declaring what kind of thing the section is about (in relation to nearby sections). It should be very brief. Preferably just two or three words. More than five words is a stretch. Comments of this type help readers scanning over the code to quickly decide whether to take a closer look at this section or ignore it.

PR #277, opened 2017-04-06
Modified instances of JSON to JSON::PP in Engine.pm and DB/SQlite.pm

sandoche2k:


lib/Zonemaster/WebBackend/Engine.pm
```diff
@@ -177,21 +177,21 @@ sub validate_syntax {
 
     if ( defined $syntax_input->{advanced} ) {
         return { status => 'nok', message => encode_entities( "Invalid 'advanced' option format" ) }
-          unless ( $syntax_input->{advanced} eq JSON::false || $syntax_input->{advanced} eq JSON::true );
+          unless ( $syntax_input->{advanced} eq JSON::PP::false || $syntax_input->{advanced} eq JSON::PP::true );
```
matsduf: When I print JSON::false it returns "false". When I print JSON::PP::false it returns 1. Maybe it is better to replace with "false" (and "true")?

PR #278, opened 2017-04-06
- Modify blocking tests.

vlevigneron:
In this proposal, I just modified the tests that prevent from building the module. The code has not been changed.

### DIFF
--- a/t/test01.t
+++ b/t/test01.t
@@ -4,6 +4,7 @@ use 5.14.2;
 
 use Test::More;    # see done_testing()
 use Zonemaster;
+use JSON;
 
 my $datafile = q{t/test01.data};
 if ( not $ENV{ZONEMASTER_RECORD} ) {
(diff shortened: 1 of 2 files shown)


mtoma:
This is not ok. The idea behind the boolean options evaluation is to:
consider ok everything perl evaluates as true
AND additionally consider true the JSON::(PP)::true object if it fails with =1 than it must be fixed elsewhere.

t/test01.t
```diff
@@ -4,6 +4,7 @@ use 5.14.2;
 
 use Test::More;    # see done_testing()
 use Zonemaster;
+use JSON;
```
matsduf: Should we really use JSON here if we use JSON::PP elsewhere?

pawal: If you use JSON it will automatically use JSON::PP since both are core packages. But it will also automatically pick up JSON::XS if it is installed, which I would very much prefer. It increases performance.

mtoma:
Hi Patrick,

Please have a look at this issue:
https://github.com/dotse/zonemaster-engine/issues/261

It explains why we can't use JSON::XS.

If we have a clean solution for this problem that would work without doing anything specifc code depending on particular JSON::XS and/or perl versions I'd happily replace JSON::PP with the JSON::XS module.

I think all in all the engine does exactly 1 json_encode call and the backend 4 encode/decode calls. I don't think there could be any real performance issue here especially compared to the time it takes to run a zonemaster test.

Michal

----- Original Message -----

pawal:
Yes, I commented on this yesterday. Have you reported the problem
upstream?

Also, there could also be a recommendation to not use the XS version
together with "odd" locales unless you use Perl 5.2x.

matsduf:
We do not do this update. The unit test must work with 1 and 0.

PR #282, opened 2017-04-10
Fix issue 281

mtoma:


matsduf:
Should this PR really be there? It seems to be a mixture of two things. I think PR #280 and #281 resolves the issue? Or is this meant to be an update of #281? Then please do that.

mtoma:
No I created two separate issues, this is probably due to the fact that I didn't merge the first issue before this one.

PR #291, opened 2017-06-29
Fix isssue 287

mtoma:


lib/Zonemaster/WebBackend/DB/MySQL.pm
```diff
@@ -221,6 +221,8 @@ sub get_test_history {
 
     my @results;
     
+    return \@results unless ($p->{frontend_params} && $p->{frontend_params}{domain});
```
matsduf: Is this really needed when the empty call is stopped in the API?

lib/Zonemaster/WebBackend/DB/PostgreSQL.pm
```diff
@@ -204,7 +204,10 @@ sub get_test_history {
       : ( "AND (params->'nameservers') IS NULL" );
 
     my @results;
-    my $query = "
+	
+	return \@results unless ($p->{frontend_params} && $p->{frontend_params}{domain});
```
matsduf: Is this really needed when the empty call is stopped in the API?

lib/Zonemaster/WebBackend/DB/SQLite.pm
```diff
@@ -233,6 +233,8 @@ sub get_test_history {
     my ( $self, $p ) = @_;
 
     my @results;
+	return \@results unless ($p->{frontend_params} && $p->{frontend_params}{domain});
```
matsduf: Is this really needed when the empty call is stopped in the API?

lib/Zonemaster/WebBackend/Engine.pm
```diff
@@ -375,7 +375,10 @@ sub get_test_results {
 sub get_test_history {
     my ( $self, $p ) = @_;
 
-    my $results = $self->{db}->get_test_history( $p );
+    my $results = [];
+    if ($p->{frontend_params} && $p->{frontend_params}{domain}) {
```
matsduf: In the `start_domain_test` method the following code is used to accomplish the same thing as far as I can see. Why not do it in the same way? Why returning an empty result instead of refusing to run the method?

`die "No domain in parameters\n" unless ( $params->{domain} );
`

mattias-p:
What is this fixing? I can't reproduce the problem.

mattias-p:
Anyway I think it would be more reasonable API-wise to make the domain parameter required, instead of having it be optional and return an empty list if omitted.

sandoche2k:
@mtoma can you have a look at the modification proposed by @matsduf and update the PR

mtoma:
Closing and replacing by:
https://github.com/dotse/zonemaster-backend/pull/322

PR #308, opened 2017-10-31
Update licensing

mattias-p:


mattias-p:
@matsduf Good catch! But no, the value of field is supposed to be a token, not a url.
http://search.cpan.org/~ether/Module-Install-1.18/lib/Module/Install.pod#license

PR #310, opened 2017-11-01
Release 2.0.0

mattias-p:


Changes
```diff
@@ -1,6 +1,22 @@
 Revision history for Zonemaster Backend.
```
matsduf: Should we be consistent, or it does not matter?

mattias-p: Updated the heading to be consistent with your CLI PR.

PR #312, opened 2017-11-03
without libpq-dev backend installation fails

sandoche2k:


mattias-p:
How does it fail? On the surface this looks strange to me for two reasons. First, I don't recall ever installing this package. Second, would a PostgreSQL package really be required even for MySQL installations? If it is indeed required, I suggest that the entry is moved to the PostgreSQL dependencies.

matsduf:
@sandoche2k, can you please give more details. What OS and version did you run into problem? How did it fail? When did it fail? I have tested both Ubuntu (14.04 and 16.06) and Debian (7.11) without seeing any problem.

mattias-p:
Is this about the user getting prompted about the path `pg_config`? If that's the case a better solution would be to instruct the user to just press enter in case that happens.

matsduf:
@sandoche2k, for 2017.3 release we do not support PostgreSQL on neither CentOS, Debian nor Ubuntu 16.04. We have not tested PostgreSQL and we have no instructions. For FreeBSD the only supported DB is PostgreSQL. Ubuntu 14.04 is the only OS we have support for both DBs.

sandoche2k:
@matsduf :    Debian GNU/Linux 8.8 (jessie)
@mattias-p : It has failed for MySQL. I will start a new test to check and will publish the error report

sandoche2k:
@mattias-p After pressing just enter for path pg_config, i get the following error:
sh ```
Checksum for /root/.cpan/sources/authors/id/T/TU/TURNSTEP/DBD-Pg-3.7.0.tar.gz ok
Configuring T/TU/TURNSTEP/DBD-Pg-3.7.0.tar.gz with Makefile.PL
Configuring DBD::Pg 3.7.0
Path to pg_config? 
No POSTGRES_HOME defined, cannot find automatically
No 'Makefile' created  TURNSTEP/DBD-Pg-3.7.0.tar.gz
  /usr/bin/perl Makefile.PL INSTALLDIRS=site -- NOT OK
```

sandoche2k:
From the backend Makefile.pl 
 ```sh
recommends 'DBD::mysql';
recommends 'DBD::Pg';
recommends 'DBD::SQLite';
 ```
Do we need to install these packages while installing the backend?

matsduf:
@sandoche2k, I suggest that you create an issue of these problem instead, and then we can find a solution. Also see https://github.com/dotse/zonemaster/issues/524

matsduf:
I have tested to install Backend on both Centos 7 and Debian 7.11. Both were clean installations. I followed the installations word for word. Besides the issue in #314 there were no issues at all.

@sandoche2k, did you install on a clean installation of Debian 8?

sandoche2k:
@mattias Yes, it was a clean installation. Anyway, as you are not able to reproduce the issue, i will close it down.

Issue #317, opened 2017-11-28
Complete specification of API input parameter formats


mattias-p:
### Purpose and scope

The purpose of this issue is to document the status quo and introduce reasonable constraints where appropriate. Major changes to the API such renaming or moving parts of parameter structures should not be included in this issue.

### Requirements

- [x] Every parameter of every method must be described with name and data type. If the method accepts a single scalar value as its parameter, specifying the data type is sufficient for the purpose of this issue. (#320)
- [x] Every data type with the basic data type `array` must specify the data type of the elements. (we don't have any of these)
- [x] Every data type with the basic data type `object` must specify all its keys with name and value data type.
- [x] Every data type must be fully specified with basic data type and additional restrictions (e.g. regular expression).
- [x] Every parameter and object property value must be declared to be either optional or required.
- [x] Every optional parameter and object property value must declare its default value.

pamasse:
@mattias-p I checked some checkbox based on the pr #320

sandoche2k:
@pamasse and @mattias-p Please focus on completing the specification.

sandoche2k:
user_ip, location data type, user_location_info deprecated conforming to GDPR requirements in release 2018.1. Delete these parameters in the backend, if they are already present.

pamasse:
I think this request is too strict :  "Every optional parameter and object property value must declare its default value." 
Some optional don't have default value. 
For instance, we don't have default value for "ds_info" or "nameservers".

pamasse:
We need to add some restriction on Domain name and Username data types.

sandoche2k:
@mattias-p @pamasse close?

sandoche2k:
@mattias-p update with the missing information

mattias-p:
Three more things remain before we can check the two bottom boxes:
1. *DS Info* does not specify default values for its optional properties.
2. The `"frontend_params"` parameter of the *get_frontend_params* method is neither required nor optional.
3. The `"test_params"` parameter of the *add_batch_job* method is neither required nor optional.

sandoche2k:
@mattias-p if completed we could close this issue and create specific issues for any missing input parameters

PR #320, opened 2017-12-13
Update API

pamasse:
I update the API in order to clean it. I add some examples.
I also added comments to questions that were in TODOs.

#317

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -77,7 +77,7 @@ is based on a JSON data type, but additionally imposes its own restrictions.
 
 Basic data type: string
 
-This parameter is a free-form string that represent the password of an authenticated account (see [*Privilege levels*](API.md#Privilege levels))
+This parameter is a free-form string that represent the password of an authenticated account (see [*Privilege levels*](API.md#privilege-levels))
 
 > Note: Currently there are no restrictions on what characters that are allowed.
 
@@ -284,11 +284,11 @@ Basic data type: string
  An unsigned integer is either positive or zero.
  
 
-### User name
+### Username
 
 Basic data type: string
  
-This parameter is a free-form string that represent the name of an authenticated account (see [*Privilege levels*](API.md#Privilege levels))
+This parameter is a free-form string that represent the name of an authenticated account (see [*Privilege levels*](API.md#privilege-levels))
 
 > Note: Currently there are no restrictions on what characters that are allowed.
 
@@ -497,7 +497,6 @@ Example request:
       }
     ],
     "ds_info": [],
-    "advanced": true,
     "ipv6": true,
     "ipv4": true
   }
@@ -716,10 +715,7 @@ Example request:
     "offset": 0,
     "limit": 200,
     "frontend_params": {
-      "client_id": "Zonemaster Dancer Frontend",
-      "domain": "zonemaster.net",
-      "profile": "default_profile",
-      "client_version": "1.0.1",
+      "domain": "zonemaster.net",      
       "nameservers": [
         {
           "ns": "ns3.nic.se",
@@ -729,11 +725,7 @@ Example request:
           "ns": "ns2.nic.fr",
           "ip": "192.93.0.4"
         }
-      ],
-      "ds_info": [],
-      "advanced": true,
-      "ipv6": true,
-      "ipv4": true
+      ]
     }
   }
 }
@@ -779,10 +771,10 @@ An object with the following properties:
 
 The value of "frontend_params" is an object with the following properties:
 
-* `"client_id"`: A *client id*, optional.
+* `"client_id"`: **Deprecated**. A *client id*, optional.
 * `"domain"`: A *domain name*, required.
-* `"profile"`: A *profile name*, optional.
-* `"client_version"`: A *client version*, optional.
+* `"profile"`: **Deprecated**. A *profile name*, optional.
+* `"client_version"`: **Deprecated**. A *client version*, optional.
 * `"nameservers"`: A boolean in order to return either "regular" (false) or "undelegated" (true), optional.
 * `"ds_info"`: **Deprecated**. A list of *DS info* objects, optional.
 * `"advanced"`: **Deprecated**. A boolean, optional.


docs/API.md
```diff
@@ -768,9 +775,11 @@ An object with the following properties:
 
 ## API method: `add_api_user`
 
->
-> TODO: Method description.
->
+In order to use avance api features such as the *batch test*, it's necessaire to previously create an api key.
```
mattias-p: Spelling: "avance" -> "advanced"?

docs/API.md
```diff
@@ -768,9 +775,11 @@ An object with the following properties:
...
+This key can be obtain with the creation of an user in the system.
```
mattias-p: "an user" -> "a user"

docs/API.md
```diff
@@ -801,38 +810,38 @@ Example response:
...
 
 #### `"result"`
 
-An integer.
-
->
-> TODO: Describe the possible values of the result and what they mean.
->
-
+An integer. The value could be equal to 1 if the registration is a success, or 0 if the SQL request failed.
```
mattias-p: "could be" -> "is"

docs/API.md
```diff
@@ -195,21 +203,28 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
...
 
 
 ### Timestamp
 
 Basic data type: string
 
->
-> TODO: Specify date format
->
+Default database timestamp format.
```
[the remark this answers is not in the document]
matsduf: The timestamp format should be database engine independent. We could even require that all timestamps be in UTC.

docs/API.md
```diff
@@ -195,21 +203,28 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
...
+Basic data type: object
+
+The object has five keys, `"isp"`, `"country"`, `"city"`, `"longitude"`  and `"latitude"`.
+
+
+* `"isp"`: a string. The Internet Service Provider of the user.
+* `"country"`: a string. The country of the user.
+* `"city"`: a string. The city of the user.
+* `"longitude"`: a string. The longtitude of the user.
+* `"latitude"`: a string. The latitude of the user.
```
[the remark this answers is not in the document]
matsduf: Yes, these strings must be well-defined in terms of what chararcter that may be used. The country strings should be something from the to ISO 3166-1 alpha-2 country codes. I think the validation can accept '^[A-Z]{2}$' to make it easy.

docs/API.md
```diff
@@ -451,43 +462,32 @@ Example response:
...
 
 A *test id*. The newly started *test*, or a recently run *test* with the same
-parameters.
-started within the recent configurable short time.
+parameters. Started within the recent configurable short time.
 
 >
 > TODO: Specify which configuration option controls the duration of the window
-> of *test* reuse.
+> of *test* reuse. => I don't know (and i didn't find any traces) if such a feature exists.
```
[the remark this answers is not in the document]
pamasse: Thanks to Michal, i finally understood. 
Line 292 in RPCAPI.pm, the value "10" is for the variable "$minutes_between_tests_with_same_params" which sets the interval.

docs/API.md
```diff
@@ -195,29 +203,41 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
...
+* `"ns"`: a *domain name*. The name server used by the *test module*. 
+This key is added when the module name is `"NAMESERVER"`.
 
 
 ### Timestamp
 
 Basic data type: string
 
+Default database timestamp format: "Y-M-D H:M:S.ms"
+Example: "2017-12-18 07:56:17.156939"
```
matsduf: I think the format should prescribed. Not just default format.

docs/API.md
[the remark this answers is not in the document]
mtoma: As the Timestamp data type is only used in API responses and doesn't require any understanding by anyone else that the API users I would leave this as it was:  "Default database timestamp format" with a possible rewording as so: "A time in a format defined by your local database setup and configuration". Otherwise the backend needs updates to force this to be something standard.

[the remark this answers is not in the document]
matsduf: I still think that the timestamp format should be specified.

docs/API.md
```diff
@@ -725,20 +721,26 @@ will be used to look up all tests for the given domain, separated
...
+* `"ds_info"`: A list of *DS info* objects, optional.
+* `"advanced"`: **Deprecated**. A boolean, optional.
+* `"ipv6"`: A boolean, optional. (default: `false`)
+* `"ipv4"`: A boolean, optional. (default: `false`)
+* `"config"`: A string, optional. The name of a *config profile*.
+* `"user_ip"`: An IP, optional.
+* `"user_location_info"`: An *location* object, optional.
+* `"priority"`: A *priorty*, optional.
+* `"queue"`: A *queue*, optional.
+
```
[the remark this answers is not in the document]
mattias-p: The following documented parameters are being ignored by the backend:
* `client_id`
* `profile`
* `client_version`
* `ds_info`
* `advanced`
* `ipv4`
* `ipv6`

Please remove them from both the example request and the params documentation.

N.B. When using the SQLite database engine, the `nameservers` parameter is also ignored, but for the purpose of this PR (at least) we should consider that a bug.

docs/API.md
mattias-p: @mtoma says Timestamp is only used in responses. This PR is about request parameters. I see no need to get this right before merging.

docs/API.md
```diff
@@ -100,8 +100,9 @@ Basic data type: string
 
 Basic data type: object
 
-Properties:
+DS for [Delegation Signer](https://tools.ietf.org/html/rfc3658) references DNSKEY-records in the sub-delegated zone.
```
matsduf: RFC 3658 has  been obsoleted by RFC 4034.

docs/API.md
```diff
@@ -100,8 +100,9 @@ Basic data type: string
...
 
+Properties:
```
matsduf: On the next lines it says that `digest` is required but `algoritm`, `digtype` and `keytag` are optional. All of them must be required. Or?

It also says that "extra properties in *DS info* objects are ignored when present in RPC method arguments" but it should rather be that if there are extra properties, then the call is refused.

Also further down under "name sever" it says that `ip` is required, but it should be optional.

docs/API.md
matsduf: I can accept that. In an upcoming update of the API we should get it right.

docs/API.md
```diff
@@ -451,43 +466,29 @@ Example response:
 
 An object with the following properties:
 
-* `"client_id"`: A free-form string, optional.
-* `"domain"`: A *domain name*, required.
-* `"profile"`: A *profile name*, optional.
-* `"client_version"`: A free-form string, optional.
-* `"nameservers"`: A list of *name server* objects, optional.
-* `"ds_info"`: A list of *DS info* objects, optional.
+* `"client_id"`: A free-form string, optional. Used to monitor which client uses the API.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -451,43 +466,29 @@ Example response:
...
+* `"domain"`: A *domain name*, required. The zone to test.
+* `"profile"`: A *profile name*, optional. Used to perform the test with a specific set of parameters and tests.
+* `"client_version"`: A free-form string, optional. Used to monitor which client use the API
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -451,43 +466,29 @@ Example response:
...
 An object with the following properties:
 
-* `"client_id"`: A free-form string, optional.
-* `"domain"`: A *domain name*, required.
-* `"profile"`: A *profile name*, optional.
-* `"client_version"`: A free-form string, optional.
-* `"nameservers"`: A list of *name server* objects, optional.
-* `"ds_info"`: A list of *DS info* objects, optional.
+* `"client_id"`: A free-form string, optional. Used to monitor which client uses the API.
+* `"domain"`: A *domain name*, required. The zone to test.
```
matsduf: We should have restrictions on what code points that the domain may consist of. And a lenght limit.

docs/API.md
```diff
@@ -451,43 +466,29 @@ Example response:
...
-> `"user_ip"` and `"user_location_info"`.
->
-> TODO: Clarify the purpose of each `"params"` property.
->
-> TODO: Clarify the default value of each optional `"params"` property.
->
+* `"user_ip"`: An IP, optional. Used to monitor information about the user. (We only keep the location of the IP).
+* `"user_location_info"`: An *location* object, optional. Used to monitor information about the user. 
+* `"priority"`: A *priority*, optional.
+* `"queue"`: A *queue*, optional.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -451,43 +466,29 @@ Example response:
...
-> TODO: Clarify the data type of the following `"params"` properties:
-> `"user_ip"` and `"user_location_info"`.
->
-> TODO: Clarify the purpose of each `"params"` property.
->
-> TODO: Clarify the default value of each optional `"params"` property.
->
+* `"user_ip"`: An IP, optional. Used to monitor information about the user. (We only keep the location of the IP).
+* `"user_location_info"`: An *location* object, optional. Used to monitor information about the user. 
+* `"priority"`: A *priority*, optional.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
->
-> TODO: Describe the purpose of `"offset"` and `"limit"`
->
-> TODO: Is the `"nameservers"` value a boolean in disguise?
->
-> TODO: The description of `"frontend_params"` is clearly not up to date. Can it
-> be described in a better way?
->
-
+* `"client_id"`: A free-form string, optional.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"profile"`: A *profile name*, optional.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"client_version"`: A free-form string, optional.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"nameservers"`: A boolean in order to return either "regular" (false) or "undelegated" (true).
+* `"ds_info"`: A list of *DS info* objects, optional.
```
matsduf: Shouldn't this be deprecated?

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"advanced"`: **Deprecated**. A boolean, optional.
+* `"ipv6"`: A boolean, optional. (default: `false`)
```
matsduf: Shouldn't this be deprecated?

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"ipv4"`: A boolean, optional. (default: `false`)
```
matsduf: Shouldn't this be deprecated?

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"config"`: A string, optional. The name of a *config profile*.
```
matsduf: Shouldn't this be deprecated?

docs/API.md
```diff
@@ -715,29 +721,23 @@ Example response:
...
+* `"user_ip"`: An IP, optional.
+* `"user_location_info"`: An *location* object, optional.
+* `"priority"`: A *priorty*, optional.
+* `"queue"`: A *queue*, optional.
```
matsduf: These fields have no relevance for get_test_history as far as I can see.

docs/API.md
```diff
@@ -801,38 +808,38 @@ Example response:
...
-* `"username"`: A string, optional. The name of the user to add.
-* `"api_key"`: A string, optional. The API key (in effect, password) for the user to add.
-
->
-> TODO: Are `"username"` and `"api_key"` really supposed to be optional? Because
-> they are now, is that a bug? I get `"result": 0` when I omit them. I would
-> have expected parameter validation errors.
->
-
+* `"username"`: A string, required. The name of the user to add.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
```diff
@@ -801,38 +808,38 @@ Example response:
...
+* `"api_key"`: A string, required. The API key (in effect, password) for the user to add.
```
matsduf: We should put a limit of what character to use here. And lenght.

docs/API.md
mattias-p: Please add this paragraph before the "`frontend_params`" properties list:

> The value of "`frontend_params`" is an object with the following properties:

Also the "`domain`" parameter seems to have gone missing.

docs/API.md
mattias-p: `client_id` should specify a data type and the data type should specify a regular expression.

docs/API.md
mattias-p: @matsduf Check the *domain name* data type specification.

docs/API.md
mattias-p: `client_version` should specify a data type and the data type should specify a regular expression.

docs/API.md
mattias-p: @matsduf Check the *priority* data type specification.

The *priority* data type specification should be updated with a range restriction though, and possibly a clarification on what precision the number is rounded to.

docs/API.md
mattias-p: @matsduf Check the *queue* data type specification.

The *queue* data type specification should be updated with a range restriction though, and possibly a clarification on what precision the number is rounded to.

docs/API.md
mattias-p: `client_id` should refer to the same data type as its namesake in start_domain_test.

docs/API.md
mattias-p: @matsduf Check the *profile name* data type specification.

The *profile name* data type specification will be updated as part of the profiles change. Like this: https://github.com/mattias-p/zonemaster-backend/blob/profiles-doc/docs/API.md#profile-name

I don't think this PR needs to make any additional effort to improve the *profile name* data type specification.

docs/API.md
mattias-p: `client_version` should refer to the same data type as its namesake in start_domain_test.

docs/API.md
mattias-p: `username` should specify a data type and the data type should specify a regular expression.

docs/API.md
mattias-p: `api_key` should specify a data type and the data type should specify a regular expression.

docs/API.md
pamasse: @matsduf In the current state of the implementation of the API, `algoritm`, `digtype` and `keytag` are optional and extra properties are ignored. 
For the `ip`, i will change it in the next commit.

I think this PR is to update the API documentation that reflects its current implementation. Maybe your comments should be aggregated in a new issue to update the implementation of the API.

docs/API.md
pamasse: @matsduf i agree, they should be deprecated.

docs/API.md
pamasse: @matsduf i agree

mattias-p:
I reviewed the current state of this PR with regard to the first item of #317. A few things still need to be updated.

A new *IP address* data type needs to be added for the following parameters:
* start_domain_test.user_ip
* add_batch_job.user_ip
* validate_syntax.user_ip

A new *unsigned integer* data type needs to be added for the following parameters:
* get_batch_job_result.nb_finished
* get_batch_job_result.nb_running
* get_test_history.offset
* get_test_history.limit

These parameters need to refer to the *client id* and *client version* data types:
* add_batch_job.client_id
* add_batch_job.client_version
* validate_syntax.client_id
* validate_syntax.client_version

New data types have been added to the data types section. I wish these could be moved around so that they respect the alphabetical order.

The data type of this parameter got misspelled somehow. It should be fixed at some point, and now seems like a good time:
* add_batch_job.priority

Also this parameter refers to a data type that doesn't exist. I believe it will disappear as part of the work we're planning for profiles. Therefore I think it can be left as it is. I'm mentioning it here for completeness only.
* start_domain_test.config

docs/API.md
```diff
@@ -195,32 +256,43 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
...
 
 
+### Unsigned integer
+
+ Basic data type: number
+ 
+ An unsigned integer is either positive or zero.
+ 
+
+### User name
```
mattias-p: This name is contracted to "username" elsewhere. Please make it consistent.

docs/API.md
```diff
@@ -715,29 +773,22 @@ Example response:
...
->
-> TODO: The description of `"frontend_params"` is clearly not up to date. Can it
-> be described in a better way?
->
+The value of "frontend_params" is an object with the following properties:
 
+* `"client_id"`: A *client id*, optional.
+* `"domain"`: A *domain name*, required.
+* `"profile"`: A *profile name*, optional.
+* `"client_version"`: A *client version*, optional.
```
mattias-p: `client_id`, `client_version` and `profile` are ignored by the implementation, but they are not marked as deprecated here. If they aren't marked as deprecated an issue should be created to make the implementation respect them. If you like I can create the issue.

docs/API.md
```diff
@@ -715,29 +773,22 @@ Example response:
...
+* `"nameservers"`: A boolean in order to return either "regular" (false) or "undelegated" (true), optional.
+* `"ds_info"`: **Deprecated**. A list of *DS info* objects, optional.
+* `"advanced"`: **Deprecated**. A boolean, optional.
+* `"ipv6"`: **Deprecated**. A boolean, optional. (default: `false`)
+* `"ipv4"`: **Deprecated**. A boolean, optional. (default: `false`)
+* `"config"`: **Deprecated**. A string, optional. The name of a *config profile*.
```
mattias-p: Please remove these deprecated parameters from the request example above.

docs/API.md
```diff
@@ -73,15 +73,41 @@ This sections describes a number of data types used in this API. Each data type
 is based on a JSON data type, but additionally imposes its own restrictions.
 
 
+### API key
+
+Basic data type: string
+
+This parameter is a free-form string that represent the password of an authenticated account (see [*Privilege levels*](API.md#Privilege levels))
```
mattias-p: Please update the link target to #privilege-levels. Since it's in the same document you don't have to specify API.md. Also, if you hover your mouse over rendered markdown heading on Github, a link symbol appears with a link containing the fragment for the heading itself.

docs/API.md
```diff
@@ -195,32 +256,43 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
...
+ Basic data type: number
+ 
+ An unsigned integer is either positive or zero.
+ 
+
+### User name
+
+Basic data type: string
+ 
+This parameter is a free-form string that represent the name of an authenticated account (see [*Privilege levels*](API.md#Privilege levels))
```
mattias-p: Please update the link target to #privilege-levels.

mattias-p:
AFAICT this PR now documents every parameter with name and datatype. Good work @pamasse! By the way, good call marking up required parameters as such instead of just leaving them not marked up as optional.

I guess next up is declaring datatypes for the sub-properties of *DS Info* and *Location*. I've got some suggestions if you're not sure what to do about them.

mattias-p:
Oh, and please avoid using hash tags to denote enumeration in commit messages. Github interprets them as issue references.

mattias-p:
I approve of this PR, though I don't consider it to be fully complete with regard to #317.

matsduf:
@pamasse, could you look at comments etc to finialize this PR?

@sandoche2k and @mtoma, could you review this PR?

docs/API.md
```diff
@@ -73,15 +73,41 @@ This sections describes a number of data types used in this API. Each data type
...
+
+> Note: Currently there are no restrictions on what characters that are allowed.
+
+
+### Client version
+
+Basic data type: string
+
+This parameter is a free-form string that represent the version of the client. It used to monitor which client (GUI) uses the API.
+
```
sandoche2k: Is the above statement correct?

docs/API.md
```diff
@@ -451,43 +517,29 @@ Example response:
...
->
-> TODO: Specify which configuration option controls the duration of the window
-> of *test* reuse.
->
-
+A *test id*. 
 
+If the test has been run with the same domain name within an interval of 10 mins (hard coded), 
+then the new request does not trigger a new test, but returns with the results of the last test
+
```
sandoche2k: Why has it been modified from 5 mins to 10?

docs/API.md
```diff
@@ -671,11 +725,7 @@ Example request:
           "ns": "ns2.nic.fr",
           "ip": "192.93.0.4"
         }
```
sandoche2k: The nameservers will be changed. So it is a ToRemember.

docs/API.md
matsduf: From the code it looks like it has "always" been 10 minutes. One idea would be to have it configurable, but that should be a new issue.

PR #325, opened 2017-12-14
Fix issue 287

mtoma:


lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -374,7 +374,11 @@ sub get_test_results {
 sub get_test_history {
     my ( $self, $p ) = @_;
 
-    my $results = $self->{db}->get_test_history( $p );
+    my $results;
+    
+    return $results unless ($p->{frontend_params} && $p->{frontend_params}{domain});
```
mattias-p: It would make more sense to return an error here. Would that break the current GUI implementation?

matsduf: I know that the current GUI is dependent on an empty query being replied to in this way, so I think it is OK, but I think this should be documented in the code that solution is of that reason, that it is deprecated to be dependent on that behavior and that the plan is to do as @mattias-p suggests.

mattias-p:
So what is required for approving the PR? An update to the API documentation?

matsduf:
@mattias-p, I think we should see this update as a "bug" because the API dcumentation should state that domain must not be empty. I think that a comment with code plus an issue that this should be corrected as soon as the current GUI is phased out.

matsduf:
An issue #333 has been created to remove the work-around.

sandoche2k:
@mtoma merging, since we have consensus

Issue #213, opened 2016-11-11
Internal error details are leaked


mattias-p:
When an unhandled error occurs in the web backend an error reponse is given with error code -32603. For example if the `backend_config.ini` file is missing I get this response:
```json
{"jsonrpc":"2.0","id":"140715758026879","error":{"code":-32603,"message":"UNABLE TO LOAD /etc/zonemaster/backend_config.ini","code":-32603},"id":"140715758026879"}
```
This leaks details about the internals of the service. Instead this error message should be logged and the message in the response should be simply `"Internal error"` (as prescribed by the JSON-RPC 2.0 specification).

mattias-p:
This has not been fixed.

matsduf:
Issue #214 must be fixed to fix this since we need to send error messages to a log file.

mattias-p:
I had things mixed up when I spoke earlier. RPCAPI does have a log file. #314 seems to be about the volume and quality of the log file.

mtoma:
Not really sure about the perimeter of what needs to be done here.
My idea was to catch the JSON::RPC messages and redirect them to the log file and replace with "Internal Error" sent to the user.
Can't fix right now because search.cpanorg returns "Internal Server Error" right now.

mattias-p:
Maybe we could have a special wrapper for user-facing error messages and replace everything else with "Internal Server Error"? Then we can start adding that wrapper on a case-by-case basis. How does that sound?

I'm using metacpan.org where I can while search.cpan.org is down.

mtoma:
I don't really see a way to catch the JSON::RPC errors in any way:
http://search.cpan.org/~dmaki/JSON-RPC-1.06/lib/JSON/RPC.pm#ERRORS

The only option seems to me to modify the code of the JSON::RPC module itself or to add and eval to each of the methods of Backend PM and catch the errors before JSON RPC does.

While this is feasible it seems to me we should consider if we shall use the JSON::RPC module at all as we no longer use any of its functionalities. We validate the arguments outside of the module. If we also catch all crashes what is the use of JSON::RPC then?

sandoche2k:
@mtoma @mattias-p this issue exists from 2016. Can any of you have a look?

mattias-p:
Well, there is #545. I reviewed that in October last year. I'm not sure what more I can do at this point.

matsduf:
I suggest that we consider this issue to be resolved when #545 is updated and merged.

PR #327, opened 2017-12-14
Fix issue 213

mtoma:


mattias-p:
This looks wrong to me. #213 isn't about line termination. It's about logging internal error details instead of leaking them.

mtoma:
Please read:
http://perldoc.perl.org/functions/die.html

mattias-p:
Aha, now I see! This is great, but it's the solution to a different problem.

My concern in this issue is that we leak any details about the status of the service at all through the RPC-API. When the problem is related to maintenance, just knowing that it's an internal error should be enough for the user. Administrators are interested in the details of course and that's why I suggested logging such errors.

matsduf:
@mtoma, I suggest that you rebase your branch against an updated develop branch to get rid of the conflicts. Or else we risk getting a more complex commit history.

matsduf:
I agree with @mattias-p that much of this messages should go to the log file instead of the API user. In some of the cases the message *should* go to the both, or at least to the API user, such the message "Invalid ns_ip_pair" in  lib/Zonemaster/Backend/TestAgent.pm.

mtoma:
The problem is that the JSON-RPC module converts any crash (perl die) in the code and sends them automatically to the API user. This is a practical solution for a simple API.
If we want to avoid this we need to catch any die message at the RPCAPI.pm level and sent them to some log module.

For me the real issue hare was that we were leaking system information about the PATH whre the software is installed including the linux username.

To fix this properly I think all the existing error messages should be given error codes and only the error codes could be sent to the user and the description to some kind of log.

i think this issue should be merged to avoid leaking system information and possibly a new one created to change the error catching mechanism in a deeper way.

mattias-p:
I think the changes in this PR are good. However they don't fix issue 213. If a new PR is made with the commit message and branch name changed so they don't claim to fix 213, I will approve that. Also at the same time please remove the useless merge commit.

PR #335, opened 2017-12-19
Partial fix of the #213 issue - added \n to all die statements to avoid leaking system path information

mtoma:
Partial fix of the #213 issue - added \n to all die statements to avoid leaking system path informationPartial fix of the #213 issue - added \n to all die statements to avoid leking system path information


matsduf:
@mtoma, rebase is better than merging, because it makes the commit history cleaner.

sandoche2k:
@mtoma merging, since we have consensus

PR #336, opened 2017-12-19
Fix config and start files

matsduf:
Dedicated backend ini files and start files depending on DB. Updated instructions. This should make it possible to support PostgreSQL for Debian and Ubuntu 16.04.


mattias-p:
If the CentOS instruction should have separate configs per database engine, I think the other instructions should follow suite. And backend_config.ini without the database suffix should be removed.

The instructions for the various OSes should follow the same outline as that makes for easier maintenance. I put a fair amount of effort into the outline for Debian and FreeBSD. I believe it would be a nice fit for CentOS as well. I probably should have made the update for CentOS at the same time but I didn't. Do you want to update the CentOS instruction outline, or would you rather I do it? I think it should be done.

sandoche2k:
If you already have it, i would let you update :-)

mattias-p:
Here's a CentOS installation instruction that's been updated to use the same outline as Debian and FreeBSD: https://github.com/mattias-p/zonemaster-backend/blob/centos-installation/docs/Installation.md

This far it's a pure cut-and-paste job. I have not yet tested the instruction.

My view is that the different-configs-for-different-dbs change in this PR is moot with the update outline. The updated outline version still does not have instructions for PostgreSQL.

sandoche2k:
Thanks. I will test it.

matsduf:
The reason that I keep backend_config.ini is that it shows commented options etc.

docs/Installation.md
```diff
@@ -35,156 +35,124 @@ This instruction covers the following operating systems:
 
 ## 1. Installation on CentOS
 
-### 1.1 Installing dependencies (CentOS)
+### 1.1 Install Zonemaster::Backend and related dependencies (CentOS)
+
+Install dependencies available from binary packages:
```
mattias-p: These aren't all binary.

matsduf: Corrected.

matsduf:
@sandoche2k! Can you review this PR? And approve if you approve. Should @mtoma also look at it? It contains no changes in code.

matsduf:
I have now updated the MANIFEST file.

docs/Installation.md
```diff
@@ -35,156 +35,129 @@ This instruction covers the following operating systems:
...
-### 1.4 Directory and file manipulation (CentOS)
+If you have selected **MySQL**, install files to their proper locations:
 
 ```sh
-sudo mkdir /etc/zonemaster
+sudo install -d /etc/zonemaster
+sudo install --mode=755 ./backend_config.ini-mysql /etc/zonemaster/backend_config.ini
+sudo install --mode=755 ./zm-centos.sh-mysql /etc/init.d/zm-centos.sh
 mkdir "$HOME/logs"
```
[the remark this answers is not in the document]
matsduf: Updated.

matsduf:
@mattias-p and @sandoche2k! Please review again.

docs/Installation.md
```diff
@@ -35,156 +35,129 @@ This instruction covers the following operating systems:
...
 ```
 
-### 1.8 Post-installation sanity check (CentOS)
+Check that the service has started:
+
+```sh
+sudo /etc/init.d/zm-centos.sh status
+```
+
+### 1.4 Post-installation sanity check (CentOS)
```
mattias-p: The post-installation sanity check sections are repeated and could be deduplicated and moved to a common section like this: https://github.com/dotse/zonemaster-engine/blob/master/docs/Installation.md#post-installation-sanity-check.

matsduf: @mattias-p, I both agree and disagree. It *could* be moved, but I think the flow of installation would be worse. It is a important part of installation. I do not think that the duplication is such a problem. I suggest that we keep it as is. What do you say, @sandoche2k?

docs/Installation.md
```diff
@@ -216,33 +191,29 @@ Install dependencies not available from binary packages:
...
+#### 2.2.1 Instructions for MySQL (Debian)
+
 Install files to their proper locations:
 
 ```sh
 cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
 sudo install -d /etc/zonemaster
-sudo install --mode=755 ./backend_config.ini /etc/zonemaster/
-sudo install --mode=755 ./zm-backend.sh /etc/init.d/
+sudo install --mode=755 ./backend_config.ini-mysql /etc/zonemaster/backend_config.ini
```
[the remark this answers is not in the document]
matsduf: With your solution both copying the file and editing it with sed (or something else) are required in some instances. I think the solution proposed in my PR gives a cleaner, simpler and more consistent instructions. What is your opinion, @sandoche2k?

When #343 is implemented we can change it again.

docs/Installation.md
```diff
@@ -216,33 +191,29 @@ Install dependencies not available from binary packages:
...
+sudo install --mode=755 ./zm-backend.sh-mysql /etc/init.d/zm-backend.sh
```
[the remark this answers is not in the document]
matsduf: I think we should consider that for the 2018.1 release.

mattias-p:
In response to several comments, the goal I'm trying to achieve with my suggestions is to ease the maintainability of the documentation. Deduplication is an effective means to that end.

PR #306, opened 2017-10-27
Debian installation

mattias-p:

PR #338, opened 2017-12-20
Get CentOS instruction in line with Debian and FreeBSD

mattias-p:
This is the third and final part of the effort started in #300 and continued in #306.
See also #336.


sandoche2k:
@mattias-p seems ok, but will approve once i test them

PR #339, opened 2017-12-20
Zonemaster backend issue 337

vlevigneron:


sandoche2k:
@vlevigneron can you have a look at Mats request above. Also, do you have an idea why there was no issue for the current 'master' version, and not for the current 'develop' branch. Has there been any change?

vlevigneron:
It's worst, I broke the rebase it seems... :-/

vlevigneron:
The easier is to give up that ¨PR and do another one with just new data.

sandoche2k:
closes #339 due to multiple commits together

Issue #337, opened 2017-12-20
Make fails for the backend (Debian 8) with the following results:


sandoche2k:
```
cp lib/Zonemaster/Backend/Translator.pm blib/lib/Zonemaster/Backend/Translator.pm
cp lib/Zonemaster/Backend/Config.pm blib/lib/Zonemaster/Backend/Config.pm
cp lib/Zonemaster/Backend/DB/MySQL.pm blib/lib/Zonemaster/Backend/DB/MySQL.pm
cp lib/Zonemaster/Backend/DB/PostgreSQL.pm blib/lib/Zonemaster/Backend/DB/PostgreSQL.pm
cp lib/Zonemaster/Backend/TestAgent.pm blib/lib/Zonemaster/Backend/TestAgent.pm
cp lib/Zonemaster/Backend.pm blib/lib/Zonemaster/Backend.pm
cp lib/Zonemaster/Backend/DB.pm blib/lib/Zonemaster/Backend/DB.pm
cp lib/Zonemaster/Backend/DB/SQLite.pm blib/lib/Zonemaster/Backend/DB/SQLite.pm
cp lib/Zonemaster/Backend/RPCAPI.pm blib/lib/Zonemaster/Backend/RPCAPI.pm
cp script/zonemaster_backend_rpcapi.psgi blib/script/zonemaster_backend_rpcapi.psgi
/usr/bin/perl "-Iinc" -MExtUtils::MY -e 'MY->fixin(shift)' -- blib/script/zonemaster_backend_rpcapi.psgi
cp script/zonemaster_backend_testagent blib/script/zonemaster_backend_testagent
/usr/bin/perl "-Iinc" -MExtUtils::MY -e 'MY->fixin(shift)' -- blib/script/zonemaster_backend_testagent
sandoche@vps323914:~/zonemaster-backend$ make test
PERL_DL_NONLAZY=1 PERL_USE_UNSAFE_INC=1 /usr/bin/perl "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'inc', 'blib/lib', 'blib/arch')" t/*.t
t/test01.t ................ 3/? Can't locate object method "blessed" via package "External query for afnic.fr, NS attempted to ns1.nic.fr/2001:67c:2218:2::4:1 while running with no_network at /usr/local/share/perl/5.20.2/Zonemaster/Engine/Test/Basic.pm line 276.
[...]

PR #342, opened 2017-12-21
 - Modify data for new engine version.

vlevigneron:
- fixed dotse/zonemaster-backend#337


sandoche2k:
@matsduf, @mattias-p to review

sandoche2k:
@matsduf got to verify which commit has created this issue

matsduf:
The information on the commit has been published on dotse/zonemaster-backend#337

vlevigneron:
@matsduf Yes we'll do our best to understand and a new issue should be raised but as we know how to "fix" it now if we generate a new data file, it should not prevent the merge of last code for the release.

matsduf:
@vlevigneron! If we do not understand why it happens we might introduce a bad bug in Zonemaster. The following line in `t/test01.t` emits the error message:

```
        Zonemaster::Backend::TestAgent->new( { db => "Zonemaster::Backend::DB::SQLite" } )->run( $test_id );
```

vlevigneron:
@matsduf I guess we have to modify data just because we have changed (and fixed) the way undelegated tests work now. Just try this `zonemaster-cli afnic.fr --ns ns1.nic.fr --ns ns2.nic.fr/192.134.4.1` (which is what t/test01.t does) with the 2 environnements, you'll see that the results are differents, DNS queries also.

matsduf:
@vlevigneron! I have no problem with recreating the data as long as we understand why we should do it. It is not only that the test fails. Zonemaster crashes and spits out messages that indicate that something is incompatible. 

If you run the test with just the 72 first lines in the file, then the data file seems to be incomplete for the test and you get the following error message (no crash):

```
t/test01.t .. 1/? External query for fr, NS attempted to a.root-servers.net/198.41.0.4 while running with no_network at /usr/local/share/perl/5.22.1/Zonemaster/Engine/Zone.pm line 206.
```

If you increase that to the first 73 lines you get the crash. 

1. Is the format of the data incorrect? We should have a test of the data to make sure the format is correct.
2. Is the data incomplete? Zonemaster should not crash because of incomplete data.

matsduf:
@mattias-p! Please update your comment! Where do you want to have `use Scalar::Util qw(blessed)`?

mattias-p:
As @matsduf pointed out I messed up my review comment above. Sadly I can't find an edit button for the comment. This is what I meant to say:

The #337 error message complains about not finding `blessed` in `lib/Zonemaster/Backend/TestAgent.pm`. Please `use Scalar::Util qw( blessed )` there.

vlevigneron:
@matsduf We need to recreate the data because engine behaviour has changed for undelegated tests so current recorded data are not what the engine expects with this new behaviour.

vlevigneron:
@matsduf I also guess there is an issue in `t/test01.t` where data file is created twice instead of just once at the end of the script. So the current data file is broken, we need to re-create it.

vlevigneron:
@mattias-p You are right, I'll add that in TestAgent.pm file.

matsduf:
Can you please update t/test01.t too?

vlevigneron:
@matsduf I just did it.

PR #345, opened 2018-01-04
Start script fix

matsduf:
Change user that runs Backend to "zonemaster" (must be created in the system).


sandoche2k:
@matsduf please merge, so that i can start testing the backend

PR #347, opened 2018-01-05
Adjusting the locale environment before the daemons are started

matsduf:
The language setting via JSON RPC does not work if LANGUAGE is set to something, at least in some OSs. With this fix, to be safe some environment variables are unset, and LC_CTYPE is set to the default locale. With this setting it should work, and it will not be in conflict with anything.

The PR is a work-around for the issue described in issue #350, but this should NOT be considered to be a solution.

### DIFF
--- a/share/zm-backend.sh-mysql
+++ b/share/zm-backend.sh-mysql
@@ -14,6 +14,13 @@
 #                    database engine.
 ### END INIT INFO
 
+# Unset potentially conflicting locale, setting the default
+unset LANGUAGE
+unset LANG
+unset LC_MESSAGES
+unset LC_ALL
+export LC_CTYPE="en_US.UTF-8"
+
 LOGDIR=/var/log/zonemaster
 PIDDIR=/var/run/zonemaster
 LISTENIP=127.0.0.1
--- a/share/zm-backend.sh-postgresql
+++ b/share/zm-backend.sh-postgresql
@@ -14,6 +14,13 @@
 #                    database engine.
 ### END INIT INFO
 
+# Unset potentially conflicting locale, setting the default
+unset LANGUAGE
+unset LANG
+unset LC_MESSAGES
+unset LC_ALL
+export LC_CTYPE="en_US.UTF-8"
+
 LOGDIR=/var/log/zonemaster
 PIDDIR=/var/run/zonemaster
 LISTENIP=127.0.0.1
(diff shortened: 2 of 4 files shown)


matsduf:
I followed the instructions for installing under Ubuntu 16.04. When trying directly to the Backend I hade no language support. In the start file I then added "unset LANGUAGE" and restarted Backend (both RPC and Test Agent). Now I hade support for sv and fr (not da due to another bug). I removed "unset LANGUAGE" and restarted and lost support.

The added lines solves the issue and will not create any problem.

We have the same problem in CLI. When LANGUAGE is set to something, then you cannot select language with `--locale LOCALE`.

matsduf:
@mtoma, are you against the update?

share/zm-backend.sh-mysql
```diff
@@ -14,6 +14,13 @@
 #                    database engine.
 ### END INIT INFO
 
+# Unset potentially conflicting locale, setting the default
+unset LANGUAGE
+unset LANG
+unset LC_MESSAGES
+unset LC_ALL
+export LC_CTYPE="en_US.UTF-8"
+
```
mattias-p: This shouldn't be done in a wrapper script but inside the daemon itself.

Is there a reason why you set LC_CTYPE to en_US.UTF-8?

matsduf: I agree with you that it should be done in the JSON RPC daemon itself, but that is too large change to do now. This is a work-around that makes it work.

en_US.UTF-8 is the default locale selected by Backend unless some other language is selected. I have a previous experience that you get problem with Backend unless it starts in a UTF-8 environment.

matsduf: @mattias-p, what changes do you request?

mtoma:
I'm not against but all this LOCALE related stuff is quite complex and in the past a few digit separation issues were fixed in the backend because of Perl JSON module bugs. So just trying to understand a little more what this problem is about.

share/zm-backend.sh-mysql
mattias-p: In that case I request that comments be added about the fact that it's a workaround, and previous problems with non-UTF-8 environments.

mattias-p:
Maybe the problem should be documented as a proper bug report in an issue?

matsduf:
@mtoma and @mattias-p! I have created an issue #350 describing the issue. Is it OK to merge this PR?

mattias-p:
Accepting this as it solves the problem for at least some contexts. Follow-up issue #353 was created to replace this change with a solution that works for all supported contexts.

PR #362, opened 2018-01-15
Deduplicate config files for different database engines

mattias-p:
This makes it easier to maintain backend_config.ini by deduplicating it in the repo.


mattias-p:
@matsduf We're not using the SQLite configuration anyway, so we're not really losing anything by removing it. 

It seems the GEOLOCATION had only gotten into one of four copies of the config file. I added that back. And I also rest my case.

matsduf:
The SQLite information is useful when we enable SQLite. We should save that information. Please undelete backend_config.ini-sqlite so that we have that information at hand.

The GEOLOCATION was kept in the default backend_config.ini.

PR #363, opened 2018-01-15
Unify service scripts for different database engines

mattias-p:
This makes it easier to maintain the service scripts by deduplicating them in the repo.


mattias-p:
@matsduf
1. If you've followed the installation instruction the relevant database will be installed and so it will be started.
2. Having multiple databases installed but not started is a fringe use case. If you're doing that you're an advanced user and you can edit the start script. I can add a note on how to do it.

matsduf:
@mattias-p, I propose that Required-Start/Required-Stop is updated by `sed` in the same way as you have implemented for the Backend ini file.

mattias-p:
@matsduf, What database to use should be declared in a single place. Otherwise it will be possible to create inconsistent configurations. A better solution would be to add a configuration check that (among other things) verifies that it can connect to the database according to the current config.

PR #364, opened 2018-01-19
Fix manifest after 362

mattias-p:


mattias-p:
And I also made it against the master branch. Everything is wrong.

PR #366, opened 2018-01-22
TestAgent: Fail early if database is inaccessible

mattias-p:
Make the start and restart commands of zonemaster_backend_testagent
fail before daemonization if the configured database is
inaccessible. This enables us to make service scripts that properly
report errors when the configured database engine isn't installed and
when there's a problem with the database connection configuration.

If the database connection fails early, zonemaster_backend_testagent
exits with a non-zero status and an error message to stderr.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -20,6 +20,10 @@ else {
 }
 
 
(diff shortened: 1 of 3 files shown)


mattias-p:
Note to self: Make sure the database adapter properly deals with being forked.

Issue #369, opened 2018-01-25
Service startup reaction to unavailable database engine


mattias-p:
Both the TestAgent and RPCAPI services need access to a database for their operation. The service scripts declare weak dependencies on all supported database engine services to ensure that the configured database engine is started if it's installed.

Today if the configured database isn't installed or if it fails to start, the Zonemaster Backend service scripts still report successful startup.

The service scripts must report failure to start if the configured database isn't accessible.

mattias-p:
Design proposal in two steps:
- [ ] Let `zonemaster_backend_testagent` and `zonemaster_backend_rpcapi.psgi` exit with non-zero status if they cannot connect to the database as configured.
- [ ] Let service scripts to detect non-zero exit statuses from `zonemaster_backend_testagent` and `zonemaster_backend_rpcapi.psgi` and act accordingly.

matsduf:
Looks good.

PR #370, opened 2018-01-25
FreeBSD service scripts

mattias-p:
This purpose of this PR is to add separate FreeBSD service scripts for the RPCAPI and TestAgent daemons. Having two single-daemon services instead of one dual-daemon script has two advantages:
* Users get more control.
* The implementation doesn't have to consider complex failure modes where one operation succeeds and the other one fails.

Additional drive-by fixes:
* The paths where log and pid files are written (from $HOME to /var)
* A zonemaster user is created and used (like on CentOS)
* File permissions are tweaked to be more restrictive (and reasonable)


docs/Installation.md
```diff
@@ -514,14 +518,15 @@ psql -U pgsql -f ./initial-postgres.sql template1
...
->
+Install service scripts and start services:
 
 ```sh
-starman --error-log="$HOME/logs/error.log" --pid-file="$HOME/logs/starman.pid" --listen=127.0.0.1:5000 --daemonize /usr/local/bin/zonemaster_backend_rpcapi.psgi 
-zonemaster_backend_testagent start
+install -m 775 -g zonemaster -d /var/log/zonemaster
+install -m 775 -g zonemaster -d /var/run/zonemaster
+install -m 755 ./zm_rpcapi /usr/local/etc/rc.d/
+install -m 755 ./zm_testagent /usr/local/etc/rc.d/
```
matsduf: For the file names, see MANIFEST above.

docs/Installation.md
```diff
@@ -514,14 +518,15 @@ psql -U pgsql -f ./initial-postgres.sql template1
...
+echo 'zm_rpcapi_enable="YES"' | tee -a /etc/rc.conf
+echo 'zm_testagent_enable="YES"' | tee -a /etc/rc.conf
 ```
```
matsduf: I suggest the following lines instead as it is usually written i FreeBDS instructions (e.g. FreeBSD Handbook):

Add the following lines to /etc/rc.conf:
```
zm_rpcapi_enable="YES"
zm_testagent_enable="YES"
```
Then start the daemons:
```
service  zm_rpcapi start
service zm_testagent start
```

share/zm_rpcapi
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+# PROVIDE: zm_rpcapi
+# REQUIRE: NETWORKING mysql postgresql
```
matsduf: What happens if not both of the database engines are installed?

share/zm_testagent
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+# PROVIDE: zm_testagent
+# REQUIRE: mysql postgresql
```
matsduf: What happens if not both of the database engines are installed?

docs/Installation.md
```diff
@@ -425,6 +425,12 @@ cpan -i Zonemaster::Backend
 > The command above might try to install "DBD::Pg" and "DBD::mysql".
 > You can ignore if it fails. The relevant libraries are installed further down in these instructions.
 
+Add `zonemaster` user and group:
+```sh
+pw groupadd zonemaster
+pw useradd zonemaster -g zonemaster -s /sbin/nologin -d /nonexistent -c "Zonemaster daemon user"
```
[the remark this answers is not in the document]
mattias-p: I actually tried that and failed.

share/zm_rpcapi
mattias-p: Regarding the meaning of REQUIRE I quote https://www.freebsd.org/cgi/man.cgi?rcorder(8)#end:

> The "REQUIRE" keyword is misleading: It doesn't describe which daemons have to be running before a script will be started. It describes which scripts must be placed before it in the dependency ordering. For example, if your script has a "REQUIRE" on "named", it means the script must be placed after the "named" script in the dependency ordering, not necessarily that it requires named(8) to be started or enabled.

My idea for dealing with unavailable database engines is described in #369. I believe, thought I haven't thoroughly tested it, that the service script part of #369 is implemented automagically in these scripts by virtue of the rc.d framework.

share/zm_testagent
mattias-p: Ditto.

docs/Installation.md
mattias-p: The `echo | tee -a` lines have the advantage that they are cut-and-pasteable, and I reckon we'll be adding those lines a lot.

matsduf: The `echo | tee -a` line assumes that rc.conf ends with a line ending. It also adds that line even if not needed (already there). Here is a new version:

```
Run the following command turn the daemons on in /etc/rc.conf:

sysrc zm_rpcapi_enable="YES"
sysrc zm_testagent_enable="YES"
```

mattias-p: Neat! That's much better.

mattias-p:
@matsduf Can you take another look at this?

docs/Installation.md
```diff
@@ -514,14 +518,22 @@ psql -U pgsql -f ./initial-postgres.sql template1
...
->
+Install service scripts:
+
+```sh
+install -m 775 -g zonemaster -d /var/log/zonemaster
+install -m 775 -g zonemaster -d /var/run/zonemaster
+install -m 755 ./zm_rpcapi-bsd /usr/local/etc/rc.d/zm_rpcapi
+install -m 755 ./zm_testagent-bsd /usr/local/etc/rc.d/zm_testagent
+echo 'zm_rpcapi_enable="YES"' | tee -a /etc/rc.conf
+echo 'zm_testagent_enable="YES"' | tee -a /etc/rc.conf
```
matsduf: I propose the following instead:

sysrc zm_rpcapi_enable="YES"
sysrc zm_testagent_enable="YES"

docs/Installation.md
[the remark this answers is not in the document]
matsduf: Ok.

docs/Installation.md
mattias-p: Sorry, I forgot that.. Addressed it now.

PR #372, opened 2018-02-02
Deduplicate instructions

mattias-p:
This is another step in my deduplication effort regarding installation-related materials.
Additional drive-by updates:
 * ~Grouped together miscellaneous topics under a common topic at the end of the document.~ (Moving to a future PR)
 * Gathered up some directory creation and file copying on FreeBSD.
 * Replaced some `echo | tee -a` commands with `sysrc` on FreeBSD.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -199,6 +199,21 @@ Check that the service has started:
 sudo /etc/init.d/zm-centos.sh status
 ```
 
+### 1.4 Post-installation sanity check (CentOS)
+
+If you followed this instructions to the letter, you should be able to use the
+API on localhost port 5000, like this:
+
+```sh
+curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
+```
+
+The command is expected to give an immediate JSON response similiar to :
+
+```json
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -536,10 +481,18 @@ service zm_rpcapi start
...
-### 3.4 Post-installation sanity check (FreeBSD)
 
-If you followed this instructions to the letter, you should be able to use the
-API on localhost port 5000, like this:
+## 4. Installation on Ubuntu
+
+Use the procedure for installation on [Debian](#2-installation-on-debian).
+
+
+## Post-installaiton
```
matsduf: "installaiton" -> "installation"

mattias-p:
There are lots of ways removing duplicated improves navigability, as well as maintainability and is generally a good thing. If you like I can enumerate a bunch of arguments.

The overview section is already a kind of a TOC. Replacing it with a bullet list would probably be an improvement.

I'm not a fan of the numbering. It messes with hyperlinks. Also we do have the OS suffixes in all the OS specific heading.

I'll just go ahead and push an update along these lines.

matsduf:
@mattias-p, we need a document that is easy to navigate in.

I remember that numbering created problems with the links, but then I suggest removing the numbers from all headlines. I.e. make it consistent.

With your solution we need a table of contents at the start of the document pointing at the level one and level two headlines (level two and level three from a markdown perspective).

mattias-p:
On second thought. Lets keep the discussion of the outline separate from this thread. I reduced the contents of this PR to only deal with file copying and sysrc.

PR #374, opened 2018-02-05
Numbered headlines and TOC

matsduf:
Updated installation instructions with numbered headlines and a table of contents in the beginning of the file to make it easy to find relevant section and to get an overview.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -28,9 +28,7 @@
 
 ## 1. Overview
 
-This document describes prerequisites, installation, configuration, startup and
-post-install sanity checking for Zonemaster::Backend. The final section wraps up
-with a few pointer to interfaces for Zonemaster::Backend. For an overview of the
+This document contains all steps needed to install Zonemaster::Backend. For an overview of the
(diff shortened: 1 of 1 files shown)


mattias-p:
I don't like the idea of having to maintain such a table by hand. I'm in favor if there are good tools for generating the TOC and for checking those links. Otherwise I'm against.

matsduf:
@mattias-p, I just created the TOC and it was not complicated. The editor provides the link target and validates the links. For Backend it think it would be an improvement.

@sandoche2k, what do you think?

mattias-p:
@matsduf I'm worried that we'll forget to update the TOC when we update the document, and that it'll be bothersome to check whether we've forgotten or not. Apart from that I like the TOC. Oh, and with the TOC in place most of the text in the "Overview" section could be removed as it's just an enumeration of what you'll find in the document.

matsduf:
@mattias-p, I think there is no risk that we will forget since the TOC sits in the top of the document. I was not cumbersome to create the TOC. Yes, we could reduce the "Overview". I have included a proposal.

@sandoche2k, can you review?

PR #379, opened 2018-02-16
Deduplicate LSB service scripts across OSes

mattias-p:
All of CentOS, Debian and Ubuntu use LSB for service configuration. Today we have a separate service script for CentOS and another one shared between Debian and Ubuntu. The two scripts differ in two ways:
* The CentOS script starts the daemons as a zonemaster-specific user while Debian/Ubuntu starts them as www-data.
* The CentOS script specifies the starman executable with an absolute path (because it's not in the PATH of the root user - presumably because it was installed from CPAN). Debian/Ubuntu has its starman executable in the PATH.

This PR:
* Updates the Debian/Ubuntu script to work on CentOS as well.
* Adopts the zonemaster user for daemons on Debian/Ubuntu (following CentOS and FreeBSD).
* Updates the CentOS and Debian/Ubuntu instructions to work with the updated common script.
* Removes the old CentOS-specific script.


matsduf:
I find this "deduplication" less important than separating the two daemons in separate scripts as with FreeBSD.

mattias-p:
It's easier and to me it makes more sense to combine these two OS-scripts into one before splitting the single script into two daemon-scripts, rather than to split these two OS-scripts into four OS-daemon-scripts before combining those into two daemon-scripts.

mattias-p:
@sandoche2k Can you review this PR or otherwise indicate that you don't intend to? I'd like to make another PR that overlaps these changes but I'd rather not until this PR has been merged or closed.

sandoche2k:
@mattias-p if you want me to approve it, i need to test with a new installation, which i think will not be possible this week.

mattias-p:
My view is that this change is fairly low-risk. I've successfully installed this from scratch on CentOS 7 and Debian 9, and I made the initial testing on a Debian 8 system that was not clean.

PR #382, opened 2018-02-20
Corrected error message for IDN

matsduf:
* Corrected error message when a non-ascii domain label cannot be converted to A-label, i.e. it is not a valid IDNA label.
* Corrected error message when there is no support for non-ascii domain labels, i.e. zonemaster-ldns is not built with libidn support.


sandoche2k:
@mattias-p and @mtoma to review

sandoche2k:
@matsduf please merge

PR #394, opened 2018-03-07
Installation instructions

matsduf:
Now that the API test has been decoupled from the OS installations, the phrasing needs an update. "Sanity" check does not feel appropriate in these instructions. It is just a test that it works, not a test of sanity.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -21,7 +21,7 @@
   * [5.4 Post-installation (FreeBSD)](#54-post-installation-freebsd)
 * [6. Installation on Ubuntu](#6-installation-on-ubuntu)
 * [7. Post-installation](#7-post-installation)
-  * [7.1 Smoke test of Zonemaster::Backend](#71-smoke-test-of-zonemasterbackend)
+  * [7.1 Smoke test](#71-smoke-test)
   * [7.2 What to do next?](#72-what-to-do-next)
   * [7.3 Cleaning up the database](#73-cleaning-up-the-database)
 
@@ -47,7 +47,7 @@ For details on supported versions of Perl, database engine and operating system
 for Zonemaster::Backend, see the [declaration of prerequisites].
 
 > **Note:** In addition to the normal dependencies, the post-installation
-> check instruction assumes that you have curl installed.
+> smoke test instruction assumes that you have curl installed.
 
 ## 3. Installation on CentOS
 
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -520,10 +520,11 @@ Use the procedure for installation on [Debian](#2-installation-on-debian).
...
 ## 7. Post-installation
 
-### 7.1 Sanity check
+### 7.1 Check of Zonemaster::Backend
 
-If you followed this instructions to the letter, you should be able to use the
-API on localhost port 5000, like this:
+If you have followed the installation instructions for Zonemaster::Backend and chosen database engine 
+for your OS above, you should be able to use the
+API on localhost port 5000 as below. The command requires that `curl` is installed.
```
mattias-p: These are more words, but do they really add anything?

docs/Installation.md
```diff
@@ -21,7 +21,7 @@
   * [5.4 Post-installation (FreeBSD)](#54-post-installation-freebsd)
 * [6. Installation on Ubuntu](#6-installation-on-ubuntu)
 * [7. Post-installation](#7-post-installation)
-  * [7.1 Sanity check](#71-sanity-check)
+  * [7.1 Check of Zonemaster::Backend](#71-check-of-zonemasterbackend)
```
mattias-p: "Check" is a bit vague for a heading. "Sanity check" gives more context. "Smoke test" might be an alternative if you dislike the "sanity" word.
https://en.wikipedia.org/wiki/Sanity_check#Software_development
http://www.innovateus.net/innopedia/what-does-sanity-check-mean
https://en.wikipedia.org/wiki/Smoke_testing_(software)

docs/Installation.md
```diff
@@ -520,10 +520,11 @@ Use the procedure for installation on [Debian](#2-installation-on-debian).
 
 ## 7. Post-installation
 
-### 7.1 Sanity check
+### 7.1 Smoke test of Zonemaster::Backend
```
mattias-p: Do we really need to spell out Zonemaster::Backend in the heading? I think that is pretty much implied considering which document this is.

docs/Installation.md
```diff
@@ -46,7 +46,7 @@ Zonemaster::Engine installation].
 For details on supported versions of Perl, database engine and operating system
 for Zonemaster::Backend, see the [declaration of prerequisites].
 
-> **Note:** In addition to the normal dependencies, the post-installation sanity
+> **Note:** In addition to the normal dependencies, the post-installation
```
mattias-p: Call it smote test here too for consistency?

Issue #340, opened 2017-12-21
Logging in Test Agent


mattias-p:
The zonemaster_backend_testagent service sometimes encounters errors such as configuration errors and crashed tests. The daemon part of zonemaster_backend_testagent has no means of communicating what happened to sys admins and developers. The zonemaster_backend_testagent service must write available details about crashed tests and configuration errors to a log file.

mattias-p:
I have a proof of concept for this. Will PR shortly.

matsduf:
Probably the best solution is to send log messages to syslog to use the normal infrastructure on a Unix/Linux server.

sandoche2k:
@mattias-p can you update the PR

mattias-p:
@sandoche2k No, as we discussed I cannot prioritize this PR right now. I'll gladly do it for the next release. If you want it for this release, feel free to assign someone else.

mtoma:
This issue will be partially solved with pull request #398 
The merged code is incomplete and needs to be completed (basically find and replace all eval blocks and add the new logging mechanism)

matsduf:
Solved #398.

PR #398, opened 2018-04-05
Logging in Test Agent

mattias-p:
This is a stale work in progress. For the time being I won't be able to put any more work into it. Feel free to pick it up and bring it over the finish line.

This PR outlines a way to get robust logging. The commits themselves are quite concise. However I don't think all of them are relevant. The USER and BASEDIR commits are only there to improve my own development setup and I'm not sure if they make sense anymore.

See #340.

**Edit:** Added link to issue.


sandoche2k:
@mtoma will have a look at this PR

sandoche2k:
@mattias-p can you check the conflicts?

matsduf:
I would suggest rebasing the branch.

mattias-p:
I thought @mtoma was taking over this. Are you giving it back to me? Of course I have no problem with that. Not sure when I'll be able to fit the time though. Do you want me to close this PR in the mean time? I only created it to illustrate the general ideas I started out with before I abandoned my effort, and you guys opined that a plain branch was insufficient.

mattias-p:
Rebased the changes but have not yet tested them (at least some conflicts were resolved).
I did notice however that some dependencies are introduced on modules that aren't packaged for all of the OSes.

mattias-p:
@mtoma @matsduf @pamasse I've updated this PR so it works with the current develop branch. Please re-review.

share/zm-backend.sh
```diff
@@ -37,7 +37,7 @@ start() {
 stop() {
     if [ -f $PIDDIR/zonemaster_backend_testagent.pid ]
     then
-        $BASEDIR/bin/zonemaster_backend_testagent --user=$USER --group=$GROUP --pidfile=$PIDDIR/zonemaster_backend_testagent.pid stop
+        $BASEDIR/bin/zonemaster_backend_testagent --logfile=$LOGDIR/zonemaster_backend_testagent.log --user=$USER --group=$GROUP --pidfile=$PIDDIR/zonemaster_backend_testagent.pid stop
```
matsduf: Is this really correct? I would expect the log file to be stated at start, not at stop.

mattias-p: It looks a bit weird but it is in fact correct. From what I can tell we're not making use of it at the moment, but Daemon::Control (which we're using) provides hooks for the stop command, and if add logging in such a hook, we want the log lines to end up in the correct log file.
However, of course the logfile argument should be included in the start command as well. So I added that.

script/zonemaster_backend_testagent
```diff
@@ -27,12 +33,39 @@ BEGIN {
...
     'group=s' => \$group,
+    'logfile=s' => \$logfile,
+    'loglevel=s' => \$loglevel,
 );
 $pidfile //= '/tmp/zonemaster_backend_testagent.pid';
+$logfile //= '/var/log/zonemaster/zonemaster_backend_testagent.log';
+$loglevel //= 'info';
+$loglevel = lc $loglevel;
+
+$loglevel =~ /^(?:trace|debug|info|inform|notice|warning|warn|error|err|critical|crit|fatal|alert|emergency)$/ or die "Error: Unrecognized --loglevel $loglevel\n";
```
matsduf: I propose some documentation, POD, in this script that describes log settings.

PR #402, opened 2018-04-11
Fix issue 302

mtoma:


sandoche2k:
@matsduf can you make the commit on the develop branch?

matsduf:
@sandoche2k and @mtoma, the conflicts will be resolved if the change in Config.pm is applied on a fresh copy of the develop branch. That is a better solution than adding commits resolving the conflicts.

matsduf:
@mtoma, did you create the branch from an updated copy of the develop branch? The stuff seems to be before last release. The first thing I do is to update all local branches and the branches on my Github fork.

PR #430, opened 2018-05-14
Updated the install instructions

sandoche2k:


docs/Installation.md
```diff
@@ -291,6 +291,19 @@ Initialize the database:
 mysql --user=root --password < ./initial-mysql.sql
 ```
 
+For latest versions of MySQL, wherein the root password is blank and you get an
+error "Access denied for the above command", follow the below procedure:
```
mattias-p: Do you get this from a clean install? If so, on which OS versions? I don't believe I've seen this behavior.

docs/Installation.md
```diff
@@ -527,7 +540,7 @@ you should be able to use the
 API on localhost port 5000 as below. The command requires that `curl` is installed.
 
 ```sh
-curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
```
mattias-p: Both strings and numbers are allowed here by the JSON-RPC specification. If nothing else, using a number is two slightly shorter.

docs/Installation.md
```diff
@@ -527,7 +540,7 @@ you should be able to use the
 API on localhost port 5000 as below. The command requires that `curl` is installed.
 
 ```sh
-curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
+curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":1}' http://localhost:5000/ && echo
```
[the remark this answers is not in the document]
sandoche2k: confirmed with @pamasse that it is an integer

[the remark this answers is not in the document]
sandoche2k: Doesn't work for me "1"
```
sandoche@vps323914:/usr/local/share/perl/5.20.2/auto/share/dist/Zonemaster-Backend$ curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
{"error":{"data":"/id: Expected integer - got string.\n","message":"The JSON sent is not a valid request object.","code":"-32600"},"jsonrpc":"2.0","id":null}
```

docs/Installation.md
```diff
@@ -228,22 +228,10 @@ See the [post-installation] section for post-installation matters.
...
-```sh
-sudo apt-get install libclass-method-modifiers-perl libconfig-inifiles-perl libdata-dump-perl libdbd-sqlite3-perl libdbi-perl libfile-sharedir-perl libfile-slurp-perl libhtml-parser-perl libintl-perl libio-captureoutput-perl libjson-pp-perl libmoose-perl libplack-perl librole-tiny-perl librouter-simple-perl libstring-shellquote-perl libtest-requires-perl libtest-warn-perl libtext-microtemplate-perl libtie-simple-perl starman
-```
-
-Install dependencies not available from binary packages:
-
-```sh
-sudo cpan -i Daemon::Control JSON::RPC Net::IP::XS Parallel::ForkManager Plack::Middleware::Debug
-```
-
```
matsduf: We can have two approaches:

1. Install as much of the Perl dependencies as binary packages.
2. Let cpan install all Perl dependencies.

I think we have followed principle 1 so far. Now you have removed installation of Perl dependencies as binary packages. @sandoche2k, can you explain why?

docs/Installation.md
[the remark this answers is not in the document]
matsduf: That could be a change in Backend in develop branch. If so this would be a breaking change.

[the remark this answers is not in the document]
mattias-p: I created #431 for this.

docs/Installation.md
sandoche2k: AFAIK we had to install the dependencies since we could install the zonemaster module without pre-installing them. When we had to install the dependencies, we had the two approaches that you mentioned above.

Now with CPANM if it is installing by itself the required dependencies, i do not understand why we should pre-install them.

docs/Installation.md
sandoche2k: Yes Debian 9

docs/Installation.md
mattias-p: From the top of my mind I can see two good reasons why we should choose OS packages whenever we can.
* To be reasonably sure that people don't install newer versions than the ones we tested with during installation testing.
* To make sure Zonemaster works with the packaged versions of libraries for the OSes we support. This is important since we want OS packaging for Zonemaster.

sandoche2k: >> whenever we can
That is the question i have too. Either we install everything from OS packages or everything from CPAN. 
Until now, we had the installation instructions wherein if there is a module to be installed, then we look at the binary packages. If we do not get them, then we install them using CPAN. It was adhoc. 
Now by letting CPANM installing them and reducing the burden for the user, we have new requirements. I think we should be coherent on what we want?

matsduf: I think there are pros and cons. By using the CPAN approach the installation instructions will be simpler. If we follow that approach I think that we should take note of the versions of as many libraries as possible to make sure that Makefile.PL is updated with what version that is required. To make it simple for us, require the lowest version that we have found to work.

I guess that the con side is that there is not much help for packaging Zonemaster.

mattias-p: While it is possible to lock down the versions of our dependencies in Makefile.PL, I don't think that's the solution we're looking for. We'd have to do specify the versions of all transitive dependencies rather than just the direct dependencies like we do now.

docs/Installation.md
```diff
@@ -291,6 +294,20 @@ Initialize the database:
 mysql --user=root --password < ./initial-mysql.sql
 ```
 
+For latest versions of MySQL, wherein the root password is blank and you get an
+error "Access denied for the above command" on running the above command ,
+follow the below procedure:
```
matsduf: @sandoche2k, have you tried my suggestion? I think the answer is simpler than that. If the password is blank I think you can connect to the database by omitting the password switch:

```sh
 mysql --user=root < ./initial-mysql.sql
 ```

File: docs/Installation.md (hunk 1 above)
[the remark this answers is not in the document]
Comment (sandoche2k) [reply docs/Installation.md]: Your suggestion did not work 

```sh
sandoche@vps323915:/usr/local/share/perl/5.24.1/auto/share/dist/Zonemaster-Backend$ mysql --user=root < ./initial-mysql.sql
ERROR 1698 (28000): Access denied for user 'root'@'localhost'
```

sandoche2k: @mattias-p I have tested thrice now. I am getting the same results. 

Did you test?

PR #435, opened 2018-06-01
Update sed commands on FreeBSD

mattias-p:
FreeBSD sed doesn't support the `\b` word boundary syntax.
FreeBSD sed -i option requires an argument.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -436,7 +436,7 @@ of operating system version and database engine version is supported.
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sed -i '/\bengine\b/ s/=.*/=MySQL/' /usr/local/etc/zonemaster/backend_config.ini
+sed -i '/[[:<:]]engine[[:>:]]/ s/=.*/=MySQL/' /usr/local/etc/zonemaster/backend_config.ini
 ```
 
 Install, configure and start database engine (and Perl bindings):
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -436,7 +436,7 @@ of operating system version and database engine version is supported.
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sed -i '/\bengine\b/ s/=.*/=MySQL/' /usr/local/etc/zonemaster/backend_config.ini
+sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/=MySQL/' /usr/local/etc/zonemaster/backend_config.ini
```
matsduf: Instead of an meaningless `-i` use `-n` to surpress unneeded printout.

docs/Installation.md
```diff
@@ -467,7 +467,7 @@ mysql < ./initial-mysql.sql
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sed -i '/\bengine\b/ s/=.*/=PostgreSQL/' /usr/local/etc/zonemaster/backend_config.ini
+sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/=PostgreSQL/' /usr/local/etc/zonemaster/backend_config.ini
```
matsduf: Instead of an meaningless `-i` use `-n` to surpress unneeded printout.

docs/Installation.md
mattias-p: The `-i` is not meaningless. It makes sed edit the file in-place, which is what's intended. Replacing the `-i ''` with `-n` would, on the other hand, make the command into a noop.

docs/Installation.md
mattias-p: See above.

docs/Installation.md
matsduf: You are right. For some reason the suffix must be set in FreeBSD `sed`. To make the command consistent between different OSs we should use Perl instead. Perl will always be available so it would not be a problem.

PR #442, opened 2018-06-25
Make file locations overridable

mattias-p:
This is helpful when you want to start a development version of the service.


sandoche2k:
Will it be part of next release ?

mattias-p:
No. It's better to include it in the next one.

sandoche2k:
Yes, that was my question too

matsduf:
Yes, not this release.

PR #461, opened 2018-12-14
Fix code part of issue #405 - use the new profiles feature of the Zonemaster-Engine

mtoma:
Fixes issue #405
Adds the profiles feature to the backend.
Changes the way the config file is read to allow passing it as a parameter to the test agent.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -24,21 +24,14 @@ else {
 
 =cut
 
-sub new {
+sub load_config {
     my ( $class, $params ) = @_;
     my $self = {};
     
-    $self->{cfg} = _load_config();
+    $self->{cfg} = Config::IniFiles->new( -file => $path );
+    die "UNABLE TO LOAD $path ERRORS:[".join('; ', @Config::IniFiles::errors)."] \n" unless ( $self->{cfg} );
     bless( $self, $class );
     return $self;
-}
-
-
-sub _load_config {
-    my $cfg = Config::IniFiles->new( -file => $path );
-    die "UNABLE TO LOAD $path ERRORS:[".join('; ', @Config::IniFiles::errors)."] \n" unless ( $cfg );
-
-    return $cfg;
 }
 
 sub BackendDBType {
@@ -160,17 +153,17 @@ sub force_hash_id_use_in_API_starting_from_id {
 }
 
 sub ReadProfilesInfo {
-	my ($self) = @_;
-	
-	my $profiles;
-	foreach my $public_profile ($self->{cfg}->Parameters('PUBLIC PROFILES')) {
-		$profiles->{lc($public_profile)}->{type} = 'public';
-		$profiles->{lc($public_profile)}->{profile_file_name} = $self->{cfg}->val('PUBLIC PROFILES', $public_profile);
+    my ($self) = @_;
+    
+    my $profiles;
+    foreach my $public_profile ($self->{cfg}->Parameters('PUBLIC PROFILES')) {
+        $profiles->{lc($public_profile)}->{type} = 'public';
+        $profiles->{lc($public_profile)}->{profile_file_name} = $self->{cfg}->val('PUBLIC PROFILES', $public_profile);
     }
 
-	foreach my $public_profile ($self->{cfg}->Parameters('PRIVATE PROFILES')) {
-		$profiles->{lc($public_profile)}->{type} = 'private';
-		$profiles->{lc($public_profile)}->{profile_file_name} = $self->{cfg}->val('PRIVATE PROFILES', $public_profile);
+    foreach my $private_profile ($self->{cfg}->Parameters('PRIVATE PROFILES')) {
+        $profiles->{lc($private_profile)}->{type} = 'private';
+        $profiles->{lc($private_profile)}->{profile_file_name} = $self->{cfg}->val('PRIVATE PROFILES', $private_profile);
     }
     
     return $profiles;
@@ -193,7 +186,7 @@ The adapter connects to the database before it is returned.
 =head3 INPUT
 
 The database adapter class is selected based on the return value
-of L<Zonemaster::Backend::Config->new()->BackendDBType()>. The database
+of L<Zonemaster::Backend::Config->load_config()->BackendDBType()>. The database
 adapter class constructor is called without arguments and is expected
 to configure itself according to available global configuration.
 
@@ -219,7 +212,7 @@ A configured L<Zonemaster::Backend::DB> object.
 
 sub new_DB {
     # Get DB type from config
-    my $dbtype = Zonemaster::Backend::Config->new()->BackendDBType();
+    my $dbtype = Zonemaster::Backend::Config->load_config()->BackendDBType();
     if (!defined $dbtype) {
         die "Unrecognized DB.engine in backend config";
     }
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -24,6 +24,16 @@ else {
 
 =cut
 
+sub new {
+    my ( $class, $params ) = @_;
+    my $self = {};
+    
+    $self->{cfg} = _load_config();
+    bless( $self, $class );
+    return $self;
+}
+
+
 sub _load_config {
     my $cfg = Config::IniFiles->new( -file => $path );
     die "UNABLE TO LOAD $path ERRORS:[".join('; ', @Config::IniFiles::errors)."] \n" unless ( $cfg );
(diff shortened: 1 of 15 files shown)


lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -124,29 +136,14 @@ sub run {
...
-            }
+        $params->{profile} = lc($params->{profile});
+        if (defined $self->{profiles}->{$params->{profile}} && $self->{profiles}->{$params->{profile}}->{zm_profile}) { 
+            my $profile = Zonemaster::Engine::Profile->default;
+            $profile->merge( $self->{profiles}->{$params->{profile}}->{zm_profile} );
+            Zonemaster::Engine::Profile->effective->merge( $profile );
         }
         else {
-            die "Unknown test configuration: [$params->{config}]\n"
+            die "Can't load prfile: [$params->{profile}]" if ($params->{profile} ne 'default')
```
vlevigneron: "Can't load prOfile"

lib/Zonemaster/Backend/Config.pm
```diff
@@ -32,143 +42,144 @@ sub _load_config {
...
 
 sub LogDir {
-    my $cfg = _load_config();
+    my ($self) = @_;
 
-    return $cfg->val( 'LOG', 'log_dir' );
+    return $self->{cfg}->val( 'LOG', 'log_dir' );
 }
 
 sub PerlIntereter {
```
vlevigneron: Typo => PerlInterPReter

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -60,12 +72,12 @@ sub run {
     $domain = $self->to_idn( $domain );
 
     if (defined $params->{ipv4} || defined $params->{ipv6}) {
-        Zonemaster::Engine->config->get->{net}{ipv4} = ( $params->{ipv4} ) ? ( 1 ) : ( 0 );
-        Zonemaster::Engine->config->get->{net}{ipv6} = ( $params->{ipv6} ) ? ( 1 ) : ( 0 );
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, ( $params->{ipv4} ) ? ( 1 ) : ( 0 ) );
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, ( $params->{ipv4} ) ? ( 1 ) : ( 0 ) );
```
vlevigneron: $params->{ipv6}

lib/Zonemaster/Backend/Config.pm
```diff
@@ -32,143 +42,144 @@ sub _load_config {
...
-	my ($slef, $section, $param_name) = @_;
+sub ReadProfilesInfo {
+	my ($self) = @_;
 	
-    my $cfg = _load_config();
+	my $profiles;
+	foreach my $public_profile ($self->{cfg}->Parameters('PUBLIC PROFILES')) {
+		$profiles->{lc($public_profile)}->{type} = 'public';
+		$profiles->{lc($public_profile)}->{profile_file_name} = $self->{cfg}->val('PUBLIC PROFILES', $public_profile);
+    }
```
mattias-p: Could you update the indentation to be consistent with regard to spaces/tabs?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -32,143 +42,144 @@ sub _load_config {
...
 
-    return $cfg->val( $section, $param_name );
+	foreach my $public_profile ($self->{cfg}->Parameters('PRIVATE PROFILES')) {
```
mattias-p: Did you forget to change the name of the loop variable here?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -24,6 +24,16 @@ else {
 
 =cut
 
+sub new {
```
mattias-p: Have you considered giving this method a name that indicates that it actually fetches values from disk? E.g. `load`.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -60,12 +72,12 @@ sub run {
...
-        Zonemaster::Engine->config->get->{net}{ipv4} = ( $params->{ipv4} ) ? ( 1 ) : ( 0 );
-        Zonemaster::Engine->config->get->{net}{ipv6} = ( $params->{ipv6} ) ? ( 1 ) : ( 0 );
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, ( $params->{ipv4} ) ? ( 1 ) : ( 0 ) );
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, ( $params->{ipv6} ) ? ( 1 ) : ( 0 ) );
     }
     else {
-        Zonemaster::Engine->config->get->{net}{ipv4} = 1;
-        Zonemaster::Engine->config->get->{net}{ipv6} = 1;
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 1 );
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 1 );
```
mattias-p: Backend should either respect the `net.ipv4` and `net.ipv6` profile properties, or they shouldn't be part of the profile at all. Allowing the user to specify those values in their profile files just so Backend can ignore them isn't helpful in any way.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -124,29 +136,14 @@ sub run {
...
-            }
+        $params->{profile} = lc($params->{profile});
+        if (defined $self->{profiles}->{$params->{profile}} && $self->{profiles}->{$params->{profile}}->{zm_profile}) { 
+            my $profile = Zonemaster::Engine::Profile->default;
+            $profile->merge( $self->{profiles}->{$params->{profile}}->{zm_profile} );
+            Zonemaster::Engine::Profile->effective->merge( $profile );
         }
         else {
-            die "Unknown test configuration: [$params->{config}]\n"
+            die "Can't load profile: [$params->{profile}]" if ($params->{profile} ne 'default')
```
mattias-p: This message is somewhat misleading since no profile loading is attempted at this point. Instead it should probably say something along the lines of "Unknown profile".

mattias-p: If the `default` profile name is requested when it's not explicitly specified in the configuration, the Engine default profile should be loaded. Unless I'm missing something, that's not being done.

script/zonemaster_backend_testagent
```diff
@@ -20,6 +20,8 @@ BEGIN {
 	$ENV{PERL_JSON_BACKEND} = 'JSON::PP';
 }
 
+$|++;
```
mattias-p: Have you considered writing this like `use English; $OUTPUT_AUTOFLUSH = 1;` and/or adding an explanatory comment? That's far more helpful to both the non-native Perl speaker and search engines.

script/zonemaster_backend_testagent
```diff
@@ -84,14 +87,15 @@ sub main {
     my $self = shift;
 
     my $db = $self->config->{db};
+    Zonemaster::Backend::TestAgent->new({ config => $config });
```
mattias-p: Have you considered adding a comment explaining the purpose of this line?

mattias-p:
Please add a proper link to the referenced issue in the PR description.

lib/Zonemaster/Backend/TestAgent.pm
mtoma: I don't see any place in CLI.pm of the cli interface where the default profile would be explicitly loaded. I assumed it is loaded when the Engine is created. 
Am I wrong?

matsduf: @mtoma, correct. The Engine default profile is always loaded. Explicitly loading the default is not needed from a CLI or Backend point of view.

mattias-p: I was assuming that TestAgent workers were being reused for multiple tests, but now I can't find any basis for that.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -24,151 +24,167 @@ else {
...
+    my ($self) = @_;
+    
+    my $profiles = [];
+    foreach my $public_profile ($self->{cfg}->Parameters('PUBLIC PROFILES')) {
+        $profiles->{lc($public_profile)}->{type} = 'public';
+        $profiles->{lc($public_profile)}->{profile_file_name} = $self->{cfg}->val('PUBLIC PROFILES', $public_profile);
+    }
 
-    return $cfg->val( $section, $param_name );
+    return @$profiles;
```
matsduf: Will this just return an array of public profile names and nothing else?

pamasse: Fix in commit cda24a7 
It will return just the profile names.

sandoche2k:
@mtoma travis fails

Issue #471, opened 2019-01-22
Invalid Zonemaster::Backend::Config call


mtoma:
One Zonemaster::Backend::Config invalid call was left.

PR #472, opened 2019-01-22
Fix issue 471

mtoma:
Fixes issue #471


sandoche2k:
@mtoma i am merging it

PR #482, opened 2019-02-06
Memory leaks

vlevigneron:


vlevigneron:
@mtoma Any idea why it fails only with SQLite ? Can we merge anyway ?

mtoma:
No idea. Some problem with the test file most likely.
But merging this fix is more important than the issue with SQLite.
So I would say go head and merge.

----- Mail original -----
De: "vlevigneron" <notifications@github.com>
À: "zonemaster/zonemaster-backend" <zonemaster-backend@noreply.github.com>
Cc: "Michal TOMA" <michal.toma@afnic.fr>, "Mention" <mention@noreply.github.com>
Envoyé: Mercredi 6 Février 2019 12:45:40
Objet: Re: [zonemaster/zonemaster-backend] Memory leaks (#482)

@mtoma Any idea why it fails only with SQLite ? Can we merge anyway ?

PR #484, opened 2019-02-07
Provide configure-time-deps for Zonemaster::LDNS

mattias-p:
I expect Travis to start failing for Backend once zonemaster/zonemaster-ldns#72 is merged. This PR should pass both before and after merging zonemaster/zonemaster-ldns#72.


mattias-p:
This PR passes its checks for MySQL and PostgreSQL as expected and fails its checks for SQLite (just like #482 as expected). Until this is merged, all other PRs will fail just like #485.

Issue #486, opened 2019-02-08
Sqlite fails


matsduf:
There's a problem in the develop branch that causes Travis to fail in all PRs to the Backend develop branch. 
It was first spotted in #482 but that PR was merged anyway since it was more important to fix the memory leak.
It only occurs when running tests against SQLite. Both MySQL and PostgreSQL run fine.
It seems to be independent of Perl versions.

Here's what it looks like:
```
t/test_validate_syntax.t .. ok
t/test01.t ................ 1/? Use of uninitialized value in -e at /home/travis/build/zonemaster/zonemaster-backend/build_dir/blib/lib/Zonemaster/Backend/TestAgent.pm line 41.
t/test01.t ................ 11/? Use of uninitialized value in concatenation (.) or string at /home/travis/build/zonemaster/zonemaster-backend/build_dir/blib/lib/Zonemaster/Backend/RPCAPI.pm line 387.
Use of uninitialized value in -e at /home/travis/build/zonemaster/zonemaster-backend/build_dir/blib/lib/Zonemaster/Backend/TestAgent.pm line 41.
External query for afnic.fr, SOA attempted to m.root-servers.net/202.12.27.33 while running with no_network at /home/travis/build/zonemaster/zonemaster-backend/build_dir/blib/lib/Zonemaster/Engine/Recursor.pm line 182.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 9 just after 22.
t/test01.t ................ Dubious, test returned 9 (wstat 2304, 0x900)
All 22 subtests passed 
Test Summary Report
-------------------
t/test01.t              (Wstat: 2304 Tests: 22 Failed: 0)
  Non-zero exit status: 9
  Parse errors: No plan found in TAP output
Files=2, Tests=49,  5 wallclock secs ( 0.02 usr  0.01 sys +  4.62 cusr  0.13 csys =  4.78 CPU)
Result: FAIL
```

To reproduce in Travis, just make a dummy PR against the develop branch.

mattias-p:
The problem is not limited to those PRs. The problem is in the develop branch so it affects all PRs against the develop branch. It was introduced in #482.

matsduf:
Please create a better issue. I just wanted to create something so that we do not forget about the problem.

mattias-p:
I'll update the description of this one.

matsduf:
@mattias-p, please go ahead!

mattias-p:
[...]

PR #487, opened 2019-02-15
Revert "Memory leaks"

mattias-p:
Reverts zonemaster/zonemaster-backend#482

Trying to see if this solves #486.


mattias-p:
Reverting does not solve the problem. SQLite still fails in the same way.

Issue #481, opened 2019-01-30
CRITICAL - Memory leak of the test_agent module


mtoma:
The zonemaster test agent crashes on a regular basis with out of memory error:
2019-01-29T15:32:17Z [78459] CRITICAL - Cannot fork: Cannot allocate memory at /usr/local/share/perl5/Parallel/ForkManager.pm line 52.
2019-01-29T15:32:17Z [78459] NOTICE - Daemon terminating

This is very likely a memory leak either in the backend or in the engine.

mtoma:
The process is eating memory (should be heap) very quickly. This is the output of a pmap command on the process every minute:

grep 0000000004e23000 pmap_out.txt 
0000000004e23000  11496K rw---   [ anon ]
0000000004e23000  14556K rw---   [ anon ]
0000000004e23000  14680K rw---   [ anon ]
0000000004e23000  16152K rw---   [ anon ]
0000000004e23000  16516K rw---   [ anon ]
0000000004e23000  16760K rw---   [ anon ]

sandoche2k:
Is the problem relating to the engine?

mtoma:
Not sure. It might be. As of now I would say 70% chance backend code and 30% chance engine code.
But it is extremely difficult to debug.

matsduf:
PR #482 is assumed to fix this issue, but it has not been verified that it does solve the issue.

sandoche2k:
Machine 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"

mtoma:
We need an easy way to reproduce the issue.
@mtoma will work on a method to reproduce this issue

sandoche2k:
I have sent you information on connecting to the test machine

mtoma:
to reproduce the pmap output:
1. Find the pid of the zonemaster_test_agent process
2. As root run: pmap $PID
3. The output will start like this:
[...]

PR #489, opened 2019-02-22
Fixes issue #481 - Memory leak

mtoma:
Fixes the memory leak described in issue #481 
The line causing the issue is only commented out now as it was added there for a reason I can't remember.


matsduf:
The line of code, commented in this PR, was introduced by zonemaster-backend#461.

mtoma:
A follow up issue is #490 (to remove the dead code if no side effects are observed)

sandoche2k:
@mtoma is the travis failure related to the other two issues?

matsduf:
What fixed the memory leakage? The fix in this PR or the reset() (see #486).

PR #492, opened 2019-03-18
Fixes #486 - Updated test01.data with new cache format from the engine

### LINKED ISSUE #486: Sqlite fails

mtoma:
Fixes #486 - Updated test01.data with new cache format from the engine


mtoma:
This is still crashing in Travis...
I reproduced the problem in my local environment, recorded a new .dat file and all tests were successfully passing the SQLite tests.
In Travis it still makes external queries on the network.

Any ideas anyone?

mtoma:
Fixed

matsduf:
@mtoma, can you explain why loading the cache file twice should solve the issue?

mtoma:
Sure, the test01.t file creates 2 instances of Zonemaster::Engine.
One for each of the:

run_zonemaster_test_with_backend_API(1);
run_zonemaster_test_with_backend_API(2);

calls. We recently added Zonemaster::Engine->reset() call after each zonemaster engine runs (don't remember the exact reason, but it was advised by Vincent)

Seems the reset() call resets the Engine global to some defaults including the default "go online" option.

So the solution is to set the "don't go online" option before each test.

t/test01.t
```diff
@@ -59,6 +59,11 @@ sub run_zonemaster_test_with_backend_API {
 	use_ok( 'Zonemaster::Backend::Config' );
 	my $config = Zonemaster::Backend::Config->load_config();
 
+	if ( not $ENV{ZONEMASTER_RECORD} ) {
+		Zonemaster::Engine->preload_cache( $datafile );
```
mattias-p: Have you considered moving the call to reset() from run() to main() in zonemaster_backend_testagent instead of re-setting this stuff here?

matsduf:
@mtoma, do you know why the test of sqlite runs test01.t and test_validate_syntax.t whereas the test of postresql and mysql runs test_DB_backend.pl instead?

And why test01.t uses recorded data and why test_DB_backend.pl does not (use online data instead)?

t/test01.t
mtoma: Well, the issue was that something needed to be reset after each call to the engine (sorry can't remember what it was). So the point to put it in run was to call it after each run.

mtoma:
Yes Mats, the point of the SQLite DB backend is to be used as the default test testing as much as possible of the backend functionality including the TestAgent when the "make test" install step is run. This is not feasible using any other of the DB backends.
This is more or less also the case for the difference between the online/offline difference.
The idea is to have "make test+install" compatible with as many test environments as possible on CPAN. We can't assume that a working network + reasonable DNS environment will be available.
On the other hand if we want to install something more production oriented with a real DB backend we probably also want to test if everything works as it should including online tests.

mattias-p:
I tried this (https://github.com/zonemaster/zonemaster-backend/pull/492/commits/dede1595ef740b3951f4b4c1b8ce29a9e2e8a12b) on my development machine but it fails just like the develop branch. (Debian 8)

mtoma:
Is it surprising? You should also add the part that sets again the no_network parameter if you merge the Engine->reset() call.

t/test01.t
vlevigneron: Well the reset was added because we thought it could be a reason for memory leaks. If it is not the case and if you think it raises other issues, feel free to remove it.

mtoma: @mattias-p I removed the call as in the TestAgent main method there is no dependency on the Engine and it seems to me cleaner not to add it for a call we basically not seem to need.

mattias-p:
@mtoma wrote:
> @mattias-p I removed the call as in the TestAgent main method there is no dependency on the Engine and it seems to me cleaner not to add it for a call we basically not seem to need.

It seems to me that having the test agent clear the cache between tests is the correct thing to do. Otherwise, aren't we leaking cached responses between tests? Maybe that was the reason for the memory leak? Or do we have some other mechanism for cache eviction that I'm just not aware of?

If you don't want zonemaster_backend_testagent to import Engine directly, you could make TestAgent expose Engine->reset() through a wrapper method. I.e.:
```perl
# in zonemaster_backend_testagent:
TestAgent->new( ... )->run( ... );
TestAgent->reset();
```
and
```perl
# in TestAgent.pm
sub clear {
  Engine->clear();
}
```

sandoche2k:
@matsduf @mattias-p @vlevigneron pls. review

matsduf:
@mtoma, is there a PR to add the reset() in zonemaster_backend_testagent instead? Please put a reference to that from this PR.

PR #494, opened 2019-04-03
Update RPC interface

mattias-p:
This is the last stretch of #317. Introduces regex restrictions for previously free-form strings. Some of the restrictions are somewhat arbitrary.
Also includes a couple of collateral fixes.


mattias-p:
Test failures are caused by the usual suspects.

PR #495, opened 2019-04-08
Drop support for Ubuntu 14.04

mattias-p:
After dropping support for Ubuntu 14.04, none of the supported platforms use Perl 5.18.


mattias-p:
Test failure is due to the [usual suspect](https://github.com/zonemaster/zonemaster-backend/pull/486).

PR #502, opened 2019-05-07
Improvements to installation instruction

mattias-p:
I made some improvements to the installation instruction for Debian while testing.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -291,23 +291,9 @@ sudo apt-get install mysql-server libdbd-mysql-perl
 Initialize the database:
 
 ```sh
-mysql --user=root --password < ./initial-mysql.sql
+sudo mysql --password < ./initial-mysql.sql
 ```
 
-For latest versions of MySQL, wherein the root password is blank and you get an
-error "Access denied for the above command" on running the above command ,
-follow the below procedure: 
-
-```sh
-If you know the root password
-        mysql -u root 
-In MySQL prompt type the following 
-        SET PASSWORD FOR 'root'@'localhost' =PASSWORD("root-password");
-        flush privileges;
-        quit;
-Run again 
-	mysql --user=root --password < ./initial-mysql.sql
-```
 > **Note:** This creates a database called `zonemaster`, as well as a user
 > called "zonemaster" with the password "zonemaster" (as stated in the config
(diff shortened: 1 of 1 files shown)


sandoche2k:
@mattias-p we could approve this only if we test it on Debian. So the approval may take time.

mattias-p:
@sandoche2k I hope this doesn't invalidate too much of your testing.

docs/Installation.md
```diff
@@ -544,7 +544,7 @@ you should be able to use the
 API on localhost port 5000 as below. The command requires that `curl` is installed.
 
 ```sh
-curl -s -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
+wget -q -O - --header "Content-Type: application/json" --post-data '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
```
matsduf: Please give a strong argument to change from curl to wget at this late stage. I am against the change for v2019.1.

matsduf:
The commit comment is "Use wget instead of curl on Debian", but the change is on OSs.

sandoche2k:
WGET vs CURL
https://daniel.haxx.se/docs/curl-vs-wget.html. 
I do support @matsduf that we may discuss it for next version

mattias-p:
I don't think curl vs wget is a big deal in this case. Feature-wise both meet our needs. Wget comes pre-installed, at least on some platforms, which is convenient. Removed the commit to sidestep the argument.

matsduf:
@mattias-p, the change in installation of dependencies doe not look like an editorial change. Is it? Can you please describe?

mattias-p:
@matsduf Well, as I described in the commit message, I've added some forgotten dependencies,
shifted to install some dependencies from APT instead of CPAN, and
removed some transitive and/or unused dependencies.

matsduf:
Sorry, the Github hides long commit messages and I did not pay attention to the dots. I think it is more visible when repeated in the description of the PR.

mattias-p:
Thanks, @matsduf.

Waiting to merge until it's been tested on Ubuntu.

sandoche2k:
@mattias-p merge?

PR #514, opened 2019-05-22
Preparation for release 4.0.0

matsduf:
* Updated Changes
* Updated Backend.pm

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,26 @@
 Release history for Zonemaster component Zonemaster-Backend
 
+v4.0.0 2019-05-22 (public release version)
+
+ [Status]
+ - This a public release fully tested before release. This version
+   will be available on CPAN.
+
+ [Fixes]
+ - Updated installation instructions (#502, #510, #511)
+ - Updated initial sql script for postgres (#504)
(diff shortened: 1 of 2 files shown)


matsduf:
Since there is a change in the IPA. The IPA has become stricter, see #494 and #317. This means that previously accepted calls may now be refused.  I followed [Versions and Releases].

[Versions and Releases]: https://github.com/zonemaster/zonemaster/blob/master/docs/design/Versions%20and%20Releases.md

pamasse:
@matsduf, yes, it's true, the documentation says: "MAJOR version when you make incompatible API changes". But, these changes have been made between  2.X and 3.X versions.
This version only removes backward compatibility solutions. 
In my point of view, it's more the final step of the new API migration (version 3.X) than a new new API  (version 4.X).

matsduf:
As far as I can see, the changes you can see in #494 will mean that fewer values will pass, i.e. we have a change of API. And we have the changes in both the documented API and in the implemented API. I think that from a strict interpretation it should be v4.0.0, but if you have a problem with that I could change it to v3.0.1, but we would break our specifications.

PR #526, opened 2019-06-13
Partial fix for translation on FreeBSD 11

### LINKED ISSUE #353: Translation on FreeBSD 11
Selecting language does not work in FreeBSD 11. There is currently no work-around. See also issues dotse/zonemaster-cli#46 and dotse/zonemaster-cli#64.

On FreeBSD the `language` parameter of the `get_test_results` method is ignored. Instead the translation language is determined from the `LC_ALL`, `LC_MESSAGES` and `LANG` environment variables when the service is started.

**Edit:** Updated links to the correct issues.
**Edit:** Clarified version of FreeBSD.

### LINKED ISSUE #512: Translation does not work under FreeBSD 11
This issue has been found on FreeBSD version 11.2-RELEASE-p10 (but it does not exist on tested version 12.0-RELEASE-p3) when running on develop branch of Zonemaster-Engine ([3d0c148]) and Zonemaster-Backend (a9ede9c). They have both the same versions of p5-Locale-Msgfmt (0.15_2), p5-Locale-libintl (1.31) and perl5 (5.28.1_1).

This issue is possibly related to issue zonemaster/zonemaster-engine#546.

When the test result is returned from the Backend RPC, then strings are not translated. In the extract below, the strings should have been translated to English (i.e., msgstr should be shown, not msgid), but are not. Compare with zonemaster/zonemaster-engine#546.

```
# curl -s -H "Content-Type: application/json" -d '{"jsonrpc": "2.0","id": 6,"method": "get_test_results","params": {"id": "23dc863d3490ee17","language": "en"} }' http://localhost:5000/ | jq .
{
  "jsonrpc": "2.0",
  "result": {
    "hash_id": "23dc863d3490ee17",
    "params": {
      "queue": 0,
      "domain": "zonemaster.net",
      "priority": 10,
      "ipv6": false
    },
    "id": 3,
    "creation_time": "2019-05-19 10:25:49",
    "results": [
(...)
      {
        "level": "INFO",
        "module": "ADDRESS",
        "message": "Reverse DNS entry exist for all Nameserver IP addresses.\n"
      },
      {
        "module": "ADDRESS",
        "level": "INFO",
        "message": "All reverse DNS entry matches name server name.\n"
      },
(...)
```

[3d0c148]: https://github.com/zonemaster/zonemaster-engine/commit/3d0c148d65bbab5c2a0fdca8bed76c04e3e25bfc

mattias-p:
This change places the responsibility for how to switch languages on Engine while keeping the responsibility of when that should be done in the Backend.

In combination with zonemaster/zonemaster-engine#562 it enables choosing translation language in requests to the RPCAPI on FreeBSD 11. However once you've chosen a language it gets stuck on that language no matter what language you ask for in subsequent requests. Compare that with the current behavior where the language is simply stuck on English. In my view this is an improvement.

Curiously, when I tested this on FreeBSD 11 I was consistently able to choose different languages twice, but then it got stuck on the second language. I haven't looked into why this happens.

Other OSes are unaffected - I've only verified this on FreeBSD 12 and Debian 8 and 9.

Partially fixes #353 and #512.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -17,7 +17,6 @@ requires
   'JSON::PP'                    => 0,
   'JSON::RPC'                   => 1.01,
   'JSON::Validator'             => 2.07,
-  'Locale::TextDomain'          => 0,
   'Log::Any'                    => 0,
   'Log::Any::Adapter::Dispatch' => 0,
   'Log::Dispatch'               => 0,
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -237,7 +237,7 @@ See the [post-installation] section for post-installation matters.
 Install dependencies available from binary packages:
 
 ```sh
-sudo apt-get install libclass-method-modifiers-perl libconfig-inifiles-perl libdbd-sqlite3-perl libdbi-perl libfile-sharedir-perl libfile-slurp-perl libhtml-parser-perl libintl-perl libio-captureoutput-perl libjson-pp-perl libjson-rpc-perl liblog-any-adapter-dispatch-perl liblog-any-perl liblog-dispatch-perl libmoose-perl libplack-perl libplack-middleware-debug-perl librole-tiny-perl librouter-simple-perl libstring-shellquote-perl starman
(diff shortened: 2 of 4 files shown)


mattias-p:
I've updated the PR description to clarify regarding other affect on OSes.

mattias-p:
@matsduf What changes are you requesting? Do you think that no fix is better than this partial fix as currently implemented? I've spent a fair amount of time on this already and intend to shift my focus to other things.

mattias-p:
@matsduf Good catch! Added a commit to clean that up.

mattias-p:
I just found Locale::TextDomain in the installation instruction as well. I didn't expect it there since it's already included in the Engine installation instruction.

PR #527, opened 2019-06-14
Error handling and logging

mattias-p:
This is a collection of fixes for problems indirectly related to having incorrect database configuration in backend_config.ini.


mattias-p:
abf018b brought unintended consequences to the Test Agent. Because of Test Agent's frivolous reloading of the config file (see #528) these events would fill up the disk if they were logged. To avoid that I'm removing the logging commit from this PR.

Issue #214, opened 2016-11-11
More and better logging from the backend


matsduf:
Issue #213 give an example of error messages are leaking out through the API. Instead those error messages should be logged with a time stamp for every log message. It is better to have extensive logging than no logging.

For trouble shooting it is good to be able to increase the amount of logging.

mattias-p:
I've created issue #340 for logging in the Test Agent specifically. I have a proof of concept for that.

This issue also implies that RPCAPI needs better logging but that's out of scope for #340.

mattias-p:
The requirements for closing this issue should be clarified. The issue would be easier to understand if the requirements were listed separate from the background reasoning. In particular "more" and "better" needs to be quantified somehow. Is adding timestamps enough for "better"? If "more" isn't specified this becomes more of a political statement than an action point.

PR #529, opened 2019-06-19
Avoid reloading configuration in Test Agent

### LINKED ISSUE #528: Avoid reloading configuration in Test Agent
Today the Test Agents reloads the configuration at every change it gets. This is problematic because a single request could be processed using values taken from different versions of the configuration file.

A better design would be to load the configuration file only once during startup. If the configuration needs to be updated, the Test Agent could be restarted. The loading of the config file (including its path and the results of reading, parsing and pre-flight check) should be logged.

A further improvement upon that would be to have the Test Agent master process reload the configuration file on SIGHUP and only accept the new configuration if it passes the preflight check. The reloading should be logged similarly to the initial loading.

mattias-p:
Fixes #528. The further improvement with reloading on SIGHUP is not included here. I'll create a separate issue for that.

This PR also takes a step towards #214 by introducing a log entry for loading of the config file (which wasn't feasible before with the relentless config reloading), and also beginning and end entries for the "pre-flight check".

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -28,6 +28,8 @@ else {
 sub load_config {
     my ( $class, $params ) = @_;
     my $self = {};
+
+    $log->notice( "Loading config: $path" );
     
     $self->{cfg} = Config::IniFiles->new( -file => $path );
     die "UNABLE TO LOAD $path ERRORS:[".join('; ', @Config::IniFiles::errors)."] \n" unless ( $self->{cfg} );
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -239,7 +239,7 @@ sub new_DB {
     $dbclass->import();
     $log->notice("Constructing database adapter: $dbclass");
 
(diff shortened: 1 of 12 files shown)


lib/Zonemaster/Backend/Config.pm
```diff
@@ -235,12 +239,56 @@ sub new_DB {
...
+    my $db = $dbclass->new({ config => $self });
 
     # Connect or die
     $db->dbh;
 
     return $db;
 }
 
+sub new_PM {
+    my $self = shift;
```
matsduf: This method is not documented. What is "PM"?

mattias-p: PM is short for processing manager. I added documentation.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -235,12 +239,80 @@ sub new_DB {
...
     return $db;
 }
 
+=head2 new_PM
+
+Create a new processing manager object according to configuration.
+
+=head3 INPUT
+
+The values of the following attributes affect the consturction of the returned object:
```
matsduf: "consturction"--> "construction"

PR #531, opened 2019-06-26
Log details of terminated tests

### LINKED ISSUE #341: Log details of terminated tests
Test Agents allocate new test requests, perform them and register their results.

Today when a test crashes, all context is discarded (except for remains in the database). This makes it difficult to make good bug reports and to reproduce and fix bugs.

Today when a test crashes, its registration stays allocated and the test is never resumed. It's difficult to discern in the database which test registrations have running test agents and which ones are remains of crashed tests. This is especially problematic for test batches. If a single test in a test batch crashes the test batch will never complete. This renders the test batch's test user defunct as it can't start another batch until the previous one has completed.

The Test Agent must log details about the termination of all tests so that administrators and developers can analyze the logs to notice problems and take appropriate action (e.g. cleaning up erroneous remains in the database, starting a new test or reporting/reproducing/fixing a bug).

The following details must be logged:
1. When a test terminates normally, the **test id** must be logged.
2. When a test dies, the test id and **$EVAL_ERROR** must be logged.
3. When a test terminates from a signal, the **test id**, the **signal name** and the **stack trace** must be logged.
4. When the Test Agent master process sends a signal to a worker, the **worker pid** and the **signal name** must be logged.

The log facility itself is described in #340.

mattias-p:
Fixes #341.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -286,9 +286,11 @@ sub new_PM {
     $pm->run_on_wait(
         sub {
             foreach my $pid ( $pm->running_procs ) {
-                my $diff = time() - $times{$pid};
+                my $diff = time() - $times{$pid}[0];
+                my $id   = $times{$pid}[1];
 
                 if ( $diff > $timeout ) {
+                    $log->warning( "Worker process (pid $pid, testid $id): Timeout, sending SIGKILL" );
                     kill 9, $pid;
                 }
             }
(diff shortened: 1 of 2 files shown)


mattias-p:
This needs some more work. Not sure if these changes introduce a bug or expose an old one. In any case the master process fails to `wait` for the worker processes, resulting in a new attempt to kill it every second, and equally many log lines about doing so.

mattias-p:
I rebased the PR to make it neat in case we need to come back to it at a later date. Sorry for the inconvenience. `Log worker process exit status` is unchanged in contents.

matsduf:
@mtoma, could you review too?

matsduf:
Fixes #341.

PR #534, opened 2019-06-28
Add experimental CLI tools for backend

mattias-p:
Over a couple of years I've been cultivating a two helper programs that I've been using during development of the Backend. They've been a huge help in my work. Over time they've become easier to use and now I believe they've reached a point where other people should be able to use them without much trouble, even though the documentation is scant.

`zmb` is a tool for sending single requests to the RPCAPI. I deliberately made it possible to send invalid requests in order to try out the behavior of the RPCAPI in the presence of such. Also it only covers the most common commands and the most common parameters. I simply added things as I needed them.

`zmtest` is a synchronous client for running a test though the asynchronous Backend. I just starts a tests, polls it until its at progress 100 and then prints its results.

I recommend piping the output from both of these to the `jq` tool. I should also mention the `jq --sort-keys` option in case you haven't noticed it.

### DIFF
--- a/MANIFEST.SKIP
+++ b/MANIFEST.SKIP
@@ -6,6 +6,10 @@ maint
 ^.*\.swp$
 ^MANIFEST\.SKIP$
 
+# Avoid experimental tools
+script/zmb
+script/zmtest
+
 #!start included /usr/share/perl/5.20/ExtUtils/MANIFEST.SKIP
 # Avoid version control files.
 \bRCS\b
--- a/script/zmb
+++ b/script/zmb
@@ -1,0 +1,424 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+use feature 'say';
+
+use Getopt::Long qw( GetOptionsFromArray :config require_order );
+
+use Pod::Usage;
+
+use JSON::PP;
+use LWP::UserAgent;
+
+=head1 NAME
+
+B<zmb> - Shell bindings for the Zonemaster::Backend RPC API
+
+Zmb is meant to be pronounced I<Zimba>.
+
+=head1 SYNOPSIS
+
+zmb COMMAND [OPTIONS]
+
+=head1 GLOBAL OPTIONS
+
+ --help        Show usage
+ --verbose     Show RPC query
+
+=cut
+
+sub main {
+    my @argv = @_;
+
+    my $opt_help;
+    my $opt_verbose;
+    GetOptionsFromArray(
+        \@argv,
+        'help'    => \$opt_help,
+        'verbose' => \$opt_verbose,
+    ) or pod2usage( 2 );
+    if ( !@argv ) {
+        pod2usage( -verbose => 99, -sections => ['SYNOPSIS', 'GLOBAL OPTIONS'], -exitval => 'NOEXIT' );
+        show_commands();
+        exit 1;
+    }
+    my $cmd = shift @argv;
+    pod2usage( 1 ) if !defined $cmd;
+    my $cmd_sub = \&{ "cmd_" . $cmd };
+    pod2usage( "'$cmd' is not a command" ) if !defined &$cmd_sub;
+    pod2usage( -verbose => 99, -sections => ["COMMANDS/$cmd"] ) if $opt_help;
+
+    my $json = &$cmd_sub( @argv );
+
+    if ( $json ) {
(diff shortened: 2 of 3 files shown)


matsduf:
The jq `--sort-key` is only available for newest versions of jq.

script/zmb
```diff
@@ -0,0 +1,424 @@
...
+
+    zmb start_domain_test [OPTIONS]
+
+Options:
+
+    --domain DOMAIN_NAME
+    --nameserver DOMAIN_NAME:IP_ADDRESS
+    --client-id CLIENT_ID
+    --client-version CLIENT_VERSION
+    --ds-info DS_INFO
```
matsduf: State what options that are mandatory.

script/zmb
```diff
@@ -0,0 +1,424 @@
...
+
+=head2 get_test_history
+
+ zmb get_test_history [OPTIONS]
+
+ Options:
+   --domain DOMAIN_NAME
+   --nameserver true|false|null
+   --offset COUNT
+   --limit COUNT
```
matsduf: State what options that are mandatory.

script/zmb
```diff
@@ -0,0 +1,424 @@
...
+}
+
+
+=head2 add_api_user
+
+ zmb add_api_user [OPTIONS]
+
+ Options:
+   --username USERNAME
+   --api-key API_KEY
```
matsduf: State that both options are mandatory.

script/zmtest
```diff
@@ -0,0 +1,46 @@
...
+zmb () {
+    json="`"${ZMB}" "$@"`"
+    if echo "${json}" | "${JQ}" -e .error > /dev/null ; then
+        error 1 "method $1: ${json}"
+    else
+        echo "${json}"
+    fi
+}
+
+[ -n "${JQ}" ] || error 2 "Dependency not found: jq"
```
matsduf: Change error message to "You have to install jq to run this script" or similar.

script/zmtest
```diff
@@ -0,0 +1,46 @@
...
+    printf "\r${progress}%% done" >&2
+    if [ "${progress}" -eq 100 ] ; then
+        echo >&2
+        break
+    fi
+    sleep 1
+done
+
+# Get test results
+zmb get_test_results --testid "${testid}" --lang en
```
matsduf: Add 
```
| "${JQ}" --sort-keys .
```
at the end to make it nicer.

script/zmtest
```diff
@@ -0,0 +1,46 @@
...
+echo "testid:" ${testid} >&2
```
matsduf: Add an extra "echo" before to separate it from the JSON blob.

MANIFEST.SKIP
```diff
@@ -6,6 +6,10 @@ maint
 ^.*\.swp$
 ^MANIFEST\.SKIP$
 
+# Avoid experimental tools
```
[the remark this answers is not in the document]
mattias-p: I take your point about more convenient testing. I created #536 for this purpose.

script/zmb
mattias-p: All of them are optional by design. If zmb crashes because you omitted some of them, that's a bug in zmb.

script/zmtest
mattias-p: I'd rather not. You can always run `zmtest zonemaster.net | jq -S` if you need sorting.

script/zmtest
mattias-p: In my view it's better the way it is. It is clearly visible to the human consumer that the testid line is not part of the JSON blob, and it's also clearly visible to the machine consumer that the two parts are separated into different streams.

mattias-p:
@mtoma Sounds like a reasonable feature. Thanks for reviewing.

script/zmb
```diff
@@ -0,0 +1,424 @@
...
+    if ( exists $args{params} ) {
+        $request->{params} = $args{params};
+    }
+    return encode_json( $request );
+}
+
+sub to_request {
+    my $json = shift;
+
+    my $req = HTTP::Request->new( POST => 'http://localhost:5000/' );
```
matsduf, line 403: It should be possible to select another host and another port.

script/zmb
matsduf: You cannot run a test without domain name.

IPv4/6 on/off is missing.

Issue #507, opened 2019-05-16
Warning message from RPCAPI


matsduf:
Read in the zm_rpcapi.log file:

```
2019/05/16-21:38:09 Starman::Server (type Net::Server::PreFork) starting! pid(26464)
Binding to TCP port 5000 on host 127.0.0.1 with IPv4
coerce(1) will be deprecated. at /usr/local/lib/perl5/site_perl/JSON/Validator/Joi.pm line 84.
```

coerce(1) should be reviewed.

matsduf:
Find the recommended replacement solution.

mattias-p:
@mtoma Sorry I nicked this issue from you. I didn't realize you were the assignee until now. I guess I should have paid more attention and I hope you didn't start working on it. Assigning myself since I already got the fix accepted upstream and I already made the PR for the workaround.

mattias-p:
@matsduf Shouldn't this be closed as fixed?

matsduf:
Since this is just a work-around so we agreed to keep this issue, but for me it is OK to close it. Should an issue to remove the work-around, when possible, be created?

matsduf:
See #562.

PR #537, opened 2019-08-06
Work around deprecation warning

mattias-p:
This works around the deprecation warning reported in #507. The offending code is in JSON::Validator::Joi::validate. I created mojolicious/json-validator#165 to fix it. This PR inlines the fixed code as a stop-gap until a fixed version of JSON-Validator is available on all supported platforms. I've added a comment about reinstating the call to JSON::Validator::Joi::validate and when it is appropriate to do so.


mattias-p:
Rebased onto current develop.

PR #538, opened 2019-08-06
Try to debug Travis problem

mattias-p:


mattias-p:
The user hosts update got us further (08f075f), but the default value change did nothing (01c9f45).

matsduf:
Is the change compatible with older versions of Mysql? And with MariaDB?

Issue #540, opened 2019-09-24
Travis fails in all kinds of ways


mattias-p:
All of a sudden Travis started failing in interesting ways, as can be seen in the documentation-only PR #535. Travis fails for all database configuration, but what's interesting is that they fail in different ways.

Here are Travis logs from #535:
[postgresql.txt](https://github.com/zonemaster/zonemaster-backend/files/3646301/postgresql.txt)
[sqlite.txt](https://github.com/zonemaster/zonemaster-backend/files/3646302/sqlite.txt)
[mysql.txt](https://github.com/zonemaster/zonemaster-backend/files/3646300/mysql.txt)

mattias-p:
Fixed by #542.

PR #541, opened 2019-09-24
Updated Travis config after zonemaster-ldns#51

mattias-p:
zonemaster/zonemaster-ldns#51 adds feature checks for Ed25519, but Travis provides an older version of OpenSSL.

This is blocked by #540.

I'm disabling Perl 5.16 in Travis because the tests fail for this version. I've created #551 to re-enable it. @matsduf and I discussed this and it is our view that for this release it's better to disable these tests and have zonemaster/zonemaster-ldns#51 implemented, than not having that feature implemented but having enabled tests for Perl 5.16 in Zonemaster Backend.


.travis.yml
```diff
@@ -20,14 +20,17 @@ perl:
 before_install:
     - sudo apt-get install -y libidn11-dev
     - eval $(curl https://travis-perl.github.io/init) --auto
-    - local-lib
+
+    - sudo apt-get install -y libidn11-dev
```
matsduf: Should that really be installed a second time?

mattias-p: They should not. I'll remove one of the commands.

.travis.yml
```diff
@@ -15,19 +15,85 @@ perl:
     - "5.24"
     - "5.22"
     - "5.20"
-    - "5.16"
```
matsduf: I suggest that 5.16 is kept, but commented with a short notice of why.

mattias-p:
@matsduf I created #551 to re-enable Perl 5.16, added an explanation regarding this to the PR description above, and a comment to the configuration about 5.16 disabled.

mattias-p:
@sandoche2k, @mtoma Do you agree with disabling Perl 5.16 in Travis for the benefit of merging zonemaster/zonemaster-ldns#85?

Issue #525, opened 2019-06-04
Garbage collector for unfinished tests


matsduf:
A test is created in the database by RPCAPI and then processed (executed) by TestAgent. Until a TestAgent has set "progress" to 100 (%) RPCAPI will not present the result of that test.

If the TestAgent does a timeout or crashes on a test, the "progress" will never go to 100, which means that the test will never be visible from the RPCAPI. Especially for batches this is a problem, but also for a single test this is a problem because it is never completed.

* Some kind of garbage collector is needed that can make a test, that has not completed within some set time, finished (progress = 100). Or it could be a new status that is set to "done".
* The timeout time that the garbage collector should use must be configurable in the backend.ini.
* When the garbage collector makes a test finished it should also add a system message that says that the test was never finalized. It could be message in SYSTEM with a default level of CRITICAL.
* The RPCAPI must present tests that have been finished in this way. 

This issue is related to #175 and #216.

sandoche2k:
@mtoma to have a look

matsduf:
Resolved by #544.

mtoma:
What is blocking this issue from being merged again?

PR #544, opened 2019-10-08
Fix issue #525 - Garbage collector for unfinished tests

mtoma:
Fix issue #525 - Garbage collector for unfinished tests


lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -295,6 +295,37 @@ sub add_batch_job {
     return $batch_id;
 }
 
+sub process_unifinished_tests {
```
mattias-p: There's a lot of duplication between the DB adapters here. Do you think it would be possible to refactor this to put the main logic in DB.pm and only have the DB specific details in the adapters?

share/initial-mysql.sql
```diff
@@ -18,7 +18,8 @@ CREATE TABLE test_results (
 	params_deterministic_hash character varying(32),
 	params blob NOT NULL,
 	results blob DEFAULT NULL,
-    undelegated boolean NOT NULL DEFAULT false
+    undelegated boolean NOT NULL DEFAULT false,
```
mattias-p: The indentation is inconsistent here. Have you considered replacing tab characters with spaces? Personally I like to use perltidy for this purpose.

mattias-p:
What's the idea behind retrying? Do you have an example of when this is useful?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -340,6 +340,39 @@ sub add_batch_job {
...
 
+sub process_unifinished_tests {
+	my ( $self ) = @_;
+	
+	my $dbh = $self->dbh;
+	
+	my $query = "
+		SELECT hash_id, results, nb_retries
+		FROM test_results 
+		WHERE test_start_time < DATE_ADD(NOW(), INTERVAL -".$self->config->timeout_before_retrying_test_in_minutes()." MINUTE)
```
mattias-p: You probably need to filter for queue as well, since there's no locking.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -9,7 +9,7 @@ use 5.14.2;
 use Data::Dumper;
 
 requires 'add_api_user_to_db', 'user_exists_in_db', 'user_authorized', 'test_progress', 'test_results',
-  'create_new_batch_job', 'create_new_test', 'get_test_params', 'get_test_history', 'add_batch_job';
+  'create_new_batch_job', 'create_new_test', 'get_test_params', 'get_test_history', 'add_batch_job', 'process_unifinished_tests';
```
matsduf: 'process_unifinished_tests' => 'process_unfinished_tests'?

lib/Zonemaster/Backend/DB/PostgreSQL.pm
matsduf: 'sub process_unifinished_tests' => 'sub process_unfinished_tests'?

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -294,6 +294,40 @@ sub add_batch_job {
     my ( $self, $params ) = @_;
 }
 
+sub process_unifinished_tests {
```
matsduf: 'sub process_unifinished_tests' => 'sub process_unfinished_tests'?

lib/Zonemaster/Backend/DB/MySQL.pm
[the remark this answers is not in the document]
mtoma: DATE_ADD works with negative intervals. But ok, changing it to DATE_SUB

lib/Zonemaster/Backend/DB/PostgreSQL.pm
mtoma: The only way is to add two "query creation" methods, one for the select and one for the update, to each backend specific module and use them in DP.pm.
Do you prefer to have it done like this?

mtoma:
> What's the idea behind retrying? Do you have an example of when this is useful?

When you "hot restart" a test agent on a live system with running tests?
Under heavy load when the default timeout is too short for a few tests?

But mostly because I prefer to have this small aditionnal flexibility for large batch tests.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
mattias-p: Yeah, I think so. In principle it's a better design make the database adapters have well-defined APIs of small methods in order to keep them as free from business logic as possible. That way we can test the business logic with mocked out database adapters.

lib/Zonemaster/Backend/DB/MySQL.pm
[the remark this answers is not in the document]
mattias-p: Oh, I didn't see the negation sign :)

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -340,6 +340,35 @@ sub add_batch_job {
...
 }
 
+sub build_process_unfinished_tests_select_query {
+    my ( $self ) = @_;
+    
+    if ($self->config->lock_on_queue()) {
+        return "
+            SELECT hash_id, results, nb_retries
+            FROM test_results 
+            WHERE test_start_time < DATE_SUB(NOW(), INTERVAL ".$self->config->timeout_before_retrying_test_in_minutes()." MINUTE)
```
mattias-p: Would it be possible to use `?` instead of string concatenation in cases like this? That way we'd be more resilient against SQL-injection attacks.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -112,6 +112,34 @@ sub get_batch_job_result {
 	return \%result;
 }
 
+sub process_unfinished_tests {
+    my ( $self ) = @_;
+    
+    my $dbh = $self->dbh;
```
mattias-p: Would it be possible to move the `dbh`, `prepare`, `execute` and `do` calls to the database adapters? That way we could write tests for the business logic.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -186,6 +186,22 @@ sub lock_on_queue {
     return $val;
 }
 
+sub timeout_before_retrying_test_in_minutes {
+    my ($self) = @_;
+
+    my $val = $self->{cfg}->val( 'ZONEMASTER', 'timeout_before_retrying_test_in_minutes' );
```
matsduf: That is a very long variable name. I suggest a shorter variable for the INI file, e.g. "timeout_before_retry", and I suggest in seconds instead of minutes.

This method needs documentation in POD below.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -186,6 +186,22 @@ sub lock_on_queue {
...
+
+    return ($val)?($val):(10);
+}
+
+sub maximal_number_of_retries {
+    my ($self) = @_;
+
+    my $val = $self->{cfg}->val( 'ZONEMASTER', 'maximal_number_of_retries' );
```
matsduf: I think this could be shorter too, e.g. "max_retries".

This method needs documentation in POD below.

script/zonemaster_backend_testagent
```diff
@@ -91,6 +91,7 @@ sub main {
         $self->pm->on_wait();                   # Sends SIGKILL to overdue child processes
 
         my $id = $self->db->get_test_request();
+        $self->db->process_unifinished_tests();
```
matsduf: "process_unifinished_tests" => "process_unfinished_tests"?

lib/Zonemaster/Backend/DB/MySQL.pm
mtoma: I don't think it is possible in this specific case. But not 100% sure.

lib/Zonemaster/Backend/DB.pm
mtoma: I'm not sure to understand what you mean here.

lib/Zonemaster/Backend/Config.pm
mtoma: I really prefer very long self explanatory variables that do not require to read documentation do be understood.

Adding POD for the method.

lib/Zonemaster/Backend/Config.pm
mtoma: I really prefer very long self explanatory variables that do not require to read documentation do be understood.

Adding POD for the method.

script/zonemaster_backend_testagent
mtoma: Fixed

lib/Zonemaster/Backend/Config.pm
mtoma: I don't think there is any possible use case for the timeout to require to be in seconds. This timeout should definitely be magnitudes higher than the maximal zonemaster test duration.
I don't see when seconds could be necessary here.

lib/Zonemaster/Backend/Config.pm
matsduf: Yeah, but the variable name does not need to be longer needed. That increases the risk of spelling errors.

lib/Zonemaster/Backend/DB.pm
mattias-p: What I'm trying to get at is that I'd like to have a unit test that demonstrates that the garbage collection works. But I realize this requires a larger refactoring. I'll create a separate issue for that purpose.

lib/Zonemaster/Backend/DB/MySQL.pm
mattias-p: Fair enough.

lib/Zonemaster/Backend/Config.pm
mtoma: Well, testing is not very easy here.
There are two ways. One is to restart the backend test agent at the right time where the minimal 1 minute timeout seems about right (not the recommended way as it is rather random).

Another one which is probably much more preferable is this one:
```
update test_results set test_start_time = '2019-01-01', progress = 50, results = NULL WHERE hash_id = '75adf9b0cc27042e';
```
And the retry starts instantly.

lib/Zonemaster/Backend/Config.pm
mtoma: I agree, but we live in a world of cut and paste, don't we?
There are ready made place holders in the backend_config.ini file now.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -186,6 +186,50 @@ sub lock_on_queue {
     return $val;
 }
 
+=head2 timeout_before_retrying_test_in_minutes
+
+An optionnal parameter that allows to customize the period of time in minutes after which an unfinished test will be automatically terminated by the backend with a critical error.
```
matsduf: "optionnal" => "optional"

The name of the method and the description have different focuses. The PR was to resolve the issue when tests die before they are completed and hang uncompleted. That should also be reflected by the method name and the setting in the INI file.

matsduf:
There are two timeouts now, `max_zonemaster_execution_time` and `timeout_before_retrying_test_in_minutes`, one is in seconds and the other is in minutes. I really think that both should be in seconds.

I think it is unclear how the two interact. Your really want to set one value. When you have reached that value and the test is unfinished, then you either want to restart the test or setting test to be completed but with an error.

mattias-p:
Here's an example of a sequence of events that we want to avoid. I believe it is possible with the current design.
1. a worker starts a test
2. the worker takes a long time
3. the reaper is slow to kill the worker (because of scheduling issues or a longer timeout configuration)
4. the garbage collector recycles the test
5. a second test agent starts the test
6. the test agents fight to update the progress value
7. the first test agent completes and finishes the test
8. the second test agent completes and finishes the test with a different result

IMHO the right way to solve this is design a robust DB API that makes it possible for the various actors to cooperate in a meaningful way. When I say actors I mean rpcapi, garbage collector, test agent master, a horde of test agent workers, etc.

mtoma:
> There are two timeouts now, `max_zonemaster_execution_time` and `timeout_before_retrying_test_in_minutes`, one is in seconds and the other is in minutes. I really think that both should be in seconds.
> 
> I think it is unclear how the two interact. Your really want to set one value. When you have reached that value and the test is unfinished, then you either want to restart the test or setting test to be completed but with an error.

I missed that other parameter. Shouldn't that be the same thing? I can remove `timeout_before_retrying_test_in_minutes` and only use `max_zonemaster_execution_time` instead.
Seems to me both are serving the same puropse.

matsduf:
> I missed that other parameter. Shouldn't that be the same thing? I can remove `timeout_before_retrying_test_in_minutes` and only use `max_zonemaster_execution_time` instead.

@mtoma, if you can create an implementation with `max_zonemaster_execution_time` I think that would be a better solution. It would be clearer and more robust.

In https://github.com/zonemaster/zonemaster-backend/pull/544#issuecomment-540535439 concerns are raised about the restart function. I suggest that we focus on the main issue, that we do not have hanging tests that are never completed. We make sure that works correct in the case that retry=0.

The retry function, which could be useful for batches, could also be implemented if possible in a reasonable way, but we put less effort on testing and we mark and document the function as **experimental** with the default value of zero (no retry). For next release we could possibly have it as a supported function.

mtoma:
> Here's an example of a sequence of events that we want to avoid. I believe it is possible with the current design.
> 
> 1. a worker starts a test
> 2. the worker takes a long time
> 3. the reaper is slow to kill the worker (because of scheduling issues or a longer timeout configuration)
> 4. the garbage collector recycles the test
> 5. a second test agent starts the test
> 6. the first and second test agent fight to update the progress value
> 7. the first test agent completes and finishes the test with a result
> 8. the second test agent completes and finishes the test with a different result
> 
> IMHO the right way to solve this is to design a robust DB API that makes it possible for the various actors to cooperate in a meaningful way. When I say actors I mean rpcapi, garbage collector, test agent master, a horde of test agent workers, etc.

That seems to me rather easy to fix. Lets just add a test on the update of progress to have the previous progress value <= to the want we want to write in the database. In this case the more advanced (in your example the first worker) should be the one that completes the test and never gets overriden by the second that started after a retry.

share/initial-postgres.sql
```diff
@@ -16,7 +16,8 @@ CREATE TABLE test_results (
 	progress integer DEFAULT 0,
 	params_deterministic_hash varchar(32),
 	params json NOT NULL,
-	results json
+	results json,
+    nb_retries integer NOT NULL DEFAULT 0
```
matsduf: Inconsistent indentation. Use spaces instead of tabs.

mtoma:
> > Lets just add a test on the update of progress to have the previous progress value <= to the want we want to write in the database. In this case the more advanced (in your example the first worker) should be the one that completes the test and never gets overriden by the second that started after a retry.
> 
> I had no trouble at all coming up with a scenario where that breaks down as well. When you get into parallelism and concurrency you run into lots of tricky edge cases and you really need to think things through.
> 
> Maybe we should take @matsduf's suggestion and go with the retry-less variant for now. I believe never overriding a finished result should be enough of a change to get that working properly.

Well, the retry is set to 0 already by default. But I don't agree we have a lot of tricky parallelism edge cases to handle. All what we have is a "basic prallelism" where we allow several workers override the values in the database. We only need to make sure there is not a possibility of infinite loop. And preferably let the worker still working on the test and that is the most advanced in terms of progress write it's results. For a worker that once went over the max_execution_time I don't even care what is written in the DB. Both some kind of result and a critical error are fine for me.

mattias-p:
> For a worker that once went over the max_execution_time I don't even care what is written in the DB.

You should care. The system needs take proper care to ignore those writes because otherwise they're messing with the state of the system, and we don't want that.

mtoma:
> > For a worker that once went over the max_execution_time I don't even care what is written in the DB.
> 
> You should care. The system needs take proper care to ignore those writes because otherwise they're messing with the state of the system, and we don't want that.

Let's not forget about the initial reason of this issue: unfinished tests that stay in the DB. For this having edge cases where the test execution time is nearly identical to the timeout are not really the main focus. If that means the timeout should be set to 1 hour I'm ok with this.
I want to have a retry because it is needed if I giant zonemaster batch starts and needs to be interrupted for whatever reason and keeps unfinished tests in the DB. In this case I want to be able to just restart the test agent and not need to go messing in the DB with SQL to finish the interrupted tests.

I think if we add the test on the progress when updating this objective as well as the initial one to not have unfinished tests forever in the DB if for some reason the engine is not able to finish the tests is acheved.

mattias-p:
I'll make a proposal for a robust DB API that doesn't come with headache inducing weird behaviors.

matsduf:
@mtoma, let us focus on the main issue #525 (that timed out, unfinished tests stay in the database) and not on the new feature (restarting timed out tests). For this release, let us make sure that garbage collection works correctly when retry=0. Keep retry=0 as the default value. Include the retry function, but make it experimental for this release. That it is experimental should be stated in the INI file and in the documentation.

Before next release we have time to test the retry function. We can ask @aabdnn if he sees a use case for it for RIPE.

@mtoma, please comment my proposal.

mtoma:
> @mtoma, let us focus on the main issue #525 (that timed out, unfinished tests stay in the database) and not on the new feature (restarting timed out tests). For this release, let us make sure that garbage collection works correctly when retry=0. Keep retry=0 as the default value. Include the retry function, but make it experimental for this release. That it is experimental should be stated in the INI file and in the documentation.
> 
> Before next release we have time to test the retry function. We can ask @aabdnn if he sees a use case for it for RIPE.
> 
> @mtoma, please comment my proposal.

Do I need to do enything else than updating in the .ini file comment to state that the feature is experimental?

matsduf:
> Do I need to do enything else than updating in the .ini file comment to state that the feature is experimental?

In the INI file and in the POD documentation of the retry setting in Config.pm.

mattias-p:
There are a bunch of other methods in the DB modules that don't respect the new requirement to never update tests with progress=100. Do you want to make those APIs safe as part of this PR? Otherwise I can create an issue for that so we don't forget it.

share/backend_config.ini
```diff
@@ -17,6 +17,12 @@ max_zonemaster_execution_time            = 300
 number_of_processes_for_frontend_testing = 20
 number_of_processes_for_batch_testing    = 20
 
+#
+# WARNING: This option is experimental and all edge cases are not fully tested, don't use, or use with care.
+# This option allow to customize the number of retries of a test if it is not finished aftert max_zonemaster_execution_time seconds.
+#
+#maximal_number_of_retries=3
```
matsduf: Since the default number is 0, the value should be 0.

Better language, I think (and better with shorter lines):
```
# WARNING: This option is experimental and all edge cases are not fully tested.
# Do not use it (keep the default value "0"), or use it with care.
```

mtoma: I don't think the suggested value for this option should be 0 in the .ini file.
As the default values is 0 (in the code) activating this option and putting there a value of 0 makes no sense.
The warning also applies only for values other than 0.

mattias-p: In my experience it's conventional to keep the default value commented out. But since you make the effort to state the default value explicitly I don't think it's very important.

PR #545, opened 2019-10-08
Fixes issue #213 - Internal error details are leaked

mtoma:
Fixes issue #213 - Internal error details are leaked

This fixes specifically the problem described in this issue. Another option could be to add eval blocks to all API calls and return something like "Unhandled error in API_CALL_METHOD" to all of them.


lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -41,7 +41,7 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = "$params->{db}"->new();
         };
-        die "$@ \n" if $@;
+        die "Failed to initialize the SQLite database backend module \n" if $@;
```
matsduf: I expect some general message through the API and the detailed message to a log file. Is it always SQLite problem in this case?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -50,7 +50,7 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = $backend_module->new();
         };
-        die "$@ \n" if $@;
+        die "Failed to initialize the database backend module \n" if $@;
```
matsduf: Here the message seems general, but the details should go to a log fiel.

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: This test is only true if the full perm module is given as input. This only happens during testing from the backend unit test scripts.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -41,7 +41,10 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = "$params->{db}"->new();
         };
-        die "$@ \n" if $@;
+        if ($@) {
+            warn "Failed to initialize the [$params->{db}] database backend module: [$@] \n";
+            die "Failed to initialize the [$params->{db}] database backend module \n";
```
matsduf: What will go back to the client? What will go a log file?

mattias-p:
This PR misses the point with regard to #213. The point is not to find all the places where we have source paths in exceptions, but to catch all unhandled exceptions whether we expected them or not.

I took your branch here and "accidentally" introduce a bug in `version_info`. Now it's started throwing exceptions some of the time. When that happens I get the following:
```json
{"jsonrpc":"2.0","id":1,"error":{"message":"uncaught exception from random function call at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/RPCAPI.pm line 69.\n","code":-32603}}
```

The proper solution is to catch all exceptions in the zonemaster_backend_rpcapi.psgi, get them into the log file and return an `"Internal error"` as prescribed by the JSON-RPC 2.0 specification.

> Note: I collected the above JSON output using the `zmb` tool from #534. I think it's quite handy when I'm making changes in the Backend. I would really appreciate your review of that PR, by the way.

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: The warn goes to the log file.
The die goes to the API client.

matsduf: I feel that the message to the API client is too long and contains too much information. I suggest "INTERNAL SYSTEM ERROR" plus some code, and the same code in the log message. That code could be a fixed code "INTERNAL SYSTEM ERROR # 5011" or more like a ticket ID which could be a random number that goes into both messages.

mtoma:
> This PR misses the point with regard to #213. The point is not to find all the places where we have source paths in exceptions, but to catch all unhandled exceptions whether we expected them or not.
> 
> I took your branch here and "accidentally" introduced a bug in `version_info`. Now it's started throwing exceptions some of the time. When that happens I get the following:
> 
> ```json
> {"jsonrpc":"2.0","id":1,"error":{"message":"uncaught exception from random function call at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/RPCAPI.pm line 69.\n","code":-32603}}
> ```
> 
> The proper solution is to catch all exceptions that bubbles up to zonemaster_backend_rpcapi.psgi, get them into the log file and return an `"Internal error"` as prescribed by the JSON-RPC 2.0 specification.
> 
> > Note: I collected the above JSON output using the `zmb` tool from #534. I think that tool is quite handy when I'm making changes in the Backend. I would really appreciate your review of that PR, by the way.

The best I can do is to add an eval block to all methods in the RPCAPI.pm module.
Would that be ok for you?

matsduf:
> The best I can do is to add an eval block to all methods in the RPCAPI.pm module.
> Would that be ok for you?

@mtoma, do you see any problem or drawbacks with the proposed solution?

mattias-p:
> The best I can do is to add an eval block to all methods in the RPCAPI.pm module.
> Would that be ok for you?

I thought about that approach and I think it would work. It would cover most of the code we execute and I believe `die "Internal error\n"` in the RPCAPI methods would result in a compliant error response.

Another approach I thought about is wrapping the `handle_psgi` call in an eval block. I believe that too would work.

mattias-p:
> I'm not so sure about the handle_psgi method. I tough it must have an eval block inside that method that catches and builds the response objects.

That rings a bell. I think I looked at that way back and then forgot about it. Maybe eval blocks in the RPCAPI methods is the way to go.

lib/Zonemaster/Backend/RPCAPI.pm
mattias-p: If this becomes a JSON-RPC error response with code -32603 (which seems appropriate) the message should be `Internal error` as per the [JSON-RPC specification](https://www.jsonrpc.org/specification#error_object).

This goes for the other calls to `die` in this sub and in `handle_exception` as well.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -50,29 +53,51 @@ sub new {
...
     return \%ver;
 }
 
 $json_schemas{profile_names} = joi->object->strict;
 sub profile_names {
     my ($self) = @_;
 
-    my @profiles = Zonemaster::Backend::Config->load_config()->ListPublicProfiles();
+    my @profiles;
+    eval {
```
mattias-p: I sure wish we had a more reasonable JSON-RPC library so we didn't have to jump through all these hoops...

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -97,89 +129,99 @@ sub get_data_from_parent_zone {
...
+            }
+            else {
                 return (
                     $dn,
                     {
-                        status  => 'nok',
-                        message => encode_entities( "The domain name is not a valid IDNA string and cannot be converted to an A-label" )
+                        status => 'nok',
+                        message =>
+                        encode_entities( "$type contains non-ascii characters and IDNA conversion is not installed" )
```
matsduf:
```
"$type contains non-ascii characters and IDNA conversion is not installed"
```
Suggest instead:
```
"$type contains non-ascii characters and IDNA is not supported"
```

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -97,89 +129,99 @@ sub get_data_from_parent_zone {
...
+
+        if( $dn !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
             return (
                 $dn,
                 {
-                    status => 'nok',
-                    message =>
-                      encode_entities( "$type contains non-ascii characters and IDNA conversion is not installed" )
+                status  => 'nok',
+                message => encode_entities( "The domain name contains unauthorized character(s)")
```
matsduf:
```
"The domain name contains unauthorized character(s)"
```
Suggest instead:
```
"The domain name contains character(s) not supported"
```

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -97,89 +129,99 @@ sub get_data_from_parent_zone {
...
-    }
 
-    my @res;
-    @res = Zonemaster::Engine::Test::Basic->basic00($dn);
-    if (@res != 0) {
-        return ( $dn, { status => 'nok', message => encode_entities( "$type name or label outside allowed length" ) } );
+        my @res;
+        @res = Zonemaster::Engine::Test::Basic->basic00($dn);
+        if (@res != 0) {
+            return ( $dn, { status => 'nok', message => encode_entities( "$type name or label outside allowed length" ) } );
```
matsduf:
```
"$type name or label outside allowed length"
```
Suggest instead:
```
"$type name or label is too long"
```

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -50,29 +53,52 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = $backend_module->new();
         };
-        die "$@ \n" if $@;
+        if ($@) {
+            warn "Failed to initialize the database backend module: [$@] \n";
+            die "Failed to initialize the database backend module \n" if $@;
```
matsduf: I feel that the message to the API client is too long and contains too much information. I suggest "INTERNAL SYSTEM ERROR" or something.

mattias-p: @matsduf Please see my comment about (standard) JSON-RPC error messages above.

matsduf: > @matsduf Please see my comment about (standard) JSON-RPC error messages above.

"Internal error" à la RPC is fine.

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Is there an update required here?

mattias-p: No, I was just whining. I probably should have kept it to myself.

matsduf:
@mtoma, can you please check this PR and update?

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in commit 416b48b

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in commit 416b48b

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in commit 416b48b

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in commit 416b48b

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -208,97 +251,105 @@ sub _check_domain {
...
 
-            return { status => 'nok', message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
-              unless ( !$ns_ip->{ip}
-                || Zonemaster::Engine::Net::IP::ip_is_ipv4( $ns_ip->{ip} )
-                || Zonemaster::Engine::Net::IP::ip_is_ipv6( $ns_ip->{ip} ) );
-        }
+                return { status => 'nok', message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
+                unless ( !$ns_ip->{ip}
+                    || Zonemaster::Engine::Net::IP::ip_is_ipv4( $ns_ip->{ip} )
+                    || Zonemaster::Engine::Net::IP::ip_is_ipv6( $ns_ip->{ip} ) );
```
pnax, line 321: Aren't this two `return ... unless ...` statements doing exactly the same thing ?
I think the second one is a refactoring of the first one and therefore should be kept.

matsduf: One problem is that Net::IP accepts "network" as valid address. Therefore you must have two tests. The first test will catch network, e.g. "10.2.1", that the second would accept. The first one will accept "355.955.2.1", but that will be caught by the second test.

It looks odd so I guess some comment in the code is needed.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -57,29 +60,52 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = $backend_module->new( { config => $config } );
         };
-        die "$@ \n" if $@;
+        if ($@) {
+            warn "Internal error #002: Failed to initialize the database backend module: [$@] \n";
+            die "Internal error #002 \n" if $@;
+        }
```
matsduf, line 66: Why not a call to handle_exception() here so that all cases are done in the same way?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -48,7 +48,10 @@ sub new {
             die "$@ \n" if $@;
             $self->{db} = "$params->{db}"->new( { config => $config } );
         };
-        die "$@ \n" if $@;
+        if ($@) {
+            warn "Internal error #001: Failed to initialize the [$params->{db}] database backend module: [$@] \n";
+            die "Internal error #001 \n";
+        }
```
matsduf, line 54: Why not a call to handle_exception() here so that all cases are done in the same way?

lib/Zonemaster/Backend/RPCAPI.pm
matsduf, line 321: Please add a comment that the two tests of the provided IP address are complementary, but not the same. Both are needed.

matsduf (changes requested):
I think it looks good, but I have three comments that I want an answer on.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -435,20 +503,46 @@ sub get_test_results {
...
+                if ( $res->{message} =~ /policy\.json/ ) {
+                    my ( $policy ) = ( $res->{message} =~ /\s(\/.*)$/ );
+                    if ( $policy ) {
+                        my $policy_description = 'DEFAULT POLICY';
+                        $policy_description = 'SOME OTHER POLICY' if ( $policy =~ /some\/other\/policy\/path/ );
+                        $res->{message} =~ s/$policy/$policy_description/;
+                    }
+                    else {
+                        $res->{message} = 'UNKNOWN POLICY FORMAT';
+                    }
```
mattias-p, line 523: This is (almost) repeated from above.

In fact now that I look at it there's lots of code that looks like it could be thrown out here. We don't have config and policy anymore, so we shouldn't need special handling of messages related to those either.

lib/Zonemaster/Backend/RPCAPI.pm
mattias-p: I second this.

Also, we cannot have the '#NNN' format. If you paste one of those into an issue on Github, Github mangles it into a link to an issue or a pull request. It's really completely unworkable.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -115,91 +149,102 @@ sub get_data_from_parent_zone {
...
+        if (ref \$params eq "SCALAR") {
+            $domain = $params;
+        } else {
+            $domain = $params->{"domain"};
+        }
 
-    my ( $dn, $dn_syntax ) = $self->_check_domain( $domain, 'Domain name' );
-    return $dn_syntax if ( $dn_syntax->{status} eq 'nok' );
+        my ( $dn, $dn_syntax ) = $self->_check_domain( $domain, 'Domain name' );
+        return $dn_syntax if ( $dn_syntax->{status} eq 'nok' );
```
mattias-p, line 161: This won't make `get_data_from_parent_zone` return. It only terminates the eval block with $dn_syntax as return value.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -403,52 +472,61 @@ sub get_test_results {
...
-                }
-                else {
-                    $res->{message} = 'UNKNOWN POLICY FORMAT';
-                }
+    eval{
+        $test_info = $self->{db}->test_results( $params->{id} );
+        foreach my $test_res ( @{ $test_info->{results} } ) {
+            my $res;
+            if ( $test_res->{module} eq 'NAMESERVER' ) {
+                $res->{ns} = ( $test_res->{args}->{ns} ) ? ( $test_res->{args}->{ns} ) : ( 'All' );
```
matsduf, line 482: What does this line of code do? If the module is NAMESERVER and the argument in the msgid is "{ns}" then the value is used if it exists, else the string "All"? Why and when?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -403,52 +472,61 @@ sub get_test_results {
...
-            elsif ( $res->{message} =~ /config\.json/ ) {
-                my ( $config ) = ( $res->{message} =~ /\s(\/.*)$/ );
-                if ( $config ) {
-                    my $config_description = 'DEFAULT CONFIGURATION';
-                    $config_description = 'SOME OTHER CONFIGURATION' if ( $config =~ /some\/other\/configuration\/path/ );
-                    $res->{message} =~ s/$config/$config_description/;
+
+            $res->{module} = $test_res->{module};
+            $res->{message} = $translator->translate_tag( $test_res, $params->{language} ) . "\n";
+            $res->{message} =~ s/,/, /isg;
```
matsduf, line 493: Is this meant to expand "," to ", " in an argument? But that will add an extra " " if there is a "," in the msgid, won't it?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -403,52 +472,61 @@ sub get_test_results {
...
+                        $policy_description = 'SOME OTHER POLICY' if ( $policy =~ /some\/other\/policy\/path/ );
+                        $res->{message} =~ s/$policy/$policy_description/;
+                    }
+                    else {
+                        $res->{message} = 'UNKNOWN POLICY FORMAT';
+                    }
                 }
-                else {
-                    $res->{message} = 'UNKNOWN CONFIG FORMAT';
+                elsif ( $res->{message} =~ /config\.json/ ) {
```
matsduf, line 509: There is no "config.json", is there? There is only "config.ini".

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: This is outside of the scope of this pull request.

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Outside of the scope of this PR

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Outside of the scope of this PR

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in dcf35a6

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in dcf35a6

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in dcf35a6

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in dcf35a6

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Comment added in dcf35a6

lib/Zonemaster/Backend/RPCAPI.pm
mtoma: Fixed in 416b48b

mattias-p (changes requested):
The use of return statements inside eval blocks needs to be fixed.
The exception IDs need a format that plays well with Github.

sandoche2k:
@mattias-p if it is possible, it will be good you make a PR with the suggestions and let @mtoma accept it ?

mtoma:
> The use of return statements inside eval blocks needs to be fixed.
> The exception IDs need a format that plays well with Github.

What do you mean by "return statements inside eval blocks needs to be fixed"? That's what the `my $result = eval { };` precisely does.
All IDs do now have the `#` removed.

mattias-p (approved):

mattias-p:
@mtoma I'm sorry for the confusion. I got mixed up about this PR and #578 and when I realized my mistake I didn't notice that you already had implemented the changes I requested.

matsduf (approved):

matsduf:
@mtoma, please merge as soon as possible.

PR #552, opened 2019-10-25
Empty

mattias-p:
Just checking.


mattias-p:
This fails alright, but not for the expected reasons. The errors look unrelated to the merging of zonemaster/zonemaster-ldns#85.

PR #556, opened 2019-11-01
Update instructions after dropping Debian 8

mattias-p:


mattias-p:
Libparallel-forkmanager-perl is available in a sufficient version for both Ubuntu releases. Apt should be available there as well.

PR #560, opened 2019-11-01
Updates in prepartion for release v2019.2

matsduf:


matsduf:
Must be update if anything is merged.

Changes
```diff
@@ -1,5 +1,31 @@
 Release history for Zonemaster component Zonemaster-Backend
 
+
+v5.0.0 2019-11-15
+
+ [API change]
+ - The database schema has been changed (#544). Create a new
```
mattias-p: I'm not convinced the database schema should be considered a public API. I think it's a better idea to tout Zonemaster::Backend::DB as the API and treat the database schema as encapsulated internal state.

Changes
```diff
@@ -1,5 +1,31 @@
...
+   database or migrate it will tool found in scripts folder
+
+ [Changes]
+ - Update instructions after dropping Debian 8 (#556)
+ - Updated Travis config after update of zonemaster-ldns (#541)
+ - Added garbage collector for unfinished tests (#544, #525)
+ - Updating installation instructions for FreeBSD (#535)
+ - Added graceful shutdown for TestAgent (#533, #532
```
mattias-p: This ought to be listed as an addition.

matsduf: An addition is a change, isn't it?

Changes
matsduf: The important things is to lift it up.

matsduf:
@mattias-p and @sandoche2k, please re-review after change.

Changes
mattias-p: Not in semver terminology.

PR #563, opened 2019-11-06
Fix warning in test agent daemon

### LINKED ISSUE #557: Warning from zm-backend.sh
I'm getting warning that says `Variable "$caught_sigterm" will not stay shared` in the systemd log when running
```
sudo service zm-backend.sh start
sudo service zm-backend.sh status
```

The warning is technically correct. It is raised because `catch_sigterm` is mistakenly declared as a proper sub (inside `main`) but refers to a local variable declared in `main`. Declaring `catch_sigterm` as a closure will fix the warning.

`catch_sigterm` is never called outside of `main`, so there's no risk for bad behavior because of this. Nevertheless the warning should be fixed.

mattias-p:
Fixes #557.
Should work but I haven't tested it much.

### DIFF
--- a/script/zonemaster_backend_testagent
+++ b/script/zonemaster_backend_testagent
@@ -87,13 +87,14 @@ sub main {
     my $self = shift;
 
     my $caught_sigterm = 0;
-    local $SIG{TERM} = \&catch_sigterm;
-    sub catch_sigterm {
-        $SIG{TERM} = \&catch_sigterm;
+    my $catch_sigterm;
+    $catch_sigterm = sub {
+        $SIG{TERM} = $catch_sigterm;
         $caught_sigterm = 1;
         $log->notice( "Daemon caught SIGTERM" );
         return;
-    }
+    };
+    local $SIG{TERM} = $catch_sigterm;
 
     while ( !$caught_sigterm ) {
         $self->pm->reap_finished_children();    # Reaps terminated child processes


matsduf:
@mattias-p, what release do you want the milestone to be set for?

mattias-p:
I'm sure I assigned this to 2019.3 when I created it. I certainly intended to.

It's not a big change though it is a bit technical. I guess we could include it in 2019.2 too if we really want to get rid of the warning.

matsduf:
> It's not a big change though it is a bit technical. I guess we could include it in 2019.2 too if we really want to get rid of the warning.

I think we should keep it as v2019.3.

mattias-p:
On second thought the risk of keeping a bug by not merging this is greater than the risk of breaking something by merging it.

We should test the TestAgent's behavior on the current develop branch with regard to SIGTERM.

script/zonemaster_backend_testagent
```diff
@@ -87,13 +87,14 @@ sub main {
     my $self = shift;
 
     my $caught_sigterm = 0;
-    local $SIG{TERM} = \&catch_sigterm;
-    sub catch_sigterm {
-        $SIG{TERM} = \&catch_sigterm;
+    my $catch_sigterm;
+    $catch_sigterm = sub {
```
matsduf: Why not combine the two lines into
```
my $catch_sigterm = sub {
```

script/zonemaster_backend_testagent
```diff
@@ -87,13 +87,14 @@ sub main {
...
+        $SIG{TERM} = $catch_sigterm;
```
matsduf: Why is this needed?

mattias-p: On some systems the signal handler gets cleared automatically so you have to re-enable it every time. I'm not sure exactly which systems have this behavior so I just threw it in for good measure. In any case it doesn't hurt.

script/zonemaster_backend_testagent
mattias-p: Because the variable is referenced inside the closure.

Issue #549, opened 2019-10-18
Update installation instructions for Backend for Debian 10


matsduf:
Debian 10 is a new OS planned to be supported by release v2019.2. The installation instructions probably need to be adjusted for Debian 10. There could be more binary packages available, but there could also be other changes.

Debian 10 uses PostgreSQL 11.5, which is new to Zonemaster, and MariaDB 10.3 (10.1 is used for Debian 9).

matsduf:
Resolved by PR #564

mattias-p:
Fixed by #564.

Issue #520, opened 2019-05-28
Document the DB section in backend_config.ini


mattias-p:
We've started documenting the various configuration options of Backend but it's not complete.
This issue is limited to filling in the details of the DB section.

In the PR for issue #496 we should be able to add a line documenting the new configuration property.

mattias-p:
The documentation of valid values for the `DB.engine` property should clarify that `MySQL` is a valid value for MariaDB.

matsduf:
Resolved by #730.

Issue #434, opened 2018-05-28
Switch to MariaDB


mattias-p:
Today all our installation instructions use MySQL. In some cases we're taking installation files from third party repositories. After dropping support for Debian 7, all our supported OS versions ship with MariaDB. We should start using MariaDB instead of MySQL for all our installation instructions.

- [x] CentOS (#663)
- [x] Debian/Ubuntu (#564)
- [ ] FreeBSD (does not work on FreeBSD #664)

matsduf:
OK from F2F.

matsduf:
@mattias-p, is there anything preventing this to be implemented with v2020.1?

mattias-p:
Not that I know of.

matsduf:
This issue will be resolved by updating the supported database versions in the main (Zonemaster/Zonemaster) README (see zonemaster/zonemaster/pull/890) and by updating the installation instructions.

sandoche2k:
@matsduf if the install instructions for FreeBSD is also added for MariaDB, we can close this issue

matsduf:
> @matsduf if the install instructions for FreeBSD is also added for MariaDB, we can close this issue

I will fix that.

matsduf:
Zonemaster-Backend requires `DBD::mysql` but that cannot be installed together with MariaDB in FreeBSD (using FreeBSD packages). At least for this release MySQL is kept for FreeBSD.

@mattias-p and @sandoche2k, should this be closed and some new ticket be created?

sandoche2k:
+1

mattias-p:
I added a task for each OS to the issue description and added links to the PR where they were resolved.

matsduf:
> I added a task for each OS to the issue description and added links to the PR where they were resolved.

I created a separate issue for FreeBSD and updated the issue description. Is that how you want it?

mattias-p:
> I created a separate issue for FreeBSD and updated the issue description. Is that how you want it?

My idea was to use a task list to track what OSes are done migrating to MariaDB.

matsduf:
> My idea was to use a task list to track what OSes are done migrating to MariaDB.

I removed the tick for FreeBSD. Can we close this issue?

sandoche2k:
+1

matsduf:
We moved to MariaDB on all OSs except FreeBSD, which it is not supported in our setup. An issue for FreeBSD has been created, #664. We can now close this issue.

PR #564, opened 2019-11-06
Installation on Debian

mattias-p:
The `mysql-server` package isn't available on Debian 10, but `mariadb-server` is available on all of Debian 9 (stretch), Debian 10 (buster), Ubuntu 16.04 (xenial) and Ubuntu 18.04 (bionic), so this PR switches over to install that package instead.

I removed the step to verify the PostgreSQL version because we don't support anything that gives you PostgreSQL <= 9.2 by default.

I updated the config file editing steps to make the result look more tidy.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -106,7 +106,7 @@ of operating system version and database engine version is supported.
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sudo sed -i '/\bengine\b/ s/=.*/=MySQL/' /etc/zonemaster/backend_config.ini
+sudo sed -i '/\bengine\b/ s/=.*/= MySQL/' /etc/zonemaster/backend_config.ini
 ```
 
 Install, configure and start database engine (and Perl bindings):
@@ -142,7 +142,7 @@ mysql --user=root --password < ./initial-mysql.sql
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sudo sed -i '/\bengine\b/ s/=.*/=PostgreSQL/' /etc/zonemaster/backend_config.ini
+sudo sed -i '/\bengine\b/ s/=.*/= PostgreSQL/' /etc/zonemaster/backend_config.ini
 ```
 
 Add PostgreSQL package repository needed to get the appropriate PostgreSQL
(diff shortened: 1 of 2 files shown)


matsduf:
If we move from MySQL to MariaDB we have to update the README in Zonemaster/Zonemaster.

The configuration parameter still says MySQL, but with this change we should a note that "MySQL" in the configuration file means MySQL or MariaDB.

mattias-p:
It's common knowledge that MySQL and MariaDB are closely related. Moreover the instruction is clear on how to install and configure the Backend to get it working.

In any case we should keep "MySQL" as a valid value for the configuration parameter. I guess we could make it an alias for "MariaDB" - or the other way around. I guess this would make sense if we completely moved away from MySQL. Otherwise I think we can keep the "MySQL" value as is.

mattias-p:
> Should we go to MariaDB on CentOS and FreeBSD too?

This is the subject of #434.

mattias-p:
@matsduf, @sandoche2k, @mtoma, Please review.

docs/Installation.md
```diff
@@ -277,18 +277,18 @@ sudo install -v -m 755 ./zm-backend.sh /etc/init.d/
...
 of operating system version and database engine version is supported.
 
-#### 4.2.1 Instructions for MySQL (Debian)
+#### 4.2.1 Instructions for MariaDB (Debian)
 
 Configure Zonemaster::Backend to use the correct database engine:
 
 ```sh
-sudo sed -i '/\bengine\b/ s/=.*/=MySQL/' /etc/zonemaster/backend_config.ini
+sudo sed -i '/\bengine\b/ s/=.*/= MySQL/' /etc/zonemaster/backend_config.ini
```
matsduf: The headline says MariaDB but the configuration file says MySQL. To avoid that the user starts wondering if that is a typo I think it should be stated, that "Yes, in the configuration file it should say 'MySQL'".

mattias-p: I think we can safely wait with doing that until we get the first report of a confused user.

matsduf: A comment does not hurt, does it?

mattias-p: In this case I don't think it helps anyone.

matsduf: Since it does not hurt, please add it.

sandoche2k: I do think you should add MariaDB is a fork of MySQL. There are many who does not know it.

mattias-p: This information fits better in the [configuration documentation](
https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db-section) than here. I've updated #520 so we don't forget to include this clarification.

matsduf: Then we should have a reference to that configuration document.

matsduf:
Resolves issue #549

PR #565, opened 2019-11-06
Compatibility with systemd-tmpfiles

### LINKED ISSUE #513: Backend not starting after reboot on Debian
Both daemons (RPCAPI and Agent) start just fine when following the instruction on Debian, but when you reboot the machine they aren't automatically started, despite the `sudo update-rc.d zm-backend.sh defaults` command. Presumably this is because Debian has switched to systemd, and we haven't adapted.

mattias-p:
I had strange behaviors with the RPCAPI sometimes not starting and when I dug into it I stumbled upon a solution to the reboot problem. The trick is to tell systemd-tmpfiles to (re)create `/var/run/zonemaster` at start up. The directory goes away because it resides on a tmpfs partition.

Fixes #513.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -56,6 +56,7 @@ share/cleanup-postgres.sql
 share/initial-mysql.sql
 share/initial-postgres.sql
 share/starman-zm-backend-RPC-API.conf
+share/tmpfiles.conf
 share/travis_mysql_backend_config.ini
 share/travis_postgresql_backend_config.ini
 share/travis_sqlite_backend_config.ini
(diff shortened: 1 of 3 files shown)


mattias-p:
This PR only adds systemd-tmpfiles integration to Debian and Ubuntu.

On FreeBSD we can keep creating /var/run/zonemaster as part of the installation instructions. There's no point in installing the systemd-tmpfiles configuration file on FreeBSD either.

I haven't heard anything on this subject for CentOS. Maybe it just works there? If it ain't broke I don't think it should be fixed.

mattias-p:
@matsduf, @sandoche2k, @mtoma, please review.

PR #571, opened 2019-12-13
Fixes #570 - UTF-8 issue for idn domain names

### LINKED ISSUE #570: UTF-8 issue for idn domain names
The answer for an idn domain name like café.fr comes back from the backend with a double utf8 encoding.

mtoma:
Issue #570.

### DIFF
--- a/lib/Zonemaster/Backend/DB/PostgreSQL.pm
+++ b/lib/Zonemaster/Backend/DB/PostgreSQL.pm
@@ -186,6 +186,8 @@ sub test_results {
         my ( $hrefs ) = $dbh->selectall_hashref( "SELECT id, hash_id, creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time, params, results FROM test_results WHERE $id_field=?", $id_field, undef, $test_id );
         $result            = $hrefs->{$test_id};
         
+        # This workaround is needed to properly handle all versions of perl and the DBD::Pg module 
+        # More details in the zonemaster backend issue #570
         if (utf8::is_utf8($result->{params}) ) {
                 $result->{params}  = decode_json( encode_utf8($result->{params}) );
         }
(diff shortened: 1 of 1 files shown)


mattias-p:
I'm surprised by this approach. I have a couple of questions.

1. Is #570 specific to PostgreSQL?

2. Are we inserting strings with mixed encodings into the `test_results.params` and `test_results.results` fields? Do we have a good reason for doing so? Are we sure that we never mix encodings in a single JSON blob?

3. The documentation for [is_utf8](https://perldoc.perl.org/utf8.html#Utility-functions) and [encode_utf8](https://perldoc.perl.org/Encode.html#Basic-methods) discourages the usages seen here. Is this a rare exception to those recommendations?

mtoma:
> I'm surprised by this approach. I have a couple of questions.
> 
> 1. Is #570 specific to PostgreSQL?
> 2. Are we inserting strings with mixed encodings into the `test_results.params` and `test_results.results` fields? Do we have a good reason for doing so? Are we sure that we never mix encodings in a single JSON blob?
> 3. The documentation for [is_utf8](https://perldoc.perl.org/utf8.html#Utility-functions) and [encode_utf8](https://perldoc.perl.org/Encode.html#Basic-methods) discourages the usages seen here. Is this a rare exception to those recommendations?

Indeed it is a rare exception because I see no other way to make one singe code work for all versions of perl and the DBD::Pg module.
The encoding you get from a select call has changed over time (not sure which are the exact DBD::Pg module versions) at one point there were coming as encoded and the encoding was to be done in the calling code with newer versions of perl the decoding was done by the module itself internally what was a problem for JSON data containing utf8 strings. It might have changed again now.
This is why I proposed the submited patch.

matsduf:
@mtoma, could you add a short documentation in the module to explain the code? Just a few lines is enough, maybe including a reference to this PR and other sources.

sandoche2k:
@mattias-p can you review?

matsduf:
@mtoma, can you merge now?

PR #577, opened 2020-03-19
Added missing use JSON:PP

mtoma:
When testing #525 it seems there is an use JSON::PP missing


sandoche2k:
@mtoma can you merge?

PR #579, opened 2020-03-19
Use one service script per daemon on System V

### LINKED ISSUE #576: Split System V service script
The `zm-backend.sh` script should be split into `zm-rpcapi.sysv` and `zm-testagent.sysv`.
The installation instruction will need to be updated accordingly.
No upgrade instruction should be needed as long as we treat the zonemaster_backend_rpcapi.psgi and zonemaster_backend_testagent as public interfaces.

This affects CentOS, Debian and Ubuntu. The same separation has been implemented on FreeBSD for a long time.

mattias-p:
Resolves #576.

Let's see if we can get this into the pending release.


share/zm-rpcapi.sysv
```diff
@@ -0,0 +1,51 @@
...
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: A JSON-RPC frontend for Zonemaster Backend
+# Description:       zm-rpcapi lets you add new tests and check for results in
+#                    the the Zonemaster Backend database
+### END INIT INFO
+
+BINDIR=${ZM_BACKEND_BINDIR:-/usr/local/bin}
+LOGFILE=${ZM_BACKEND_LOGDIR:-/var/log/zonemaster}/zm-starman-error.log
+PIDFILE=${ZM_BACKEND_PIDDIR:-/var/run/zonemaster}/zm-starman.pid
```
matsduf: I think it will be transparent to the users to let the default file names be "zonemaster_backend_rcpapi-error.log" and "zonemaster_backend_rcpapi.pid". And it would match the testagent file names better.

share/zm-rpcapi.sysv
```diff
@@ -0,0 +1,51 @@
...
+LISTENIP=${ZM_BACKEND_LISTENIP:-127.0.0.1}
+USER=${ZM_BACKEND_USER:-zonemaster}
+GROUP=${ZM_BACKEND_GROUP:-zonemaster}
+
+STARMAN=`PATH="$PATH:/usr/local/bin" /usr/bin/which starman`
+
+start () {
+    $STARMAN --user=$USER --group=$GROUP --error-log=$LOGFILE --pid=$PIDFILE --listen=$LISTENIP:5000 --preload-app --daemonize $BINDIR/zonemaster_backend_rpcapi.psgi
```
matsduf: Having port 5000 as default is fine, but I think it would make sense to create:

LISTENPORT=${ZM_BACKEND_LISTEN_PORT:-5000}

and to refer to that as well.

share/zm-rpcapi.sysv
mattias-p: I wanted to limit the scope of the change, but maybe this is a good time to change the file names after all.
Don't you think it would be better to make the files highlight their relation to the service script? I.e. naming the files like `zm-rpcapi.pid` etc.

share/zm-rpcapi.sysv
mattias-p: I wanted to limit the scope of the change. But sure, I'll add that.

mattias-p: By the way, the environment variable overrides are intended to only be used when developing Zonemaster-Backend. Though if someone wants to use them in some other context, I won't mind.

share/zm-rpcapi.sysv
matsduf: I agree. That is even better.

matsduf:
@mtoma, please review again.

sandoche2k:
@mattias-p merge?

PR #581, opened 2020-03-25
RPCAPI script lacks the status command

### LINKED ISSUE #580: Does not return Status in the installation instruction
CentOS-7, MySQL 
Section "3.3 Service configuration and startup (CentOS)" of the installation instruction
```
sudo /etc/init.d/zm-rpcapi status
```

mattias-p:
Fixes #580.

Regarding `sudo` for the testagent status command that I mentioned, I opted not to remove it right now. I haven't tested but I suspect sudo may be needed because of (silly?) implementation details.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -223,7 +223,6 @@ sudo /etc/init.d/zm-testagent start
 Check that the service has started:
 
 ```sh
-sudo /etc/init.d/zm-rpcapi status
 sudo /etc/init.d/zm-testagent status
 ```
 *Does not return any status as of now*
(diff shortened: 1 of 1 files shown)


matsduf:
Do you mean that the fix is to remove the command the installation instruction? Is a status command not possible?

mattias-p:
The old status command never checked the rpcapi daemon so this is no worse than what we had before. Do you want to include #582 in 2019.2?

matsduf:
>  Do you want to include #582 in 2019.2?

Yes, I think that would be really good.

mattias-p:
Closing this in favor of #586.

PR #584, opened 2020-03-25
Installation document updated for FreeBSD

matsduf:
* Updated with all direct dependencies
* All direct dependencies available as binary package installed with `pkg install`
* Direct dependency not available as binary package explicitly listed to be installed by `cpanm`

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -524,7 +524,7 @@ sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= PostgreSQL/' /usr/local/etc/zonemaster
 Install, configure and start database engine (and Perl bindings):
 
 ```sh
-pkg install postgresql95-server p5-DBD-Pg
+pkg install databases/postgresql11-server databases/p5-DBD-Pg
 sysrc postgresql_enable="YES"
 service postgresql initdb
 service postgresql start
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -501,6 +501,7 @@ exit;
 Initialize the database (and give the root password when prompted):
 
 ```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
 mysql -u root -p < ./initial-mysql.sql
 ```
 
@@ -533,6 +534,7 @@ service postgresql start
 Initialize the database:
 
 ```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
 psql -U pgsql -f ./initial-postgres.sql template1
 ```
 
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -402,7 +402,7 @@ su -l
 Install dependencies available from binary packages:
 
 ```sh
-pkg install p5-Class-Method-Modifiers p5-Config-IniFiles p5-Daemon-Control p5-DBI p5-File-ShareDir p5-File-Slurp p5-HTML-Parser p5-IO-CaptureOutput p5-JSON-PP p5-JSON-RPC p5-Moose p5-Parallel-ForkManager p5-Plack p5-Plack-Middleware-Debug p5-Role-Tiny p5-Router-Simple p5-Starman p5-String-ShellQuote
+pkg install p5-Class-Method-Modifiers p5-Config-IniFiles p5-Daemon-Control p5-DBI p5-File-ShareDir p5-File-Slurp p5-HTML-Parser p5-IO-CaptureOutput p5-JSON-PP p5-JSON-RPC p5-Moose p5-Parallel-ForkManager p5-Plack p5-Plack-Middleware-Debug p5-Role-Tiny p5-Router-Simple p5-Starman p5-String-ShellQuote net-mgmt/p5-Net-IP-XS databases/p5-DBD-SQLite devel/p5-Log-Dispatch devel/p5-Params-ValidationCompiler devel/p5-Log-Any devel/p5-Log-Any-Adapter-Dispatch
```
mattias-p: devel/p5-Params-ValidationCompiler should not be added. It's not a direct dependency, but rather a transitive one via Log-Dispatch.

matsduf: Removed again.

matsduf:
@mattias-p and @sandoche2k, please review.

docs/Installation.md
```diff
@@ -402,20 +402,27 @@ su -l
...
 pkg install curl
 ```
 
+Verify remaining dependencies:
+
+> This step is not needed for installation. It is only here for the developers of Zonemaster.
+
+```sh
+cpanm --scandeps Zonemaster-Backend-x.y.z.tar.gz # Normally you skip this step.
+```
```
mattias-p, line 420: This step should include an instruction on what to do with the command output. I.e. that the modules listed in the following `cpanm` command should match the set of direct dependencies in the output, and that transitive dependencies should be ignored.

Also, the step belongs in the ReleaseProcess document, rather than here in Installation.

matsduf: I agree that this is not the best place. As the release process document is set up there is no obvious place for it, but I'll remove it from here and create an issue to integrate it into some updated document.

matsduf:
@mattias-p and @sandoche2k, please re-review.

PR #595, opened 2020-04-02
Stricter handling of the language tag

matsduf:
The API documentation is unclear if "language" is mandatory in "get_test_results" but in the implementation it is mandatory. Here we follow the implementation and make it mandatory in the documentation.

In the API documentation and in the implementation defines some strings (or rather start of strings) to be specific languages and the remaining strings are interpreted as English. The drawbacks of this behavior are:

* When new languages are added, a string that earlier was interpreted as English, e.g. "no", will if Norwegian is added be interpreted as Norwegian.
* When testing, if the the language string is misspelled it could be interpreted as translation is not working. It is better to get a clear signal that the string was wrong.

This PR introduces the following updates to the API document, and if accepted, a new PR will be created changing the behavior of the RPCAPI.

* Only accept defined language tags (strings)
* Explicitly state that language parameter must be included with
  "get_test_results"

This PR also implements the check in the code (update from earlier version of the PR). That change was in #596 before, which is now closed.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -274,7 +274,8 @@ I.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`.
 * Any string starting with `"fr"` is interpreted as French.
 * Any string starting with `"sv"` is interpreted as Swedish.
 * Any string starting with `"da"` is interpreted as Danish.
-* Any other string is interpreted as English.
+* Any string starting with `"en"` is interpreted as English.
+* Any other string is an error.
 
 
 ### Unsigned integer
@@ -651,6 +652,10 @@ Example request:
   }
 }
 ```
+
+Both `id` and `language` are mandatory parameters. The `id` parameter must match the `result` in
+the response to a `start_domain_test` call, and that test must have been completed. The `language`
+parameter must match a language string defined above.
 
 Example response:
 ```json
--- a/lib/Zonemaster/Backend/Translator.pm
+++ b/lib/Zonemaster/Backend/Translator.pm
@@ -16,8 +16,10 @@ extends 'Zonemaster::Engine::Translator';
 
 sub translate_tag {
     my ( $self, $entry, $browser_lang ) = @_;
-
     my $previous_locale = $self->locale;
+
+    # $browser_lang is set to the first two characters of the string in
+    # "language" in "get_test_results" call.
     if ( $browser_lang eq 'fr' ) {
         $self->locale( "fr_FR.UTF-8" );
     }
@@ -27,8 +29,11 @@ sub translate_tag {
     elsif ( $browser_lang eq 'da' ) {
         $self->locale( "da_DK.UTF-8" );
     }
-    else {
+    elsif ( $browser_lang eq 'en' ) {
         $self->locale( "en_US.UTF-8" );
+    }
+    else {
+	die "Invalid language string: '$browser_lang' ";
     }
 
     # Make locale really be set. Fix that makes translation work on FreeBSD 12.1. Solution copied from


matsduf:
There is no reason why Travis should fail on this. Will restart test after review.

docs/API.md
```diff
@@ -652,6 +653,10 @@ Example request:
 }
 ```
 
+Both `id` and `language` are mandatory parameters. The `id` parameter must match the `result` in
+the response to a `start_domain_test` call, and that test must have been completed. The `language`
+parameter must match a language string defined above.
+
```
mattias-p, line 659: This paragraph should be integrated into the Params section.

mattias-p:
I really like the idea of making this change. My only question is when we should do it. If we can limit the number of releases that break the API, that's a good thing. I have a feeling this is not the only thing in the API that we want to change.
In other words I don't think we should merge this before reviewing the current API for other worthwhile breaking changes that we're confident we can include in the same release as this.

mattias-p:
On second thought I'm not sure the API is the correct place to enforce the correct language values. I think it would make sense if we could make it depend on the available.
What I do like about this proposal is that English is a catchall for any unknown language string.
Maybe we should also add "default" as a valid value while we're at it?

mattias-p:
What changed in the force push? Was it just to fix a merge conflict?

You haven't responded to my comments.
Are we breaking the RPCAPI API this release?

matsduf:
> What changed in the force push? Was it just to fix a merge conflict?

Yes.
 
> You haven't responded to my comments.

I wanted to see some other reactions first.

> Are we breaking the RPCAPI API this release? Looking through the issues tagged for 2020.1 I can't find anything else like that.

I have not checked.

On a second thought I think it is a bad idea to have specific language tags in the Translator.pm, and in the API. I think that this should be something that is configured in the configuration file. For each language supported there should be a language tag and a locale setting, e.g. "'sv' -> 'sv_SE.UTF-8'" but it could instead be "'sv' -> 'sv_FI.UTF-8'". With such a model it would be easy to add another language to backend without changing any code. If RPC-API gets a call with a language not defined in the configuration file, it should report an error.

"en" is special, but it should be possible to disallow "en" in backend, which means that messages will be shown translated to some other language.

By default, as we deliver Zonemaster-Backend, it should be configured with language and locale that we have today, but in the configuration file.

mattias-p:
I agree that it's even better to specify the set of locales in the configuration. Ideally locales should be specified as a set rather than as a mapping. This would need some careful design though to preserve backwards compatibility.

matsduf:
> Ideally locales should be specified as a set rather than as a mapping.

Please explain or exemplify.

mattias-p:
Instead of a mapping like `{ "sv": "sv_SE.UTF-8", "fr": "fr_FR.UTF-8" }`, a set like `["sv_SE.UTF-8", "fr_FR.UTF-8"]` would be a simpler interface. 

N.B. I'm only using JSON syntax here to express the idea of mappings versus sets. I'm not suggesting that we use JSON syntax in the configuration format.

matsduf:
> Instead of a mapping like `{ "sv": "sv_SE.UTF-8", "fr": "fr_FR.UTF-8" }`, a set like `["sv_SE.UTF-8", "fr_FR.UTF-8"]` would be a simpler interface.

And the element "sv_SE.UTF-8" should then be used as "'sv' --> 'sv_SE.UTF-8'"? Yes, with correct documentation that should work.

matsduf:
> Instead of a mapping like `{ "sv": "sv_SE.UTF-8", "fr": "fr_FR.UTF-8" }`, a set like `["sv_SE.UTF-8", "fr_FR.UTF-8"]` would be a simpler interface.

Today we assume UTF-8, and we could make that an explicit requirement, and then the list could be simplified to `["sv_SE", "fr_FR", "en_US", "da_DK"]`.

mattias-p:
Ideally I think clients should specify locales, not languages, since messages change across locales with the same basic language. E.g. sv_SE vs sv_FI. We should be able to support that even if we don't provide any such PO files today.

This means that we'll need to add some special handling of "en", "da", "fr" and "sv" with deprecation warnings and migration instructions, and in the future remove the support for the special handling.

matsduf:
> Ideally I think clients should specify locales, not languages, since messages change across locales with the same basic language. E.g. sv_SE vs sv_FI. We should be able to support that even if we don't provide any such PO files today.

I agree with you, in principle, but in practice I do not think that we would have different messages in Swedish for Finland and Sweden, respectively.  Let us keep the language support as language support even if we use the gettext locale mechanism.

matsduf:
Replaced by #621.

PR #596, opened 2020-04-02
Accept only valid language strings. Break on undefined strings.

matsduf:
This PR requires that PR #595 has been accepted/merged. This PR implements strict check on language strings and only accepts predefined strings.


mattias-p:
These changes should be included into #595. It doesn't make sense to merge one without the other.

matsduf:
> These changes should be included into #595. It doesn't make sense to merge one without the other.

#595 is the specification, this one is the implementation. I felt this was the correct order.

mattias-p:
I agree that it's a very good idea to make the specification first and the implementation after. But it's better to merge them together.

matsduf:
The update in this PR has been moved to PR #595 on request. This PR is now closed.

PR #603, opened 2020-04-28
Fix failing unit test on Debian 10

mattias-p:
On Debian 10 #594 breaks a unit test, and therefore installation. I have not investigated this further. At this point in the release cycle I'm more comfortable just reverting that PR than trying to make a more elaborate fix.

This reverts commit 61dbab1e20bf3d70118bc568763a96c186cc8e2d.

### DIFF
--- a/lib/Zonemaster/Backend/Translator.pm
+++ b/lib/Zonemaster/Backend/Translator.pm
@@ -6,7 +6,6 @@ use 5.14.2;
 
 use Moose;
 use Encode;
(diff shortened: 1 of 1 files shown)


matsduf:
The suggested alternative is to make sure that all the relevant locale (sv_SE.UTF-8, fr_FR.UTF-8 etc) are enabled.

matsduf:
Resolved by updated installation instruction in #606.

PR #604, opened 2020-04-28
Updated CentOS installation instructions

sandoche2k:


docs/Installation.md
```diff
@@ -62,7 +62,7 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
 Install dependencies available from binary packages:
 
 ```sh
-sudo yum install perl-Module-Install perl-IO-CaptureOutput perl-String-ShellQuote redhat-lsb-core
+sudo yum install perl-Module-Install perl-IO-CaptureOutput perl-String-ShellQuote perl-Net-Server redhat-lsb-core
```
matsduf, line 65: I note that Net::Server is required by Starman, which is installed from CPAN.

sandoche2k: Doesn't seem to work on CentOS 8. I have tried multiple times wherein many other packages were needed to be installed.

sandoche2k:
@mattias-p can you review?

PR #606, opened 2020-04-29
Debian installation

mattias-p:
Installation failed because Backend depends on a certain set of locales - namely en_US.UTF-8, sv_SE.UTF-8, fr_FR.UTF-8 and da_DK.UTF-8. If they're not present Backend crashes. In particular this happens when tests are run at installation time, causing the installation to abort. This PR fixes that problem by making sure the locales are installed.

This PR also includes a couple of other small improvements.


sandoche2k:
@mattias-p merge?

PR #613, opened 2020-05-11
Updates for database patches

matsduf:
* Corrected SQL code for Mysql
* Updated Makefile.PL so that scripts are installed.
* Updated instructions to point at correct `backend_config.ini` file for FreeBSD.


Makefile.PL
```diff
@@ -45,6 +45,11 @@ install_share;
 install_script 'zonemaster_backend_rpcapi.psgi';
 install_script 'zonemaster_backend_testagent';
 
+install_script 'patch_db_mysql_for_backend_DB_version_lower_than_1.0.3.pl';
+install_script 'patch_db_mysql_for_backend_DB_version_lower_than_5.0.0.pl';
+install_script 'patch_db_postgresq_for_backend_DB_version_lower_than_1.0.3.pl';
+install_script 'patch_db_postgresq_for_backend_DB_version_lower_than_5.0.0.pl';
```
mattias-p, line 51: I'm not too fond of dumping these into the bin directory. I'd prefer if they went into the dist_dir instead. They'll interfere with tab completion of the `patch` command and zonemaster is not in their names.

matsduf: Fine. You only use those once, if at all. What are the lines in `Makefile.PL` then?

mattias-p: It's the `install_share` directive that's already in there. Just moving `patch_*` into the `share` directory and updating the MANIFEST file should do it.

matsduf: So
```
install_share 'patch_db_mysql_for_backend_DB_version_lower_than_1.0.3.pl';
install_share 'patch_db_mysql_for_backend_DB_version_lower_than_5.0.0.pl';
install_share 'patch_db_postgresq_for_backend_DB_version_lower_than_1.0.3.pl';
install_share 'patch_db_postgresq_for_backend_DB_version_lower_than_5.0.0.pl';
```
should do the trick? `Makefile.PL` will automatically find the source files?

I will do the update, but there is more to it. The instructions have to be updated too.

matsduf:
@sandoche2k, @mattias-p, @mtoma  and @pamasse, please review this so that I can create the fix release. Please approve if you do approve.

PR #620, opened 2020-06-02
System user UID/GID for zonemaster user/group under FreeBSD

matsduf:
Current installation instruction gives the zonemaster user a UID and a GID that are meant for normal users, not system users. For system users there are UIDs and GIDs reserved for specific programs (packages) and UIDs/GIDs that are still free. This patch will make the instruction give the zonemaster user a UID/GID that is from the free space. When Zonemaster-Backend is packaged for FreeBSD it could get a fixed UID/GID.

Read [here][porters-handbook-users-and-groups] about UIDs and GIDs for applications ported to FreeBSD.

To see the current status of the list of booked and free UIDs and GIDs go to [FreeBSD ports] and scroll down to `GIDs` and `UIDs` and select the revision number for the files.

[porters-handbook-users-and-groups]: https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/users-and-groups.html
[FreeBSD ports]: https://svnweb.freebsd.org/ports/head/

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -515,7 +515,8 @@ Unless they already exist, add `zonemaster` user and `zonemaster` group
 (the group is created automatically):
 
 ```sh
-pw useradd zonemaster -s /sbin/nologin -d /nonexistent -c "Zonemaster daemon user"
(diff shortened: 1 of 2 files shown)


mattias-p (approved):
This is good AFAICT.

PR #621, opened 2020-06-10
Stricter handling of the language tag (updated version)

matsduf:
_This description has been updated 2020-06-29 as the PR has been updated._

The API documentation is unclear if "language" is mandatory in "get_test_results" but in the implementation it is mandatory. Here we follow the implementation and make it mandatory in the documentation.

In the API documentation and in the implementation defines some strings (or rather start of strings) to be specific languages and the remaining strings are interpreted as English. The drawbacks of this behavior are:

* When new languages are added, a string that earlier was interpreted as English, e.g. "no", will if Norwegian is added be interpreted as Norwegian.
* When testing, if the the language string is misspelled it could be interpreted as translation is not working. It is better to get a clear signal that the string was wrong.

The acceptable language codes and the mapping between language code and is done in the code. To add a new language you have to update the code. Also, it is not possible to exclude some of the available languages.

This PR moves the configuration of languages to the configuration file. Languages and the main part of the locale string are set there. New languages can be added without updating the code.

Summary of changes:
* Languages are configured, not hard-coded.
  * If configuration is missing, en_US will be configured by default.
* The API now only accepts specific strings, not just strings.
* Using undefined language code will result in error, not in no translation.
* A new API method reports supported language strings.

This update is a changed API against the GUI, but a default installation will not change anything for the GUI since it already uses strings matching the default configuration.

This PR replaces #595.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -334,6 +334,7 @@ Example response:
 }
 ```
 
+
 #### `"result"`
 
 An object with the following properties:
@@ -374,12 +375,13 @@ Example response:
 }
 ```
 
+
 #### `"result"`
 
 An array of *Profile names* in lower case. `"default"` is always included.
 
 
-## API method: `translation_language_strings`
+## API method: `get_language_strings`
 
 Returns valid strings for [translation language] based on the setting in the
 configuration file.
@@ -389,7 +391,7 @@ Example request:
 {
   "jsonrpc": "2.0",
   "id": 1,
-  "method": "translation_language_strings"
+  "method": "get_language_strings"
 }
 ```
 
@@ -409,9 +411,11 @@ Example response:
 }
 ```
 
+
 #### `"result"`
 
 An array of strings for *translation language*. It is never empty.
+
 
 #### `"error"`
 
@@ -473,6 +477,7 @@ value `0.0.0.0` if the lookup returned no A or AAAA records.
 >
 > TODO: If the name resolves to two or more IPv4 address, how is that represented?
 >
+
 
 #### `"error"`
 
@@ -629,13 +634,15 @@ An object with the following properties:
 > TODO: Clarify the purpose of each `"params"` property.
 >
 
+
 #### `"result"`
 
 A *test id*. 
 
 If the test has been run with the same domain name within an interval of 10 mins (hard coded), 
 then the new request does not trigger a new test, but returns with the results of the last test
- 
+
+
 #### `"error"`
 
 * If the given `profile` is not among the [available profiles], a user
@@ -799,7 +806,6 @@ In the case of a test created with `add_batch_job`:
 >
 
 
-
 #### `"error"`
 
 >
@@ -936,9 +942,11 @@ An object with the following properties:
 * `"username"`: An *username*, required. The name of the user to add.
 * `"api_key"`: An *api key*, required. The API key for the user to add.
 
+
 #### `"result"`
 
 An integer. The value is equal to 1 if the registration is a success, or 0 if it failed.
+
 
 #### `"error"`
 >
@@ -1146,6 +1154,7 @@ Example response:
     }
 }
 ```
+
 
 #### `"params"`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -268,13 +268,19 @@ Example: "2017-12-18 07:56:17.156939"
 Basic data type: string
 
 A string of alphanumeric, hyphens, underscores, full stops and at-signs (`@`),
-of at least 1 and at most 30 characters.
-I.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`.
+of at least 2 and at most 30 characters,
+i.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`, where only the first 2
+characters are used.
 
-* Any string starting with `"fr"` is interpreted as French.
-* Any string starting with `"sv"` is interpreted as Swedish.
-* Any string starting with `"da"` is interpreted as Danish.
-* Any other string is interpreted as English.
+The 2 first characters must match the list of language
+codes defined in the configuration file. Any other string
+is an error.
+
+A default installation will will accept:
+* Any string starting with `"da"` (Danish).
+* Any string starting with `"en"` (English).
+* Any string starting with `"fr"` (French).
+* Any string starting with `"sv"` (Swedish).
 
 
 ### Unsigned integer
@@ -651,6 +657,10 @@ Example request:
   }
 }
 ```
+
+Both `id` and `language` are mandatory parameters. The `id` parameter must match the `result` in
+the response to a `start_domain_test` call, and that test must have been completed. The `language`
+parameter must match a language string defined above.
 
 Example response:
 ```json
--- a/docs/API.md
+++ b/docs/API.md
@@ -282,7 +282,7 @@ The two first characters of the `language tag` are intended to be an
 [ISO 639-1] two-character language code and the optional two last characters
 are intended to be an [ISO 3166-1 alpha-2] two-character country code.
 
-A default installation will will accept the following `language tags`:
+A default installation will accept the following `language tags`:
 * `da` or `da_DK` for Danish language.
 * `en` or `en_US` for English language.
 * `fr` or `fr_FR` for French language.
--- a/docs/API.md
+++ b/docs/API.md
@@ -267,20 +267,28 @@ Example: "2017-12-18 07:56:17.156939"
 
 Basic data type: string
 
-A string of alphanumeric, hyphens, underscores, full stops and at-signs (`@`),
-of at least 2 and at most 30 characters,
-i.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`, where only the first 2
-characters are used.
+A string of A-Z, a-z and underscores matching the regular expression
+`/^[a-z]{2}(_[A-Z]{2})?$/`.
 
-The 2 first characters must match the list of language
-codes defined in the configuration file. Any other string
-is an error.
+The two first characters of the translation language string are
+expected to be an [ISO 639-1] two-character language code and the
+optional two last characters are expected to be an [ISO 3166-1 alpha-2]
+two-character country code.
 
-A default installation will will accept:
-* Any string starting with `"da"` (Danish).
-* Any string starting with `"en"` (English).
-* Any string starting with `"fr"` (French).
-* Any string starting with `"sv"` (Swedish).
+The translation language string must match a language tag in the
+configuration file. If the language translation string is a
+two-character string (language code only), it only need to match
+the first two characters of the language tag in the configuration file,
+if that is unique (there is only one language tag starting with the same
+language code), else it is an error.
+
+Any other string is an error.
+
+A default installation will will accept the following strings:
+* `"da"` or `"da_DK"` for Danish language.
+* `"en"` or `"en_US"` for English language.
+* `"fr"` or `"fr_FR"` for French language.
+* `"sv"` or `"sv_SE"` for Swedish language.
 
 
 ### Unsigned integer
@@ -326,7 +334,6 @@ Example response:
 }
 ```
 
-
 #### `"result"`
 
 An object with the following properties:
@@ -367,10 +374,53 @@ Example response:
 }
 ```
 
-
 #### `"result"`
 
 An array of *Profile names* in lower case. `"default"` is always included.
+
+
+## API method: `translation_language_strings`
+
+Returns valid [translation language strings] given the setting in the
+configuration file.
+
+Example request:
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "translation_language_strings"
+}
+```
+
+Example response:
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": [
+    "en",
+    "en_US",
+    "fr",
+    "fr_FR",
+    "sv",
+    "sv_SE"
+  ]
+}
+```
+
+#### `"result"`
+
+An array of *translation language strings*. It is never empty.
+
+#### `"error"`
+
+>
+> TODO: List all possible error codes and describe what they mean enough for
+> clients to know how react to them. Or prevent RPCAPI from starting with
+> errors in the configuration file and make it not to reread the configuration
+> file while running.
+>
 
 
 ## API method: `get_host_by_name`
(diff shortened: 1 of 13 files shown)


docs/API.md
```diff
@@ -268,13 +268,19 @@ Example: "2017-12-18 07:56:17.156939"
 Basic data type: string
 
 A string of alphanumeric, hyphens, underscores, full stops and at-signs (`@`),
-of at least 1 and at most 30 characters.
-I.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`.
+of at least 2 and at most 30 characters,
+i.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`, where only the first 2
+characters are used.
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/API.md
```diff
@@ -268,13 +268,19 @@ Example: "2017-12-18 07:56:17.156939"
...
 
-* Any string starting with `"fr"` is interpreted as French.
-* Any string starting with `"sv"` is interpreted as Swedish.
-* Any string starting with `"da"` is interpreted as Danish.
-* Any other string is interpreted as English.
+The 2 first characters must match the list of language
+codes defined in the configuration file. Any other string
+is an error.
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/API.md
```diff
@@ -652,6 +658,10 @@ Example request:
 }
 ```
 
+Both `id` and `language` are mandatory parameters. The `id` parameter must match the `result` in
+the response to a `start_domain_test` call, and that test must have been completed. The `language`
+parameter must match a language string defined above.
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
 
 TBD
 
+## LANGUAGES
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
...
+
+The LANGUAGE section has one key, `lang`.
+
+The value must be a space separated list of locale setting
+for the available translation of messages without ".UTF-8"
+which is assumed.
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
...
+
+Adding a new language to the configuration requires that the
+equivalent MO file is added to Zonemaster-Engine at the correct
+place so that getext get retreive it. Removing a lanugage from
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
...
+the configuration file just blocks that language to be displayed.
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
...
+---------|------------------|--------------------|-------------
+Danish   | da               | da_DK              | da_DK.UTF-8
+English  | en               | en_US              | en_US.UTF-8
+French   | fr               | fr_FR              | fr_FR.UTF-8
+Swedish  | sv               | sv_SE              | sv_SE.UTF-8
+
+*) RPC-API just considers the two first characters of the language
+string and disregards the remaining.
+
+The same language code may not be used more than once.
```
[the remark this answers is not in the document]
matsduf: Removed.

docs/API.md
```diff
@@ -652,6 +658,10 @@ Example request:
 }
 ```
 
+Both `id` and `language` are mandatory parameters. The `id` parameter must match the `result` in
+the response to a `start_domain_test` call, and that test must have been completed. The `language`
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/Configuration.md
```diff
@@ -32,6 +32,45 @@ SQLite            | `SQLite`
...
+
+Default setting in the configuration file:
+
+```
+lang = da_DK en_US fr_FR sv_SE
+```
+
+If the section is empty, "en_US" is set by default.
```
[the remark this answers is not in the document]
matsduf: Updated.

docs/Configuration.md
```diff
@@ -32,6 +32,54 @@ SQLite            | `SQLite`
...
+[ISO 639-1] two-character language code and the two last characters
+are expected to be an [ISO 3166-1 alpha-2] two-character country code.
+A language tag is a locale setting for the available translation
+of messages without ".UTF-8", which is assumed.
+
+Adding a new language to the configuration requires that the
+equivalent MO file is added to Zonemaster-Engine at the correct
+place so that gettext get retrieve it. Removing a language from
+the configuration file just blocks that language from being
+allowed.
```
mattias-p, line 53: I think this paragraph should be split in two. First we should clarify that having a language tag in the configuration without providing the corresponding translation files is an error. Second we should link to the documentation for adding a new translation.

docs/Configuration.md
```diff
@@ -32,6 +32,54 @@ SQLite            | `SQLite`
...
+
+English is the Zonemaster default language, but can be blocked
```
mattias-p, line 55: `but can` -> `but it can`?

docs/Configuration.md
```diff
@@ -32,6 +32,54 @@ SQLite            | `SQLite`
...
+following languages.
+
+Language | Language tag value | Locale value used
+---------|--------------------|------------------
+Danish   | da_DK              | da_DK.UTF-8
+English  | en_US              | en_US.UTF-8
+French   | fr_FR              | fr_FR.UTF-8
+Swedish  | sv_SE              | sv_SE.UTF-8
+
+Repeated language tag values are just ignored.
```
mattias-p, line 69: Isn't it better to make it an error?

docs/Configuration.md
```diff
@@ -32,6 +32,54 @@ SQLite            | `SQLite`
...
+
+Setting in the default configuration file:
+
+```
+lang = da_DK en_US fr_FR sv_SE
+```
+
+If the lang key is empty or absent, the language tag value
+"en_US" is set by default.
```
mattias-p, line 78: `language tag` -> `lang`?

docs/Configuration.md
```diff
@@ -32,6 +32,54 @@ SQLite            | `SQLite`
...
+
+Each locale, including ".UTF-8", set in the configuration file
+must also be installed or activate on the system running the RPCAPI
+daemon for the translation to work correctly.
```
mattias-p, line 82: This paragraph should probably be removed, and all of its details be integrated in the either part of the split paragraph I requested above, or in the documentation on how to add a new translation.

docs/Configuration.md
matsduf: No. They are referred to as "language tags".

docs/Configuration.md
matsduf: If there was such documentation that would be fine. As it is now I think this is good information here.

docs/API.md
```diff
@@ -320,7 +334,6 @@ Example response:
 }
 ```
 
-
```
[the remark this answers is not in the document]
matsduf: I am not sure where it was, but now I have added an extra line before the headline in several places where there was only one empty line‚ and I removed one empty line where there was three empty lines before the headline. This should solve the issue.

docs/API.md
```diff
@@ -361,12 +374,55 @@ Example response:
...
 #### `"result"`
 
 An array of *Profile names* in lower case. `"default"` is always included.
 
 
+## API method: `translation_language_strings`
```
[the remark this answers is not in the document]
matsduf: This was the longest method name, but not much longer than the second longest. Now I have shortened it to "get_language_strings", which I feel is a better name too.

docs/API.md
```diff
@@ -361,12 +374,55 @@ Example response:
...
+An array of *translation language strings*. It is never empty.
+
+#### `"error"`
+
+>
+> TODO: List all possible error codes and describe what they mean enough for
+> clients to know how react to them. Or prevent RPCAPI from starting with
+> errors in the configuration file and make it not to reread the configuration
+> file while running.
+>
```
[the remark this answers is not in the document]
matsduf: Yes, but that requires some changes to how it is done today.

docs/Configuration.md
mattias-p: When you remove a tag from the list you expect it to be inactivated. Unless the list is very short you may not notice that you need to remove more than one copy of the tag in order to disable it. Making duplicates an error makes you notice when you introduce one.

docs/Configuration.md
mattias-p: Fair enough. We should return to this after zonemaster/zonemaster-engine#594 is completed.

docs/Configuration.md
mattias-p: We should come back to this after zonemaster/zonemaster-engine#594 is completed.

docs/Configuration.md
mattias-p: We should come back to this after zonemaster/zonemaster-enging#594 is completed.

mattias-p:
What do you think of renaming the *translation language* data type to *language tag*? I think that's a more fitting name.

docs/Configuration.md
matsduf: I do not see much of a difference, but I have updated to meet your suggestion.

matsduf:
We have two different, but very related things in this system, by this PR. We have the "things type 1" that we set in the configuration files. They have the format "xx_YY", e.g. "en_US sv_SE fr_FR"

We have the "things type 2" that are legal to be value for the "language" key in the `get_test_results` method, and that is a superset of the "things type 1", but derived from "things type 1". Exampel above would give "en_US en sv_SE sv fr_FR fr". 

The new `get_language_strings` method returns the "things type 2", e.g. "en_US en sv_SE sv fr_FR fr". 

I will update so that "things type 1" are called "locale tags" and "things type 2" are called "language tags". The new method will also be renamed.

docs/API.md
```diff
@@ -367,6 +379,52 @@ Example response:
 An array of *Profile names* in lower case. `"default"` is always included.
 
 
+## API method: `get_language_tags`
+
+Returns valid [language tags][language tag] generated from the setting in the
+configuration file.
```
mattias-p, line 385: The most important thing here is that the method returns the complete set of valid tags. How it is calculated is secondary.

docs/Configuration.md
```diff
@@ -32,6 +32,60 @@ SQLite            | `SQLite`
...
+`locale tags` where each tag must match the regular expression
+`/^[a-z]{2}_[A-Z]{2}$/`.
+
+The two first characters of a `locale tag` are intended to be an
+[ISO 639-1] two-character language code and the two last characters
+are intended to be an [ISO 3166-1 alpha-2] two-character country code.
+A `locale tag` is a locale setting for the available translation
+of messages without ".UTF-8", which is implied.
+
+Adding a new `locale tag` to the configuration requires that the
```
mattias-p, line 49: What kind of requirement is this? Is it a requirement to avoid unspecified behavior? It should be clarified what happens if the requirement is not met.

Maybe the "Each locale set" paragraph is the clarification? In that case that paragraph should be integrated here.

docs/Configuration.md
```diff
@@ -32,6 +32,60 @@ SQLite            | `SQLite`
...
+It is an error to repeate the same `locale tag`.
+
+Setting in the default configuration file:
+
+```
+locale = da_DK en_US fr_FR sv_SE
+```
+
+If the `locale` key is empty or absent, the `locale tag` value
+"en_US" is set by default.
```
mattias-p, line 84: This should be moved up and placed together with the other notes on the semantics of the locale property.

docs/API.md
```diff
@@ -263,25 +263,37 @@ Default database timestamp format: "Y-M-D H:M:S.ms".
...
+that is unique (there is only one `locale tag` starting with the same two
+characters), else it is an error.
+
+Any other string is an error.
+
+The two first characters of the `language tag` are intended to be an
+[ISO 639-1] two-character language code and the optional two last characters
+are intended to be an [ISO 3166-1 alpha-2] two-character country code.
+
+A default installation will will accept the following `language tags`:
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/API.md
matsduf: It is relevant information.

docs/Configuration.md
matsduf: I think it was fine, but I reworded it.

docs/Configuration.md
```diff
@@ -32,6 +32,60 @@ SQLite            | `SQLite`
...
+`/^[a-z]{2}_[A-Z]{2}$/`.
+
+The two first characters of a `locale tag` are intended to be an
+[ISO 639-1] two-character language code and the two last characters
+are intended to be an [ISO 3166-1 alpha-2] two-character country code.
+A `locale tag` is a locale setting for the available translation
+of messages without ".UTF-8", which is implied.
+
+Adding a new `locale tag` to the configuration requires that the
+equivalent .mo file is added to Zonemaster-Engine at the correct
```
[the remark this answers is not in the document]
matsduf: I do not see the problem, but I updated.

docs/Configuration.md
```diff
@@ -32,6 +32,60 @@ SQLite            | `SQLite`
...
+place so that gettext get retrieve it. See the
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/Configuration.md
```diff
@@ -32,6 +32,60 @@ SQLite            | `SQLite`
...
+following languages.
+
+Language | Locale tag value   | Locale value used
+---------|--------------------|------------------
+Danish   | da_DK              | da_DK.UTF-8
+English  | en_US              | en_US.UTF-8
+French   | fr_FR              | fr_FR.UTF-8
+Swedish  | sv_SE              | sv_SE.UTF-8
+
+It is an error to repeate the same `locale tag`.
```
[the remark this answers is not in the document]
matsduf: Fixed.

docs/Configuration.md
matsduf: I do not see the problem, but I moved it up.

matsduf:
This PR has been replaced by #629.

Issue #607, opened 2020-04-29
Remove old GEOLOCATION from ini and possibly code


matsduf:
GEOLOCATION is not used anymore. It is still there in the `ini` file, and possibly somewhere in the code. It should be removed.

```
[GEOLOCATION]
# Requires the Geo::IP Perl module to be installed                                                                                                              
#maxmind_isp_db_file  =                                                                                                                                         

# Requires the GeoIP2::Database::Reader Perl module to be installed                                                                                             
#maxmind_city_db_file =                                                                                                                                         
```

matsduf:
Will be resolved by #625.

PR #625, opened 2020-06-30
Removed unused GEOLOCATION

matsduf:
There are still traces of previous feature GEOLOCATION in default .ini file and documentation. Since the feature is long gone, these traces will be removed with this update. This resolves issue #607.


matsduf:
@sandoche2k and @mtoma, can you review? @mattias-p reviewed and approved before, but I had to push an updated commit to handle a conflict.

PR #626, opened 2020-07-15
Updates to zmb and zmtest

### LINKED ISSUE #536: Install zmb and zmtest by default
Having `zmb` and `zmtest` installed by default makes functional testing of Backen more convenient.

At the time of writing `zmb` and `zmtest` are deemed experimental. While they're deemed experimental they should not be installed in any directories commonly included in $PATH.

mattias-p:
Zmtest ought to be ported to Perl but I'm not confident I'd be able to fit that in before my vacation. So I opted to patch up the shell implementation for now.

With these changes the interfaces to both zmtest and zmb are extended with a few options but otherwise remain basically the same. Neither zmtest nor zmb should be considered to have stable interfaces at this point.

Fixes #536 and implements suggestions from #534.


mattias-p:
Good point. I made an update but I was too quick. I'll make another update.

mattias-p:
There. Now I also updated the `--help` texts.

PR #629, opened 2020-07-16
Updated language specification handling

matsduf:
This PR tries to handle the following issues in Zonemaster-Backend when it comes to selecting language for translation.

1. The API documentation is unclear if field `language` is mandatory in RPCAPI method `get_test_results` but in the implementation it is mandatory.
2. The string that can be used in the `language` field above can be quite long and can contain a broad range of characters, but in practice only the two first characters are considered. The rest is ignored.
3. If the two first characters do not match a supported language, then the error is just silently ignored, and the messages are shown untranslated (from English).
   * The GUI actually just uses a string of two characters, "da", "en", "fr" or "sv".
4. To add support for a new language or to remove support for a some language the code must be updated.

The goals of this PR are:

1. Update the API documentation so that field `language` is mandatory for method `get_test_results`.
2. Restrict the legal strings for `language` to a meaningful set without breaking the current GUI.
3. Move the configuration of supported languages from code to the configuration file to make it possible to add and remove languages without changing the code. It will be possible to add incoming Norwegian without hardcoding it in the code.

The changes of this PR can be found in the diffs, but here is a summary of the changes:

* API.md (defines the RPCAPI)
  * Clarified that the `language` field is mandatory for method `get_test_results`.
  * Defines the value of that field to be a "language tag".
  * Defines the format of a valid "language tag" (compatible with the GUI today).
  * Defines the relation to "locale tag" in the configuration.
* Configuration.md (defines the format of backend_config.ini)
  * Defines a new section, "LANGUAGE", and a new key "locale".
  * Defines a string, "locale tag", which defines a supported language with country code.
  * Specifies that the new key "locale" should a list of "locale tags".
* Code update
  * Adds methods to read the new section and key (LANGUAGE.locale) from the configuration file.
  * From the "locale tags" generate valid "language tags".
  * Use the list of "language tags" to validate the value in `language` in `get_test_results` API calls.
  * Connect selected "language tag" with correct locale setting for Zonemaster-Engine.
  * Return error for invalid "language tag".
  * Add new RCPAPI method to get the list of valid "language tags". This could both be used by GUI to discover supported languages and for verifying that configuration was as intended.
  * Defined "en_US" to be set as "locale tag" (and "en" and "en_US" as "language tags") if the new section and key is not present or empty in the configuration file.
* Updates of test cases (unit tests)
  * Corrected language tag in current tests.
  * Adds tests of non-legal language tags.
    
This update is a changed API against the GUI, but a default installation will not change anything for the GUI since it already uses strings matching the default configuration.

This PR replaces #621.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -263,25 +263,37 @@ Default database timestamp format: "Y-M-D H:M:S.ms".
 Example: "2017-12-18 07:56:17.156939"
 
 
-### Translation language
+### Language tag
 
 Basic data type: string
 
-A string of alphanumeric, hyphens, underscores, full stops and at-signs (`@`),
-of at least 1 and at most 30 characters.
-I.e. a string matching `/^[a-zA-Z0-9-_.@]{1,30}$/`.
+A string of A-Z, a-z and underscores matching the regular expression
+`/^[a-z]{2}(_[A-Z]{2})?$/`.
 
-* Any string starting with `"fr"` is interpreted as French.
-* Any string starting with `"sv"` is interpreted as Swedish.
-* Any string starting with `"da"` is interpreted as Danish.
-* Any other string is interpreted as English.
+The `language tag` must match a `locale tag` in the configuration file.
+If the `language tag` is a two-character string, it only needs to match the
+first two characters of the `locale tag` from the configuration file, if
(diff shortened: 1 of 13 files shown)


matsduf:
@sandoche2k, can you review this?

Issue #628, opened 2020-07-16
Feature requests for zmb and zmtest


matsduf:
Feature requests for `zmtest`:
* Add support for language (translation) to zmtest. That will make it easy to test language support.
* Add support to exclude messages of a certain level and lower, e.g. exclude INFO to just see important messages.

Feature requests for `zmb`
* Add support for the remaining API methods. Especially, it would be valuable to be able execute batch tests this way.

`zmtest` and `zmb` were added by #534 and #626.

mattias-p:
#744 should be solved before we add more features to zmtest.

matsduf:
#675 has extended zmb with new API method.

matsduf:
#825 will add functions for batch as requested in this issue.

PR #632, opened 2020-09-07
Language option for zmtest

mattias-p:
Adds `zmtest --lang LANG` option.

This also includes some small fixes remotely related documentation.

Implements part of #628.

### DIFF
--- a/share/zmtest
+++ b/share/zmtest
@@ -16,6 +16,7 @@ usage () {
     echo "  --noipv4              Run the test without ipv4." >&2
     echo "  --noipv6              Run the test without ipv6." >&2
     echo "  --lang LANG           A language tag. Default is en." >&2
+    echo "                        Valid values are determined by backend_config.ini." >&2
     exit "${status}"
 }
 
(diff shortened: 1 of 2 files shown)


share/zmtest
```diff
@@ -12,9 +12,10 @@ usage () {
     echo "Usage: zmtest [OPTIONS] DOMAIN" >&2
     echo >&2
     echo "Options:" >&2
-    echo "  -s URL --server=URL   Zonemaster Backend to query. Default is http://localhost:5000/" >&2
+    echo "  -s URL --server URL   Zonemaster Backend to query. Default is http://localhost:5000/" >&2
     echo "  --noipv4              Run the test without ipv4." >&2
     echo "  --noipv6              Run the test without ipv6." >&2
+    echo "  --lang LANG           A language tag. Default is en." >&2
```
matsduf, line 18: I suggest:

```
A language tag. Default is "en".
```

Plus a reference to `backend-config.ini` and its documentation for valid language tags.

matsduf (commented):
This is a great update. See my suggestions.

share/zmtest
mattias-p: I don't see how anyone would be confused by the lack of quotes. But if you insist I'll add them?

I'll add a note about valid parameter values.

matsduf: I think the quotes makes it clearer. And a note about valid parameters would be great. Thanks.

matsduf (approved):

PR #634, opened 2020-09-07
Fix failing test

mattias-p:
Travis was broken by zonemaster/zonemaster-engine#764. This fixes it again.

While this change works AFAICT, it doesn't seem like the best idea to use basic00 to check name and label length.


matsduf (approved):

matsduf:
> While this change works AFAICT, it doesn't seem like the best idea to use basic00 to check name and label length.

Does Backend really have to check label and name lenghth instead of letting engine checking that?

Issue #449, opened 2018-12-04
Avoid installing dependencies from CPAN in installation instruction


mattias-p:
Daemon::Control
* CentOS - Neither available in CentOS nor EPEL. We should consider migrating to something else.
* Ubuntu - Version 0.1.6 is available in Ubuntu 16.04. Our version requirement is listed as 0.1.7.

Net::IP::XS
* CentOS - Neither available in CentOS nor EPEL. Adopt same solution as zonemaster/zonemaster-engine#451.
* FreeBSD - We could install from Ports instead.
* Ubuntu - Not available on Ubuntu 16.04.

Starman
* CentOS - Neither available in CentOS nor EPEL. We should consider migrating to something else.

matsduf:
Please be consistent in the description above:

Package/Library
* CentOS - Bla bla
* Debian - Bla bla
* Ubuntu - Bla bla
* FreeBSD - Bla bla

Why should we install Net::IP::XS in Engine in one OS and in Backend in another? If we migrate to NetAddr::IP in CentOS, then we should do that for all OSs.

mattias-p:
@matsduf 
1. I agree that dependencies should be installed in the correct component. However I chose to limit the scope of this issue. Feel free to file a complementary issue about rectifying where dependencies are installed.
2. I'll make your suggested structural change to the description. 
3. Please do not refer to my statements as bla bla.

matsduf:
>     3. Please do not refer to my statements as bla bla.

It was just meant as a placeholder for text and nothing else. I'm sorry if you take it as something negative.

I suggest that we focus on Engine first and there try to see if we can include all dependencies of Engine as packages. Then we can eliminate references to the same dependencies in Backend and CLI.

matsduf:
Complete what is easy to complete, leave the rest as CPAN packages. Except for FreeBSD which should be 100% covered by binary packages.

mattias-p:
Closing this. The CPAN situation is under control on FreeBSD, which is the first distro we'll be packaging for. Once that is completed we'll create a new issue that is specific to the next OS in line.

PR #638, opened 2020-09-15
Install dependencies from EPEL instead of CPAN

mattias-p:
This a step towards resolving #449. It also speeds up installation on CentOS.
The Zonemaster-Engine installation instruction already adds the EPEL repo. This merely takes advantage of that.
Consideration has been taken to the fact that EPEL has somewhat different sets of packages for CentOS 7 and CentOS 8.

Edit: A couple of the packages aren't available through EPEL on CentOS 8, but they are available through the PowerTools repository, which we're also already prescribing in Zonemaster-Engine.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -24,7 +24,6 @@ requires
   'Net::IP::XS'                 => 0.14,
   'Parallel::ForkManager'       => 1.12,
   'Plack::Builder'              => 0,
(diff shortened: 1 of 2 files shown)


matsduf (approved):

mattias-p:
@sandoche2k Please review.

sandoche2k (approved):

PR #641, opened 2020-09-18
Remove direct dependency on Net::IP::XS

### LINKED ISSUE #193: Optional Net::IP::XS
Net::IP::XS is being made optional for Zonemaster Engine in dotse/zonemaster-engine#174. Doesn't the same logic apply to Zonemaster Backend? In that case it should be made optional here as well.

mattias-p:
Net::IP::XS will still be used (via Zonemaster::Engine::Net::IP) if it's installed.

Works towards solving #449 and completely fixes #193.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -21,7 +21,6 @@ requires
   'Log::Any::Adapter::Dispatch' => 0,
   'Log::Dispatch'               => 0,
   'Moose'                       => 2.04,
-  'Net::IP::XS'                 => 0.14,
   'Parallel::ForkManager'       => 1.12,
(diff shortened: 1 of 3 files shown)


matsduf (approved):

mattias-p:
@matsduf Do you think this shouldn't be included in 2020.1?

sandoche2k (approved):

matsduf:
> @matsduf Do you think this shouldn't be included in 2020.1?

Since it has been completed I think it could be moved to v2020.1 and merged.

matsduf:
This change depends on the feature introduced by zonemaster/zonemaster-engine#689

PR #645, opened 2020-09-29
Update Travis config

### LINKED ISSUE #551: Re-enable Perl 5.16 in Travis
#541 breaks Travis for Perl 5.16. Instead of postponing #541 (and zonemaster/zonemaster-ldns#85) until the next release we disable Perl 5.16 in Travis as part of #541. 

Perl 5.16 must be re-enabled in Travis.

mattias-p:
This PR contains three parts:
1. It re-enables Perl 5.16 in Travis. This fixes #551.
2. It also enables Perl 5.32 in Travis, which is needed because it's now default on FreeBSD.
3. It reduces the Travis build matrix sizes from 21 runs to 9 while still testing all Perl versions and all databases.


matsduf:
If the test of the database was just testing the database itself limiting to one Perl version would probably be unproblematic, but we have also the Perl client library. On the other hand, we have several versions of the database engines that we do not test with all versions of Perl.

mattias-p:
Yes it is definitely a decrease in test coverage. But the decrease is quite small. The decrease in testing time is substantial at 50%.

matsduf (approved):
This is a good change. The exact combination of OS, version of Perl and version of database engine is anyway testing during QA testing.

sandoche2k (approved):

PR #649, opened 2020-09-30
Fix warning in RPCAPI

### LINKED ISSUE #631: Use of uninitialized value in zm-rpcapi.log
I'm seeing lots of these warnings repeated in zm-rpcapi.log:
```
Use of uninitialized value in concatenation (.) or string at /usr/local/share/perl/5.26.1/Zonemaster/Backend/RPCAPI.pm line 389.
```
The file I'm looking at has 82371 of them distributed across 13 restarts.

For the record this particular installation is running Zonemaster::Backend 5.0.2 on Ubuntu 18.04 and MySQL.

mattias-p:
I think I've got this figured out now.

* `$entry->{string}` never seems to exist, but  `Zonemaster::Engine::Logger::Entry::string` does exist, and it actually makes sense to call that on proper `Z::E::L::Entry` objects.
* Calling bless to get an `Z::E::L::Entry` object doesn't work because the object doesn't get properly initialized we end up crashing. Instead we should just be good citizens and call `Z::E::L::Entry::new`.
* The check of `Z::E::Translator::data` is kind of useless. The `Z::E::T::translate_tag` performs essentially the same check and returns $entry->string if the check fails. It's better to just use the `Z::E::T::translate_tag` return value.
* The early return was unfortunate because that code path failed to reset the locale, so I fixed that as well.

Fixes #631.

### DIFF
--- a/lib/Zonemaster/Backend/Translator.pm
+++ b/lib/Zonemaster/Backend/Translator.pm
@@ -49,9 +49,7 @@ sub translate_tag {
     my $octets = Zonemaster::Engine::Translator::translate_tag( $self, $entry );
     $self->locale( $previous_locale );
 
-    return ( defined $octets )
-      ? decode_utf8( $octets )
-      : $entry->string;
+    return decode_utf8( $octets );
 }
 
 1;
--- a/lib/Zonemaster/Backend/Translator.pm
+++ b/lib/Zonemaster/Backend/Translator.pm
@@ -16,7 +16,7 @@ require Zonemaster::Engine::Logger::Entry;
 extends 'Zonemaster::Engine::Translator';
 
 sub translate_tag {
-    my ( $self, $entry, $browser_lang ) = @_;
+    my ( $self, $hashref, $browser_lang ) = @_;
     my %locale = Zonemaster::Backend::Config->load_config()->Language_Locale_hash();
     my $previous_locale = $self->locale;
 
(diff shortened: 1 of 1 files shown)


matsduf:
I have no objection to your changes, but I cannot see why they should be in the same PR.

mattias-p:
This fix is broken. I'll come back with an update.

lib/Zonemaster/Backend/Translator.pm
```diff
@@ -45,17 +45,13 @@ sub translate_tag {
...
-    my $octets = Zonemaster::Engine::Translator::translate_tag( $self, $blessed_entry );
+    my $entry = Zonemaster::Engine::Logger::Entry->new( %{ $hashref } );
+    my $octets = Zonemaster::Engine::Translator::translate_tag( $self, $entry );
     $self->locale( $previous_locale );
-    my $str = decode_utf8( $octets );
-    return $str;
+
+    return ( defined $octets )
+      ? decode_utf8( $octets )
+      : $entry->string;
```
matsduf, line 54: The following is Zonemaster::Engine::Translator::translate_tag:

```
sub translate_tag {
    my ( $self, $entry ) = @_;

    return $self->_translate_tag( $entry->module, $entry->tag, $entry->printable_args ) // $entry->string;
}
```

Won't it already return $entry->string if translation (_translate_tag) fails?

mattias-p: It will. I'll make an update.

matsduf (approved):

PR #650, opened 2020-09-30
Attach adapter to Log::Any in RPCAPI

### LINKED ISSUE #640: Hook up Log::Any in RPCAPI
Our library code emits log messages to Log::Any, but RPCAPI currently doesn't hook up a log adapter for Log::Any. Consequently those messages are dropped. The RPCAPI should hook up an adapter for Log::Any in such a way that Log::Any and Starman write their log messages to the same file.

mattias-p:
This fixes #640 in a minimalist way. We do get logging but the severity threshold is hard-coded to "warning".

For the moment we want to avoid setting it any lower because the config module keeps repeating that it reloaded the config file. This happens a lot!

Once we have the config reloading under control we also want the ability to configure the logging severity threshold, but that's for another time.

All the new dependencies added here are already being used in the test agent.

### DIFF
--- a/script/zonemaster_backend_rpcapi.psgi
+++ b/script/zonemaster_backend_rpcapi.psgi
@@ -6,14 +6,17 @@ our $VERSION = '1.1.0';
 
 use 5.14.2;
 
-use JSON::RPC::Dispatch;
-use Router::Simple::Declare;
+use English qw( $PID );
 use JSON::PP;
+use JSON::RPC::Dispatch;
+use Log::Any qw( $log );
+use Log::Any::Adapter;
+use Log::Dispatch;
 use POSIX;
-use Try::Tiny;
-
 use Plack::Builder;
 use Plack::Response;
+use Router::Simple::Declare;
+use Try::Tiny;
 
 BEGIN { $ENV{PERL_JSON_BACKEND} = 'JSON::PP' };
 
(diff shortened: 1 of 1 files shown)


matsduf:
> For the moment we want to avoid setting it any lower because the config module keeps repeating that it reloaded the config file. This happens a lot!
> 
> Once we have the config reloading under control we also want the ability to configure the logging severity threshold, but that's for another time.

Issue #651 has been created to change so that loading is only done at start-up of the daemons.

matsduf (commented):
In zonemaster_backend_testagent you can set the log level. I think that make sense for debugging. Would that complicate things a lot? "warning" could still be the default level unless you explicitly set something else.

mattias-p:
I completely agree that we want the ability to set the severity threshold for the rpcapi as well. At first I intended to include it in this PR, but I couldn't see from the Plack documentation how to do that in our case. I suggest that as a first step we only log warnings and more severe messages, and as a second step we make the threshold configurable.

script/zonemaster_backend_rpcapi.psgi
```diff
@@ -103,6 +107,34 @@ my $router = router {
 	};
 };
 
+# Returns a Log::Any-compatible log level string, or throws an exception.
+sub get_loglevel {
+    my $value = $ENV{ZM_BACKEND_RPCAPI_LOGLEVEL} || 'warning';
+    for my $method ( logging_methods(), logging_aliases() ) {
```
matsduf, line 113: There is already another environment variable used as an option to this script, "ZONEMASTER_BACKEND_CONFIG_FILE". I think it would make sense if both om them come from the same name space, i.e. both being "ZONEMASTER_BACKEND_*" or both being "ZM_BACKEND_*".

matsduf (approved):

matsduf:
Issue #653 has been created to update the start scripts to include the new variable.

PR #655, opened 2020-10-02
Clean up Zonemaster::Backend::Translator::translate_tag

mattias-p:
Handle the language parameter once per request in Z::B::RPCAPI::get_test_results instead of once per message. It also removes a call to Z::E::Translator::data that's useless after Z::B::Translator::translate_tag was updated in #649.


matsduf:
In the decription above it says it "was updated in #631" but it should be #649".

matsduf (approved):
This looks like a good change. Why do things many time when once is enough.

sandoche2k (approved):

PR #657, opened 2020-10-16
Update manifest

mattias-p:
Add forgotten file to manifest.


matsduf (approved):
The file was added by me in #620. It is a good reminder to always update MANIFEST or MANIFEST.SKIP.

PR #658, opened 2020-10-19
Updated smoke test

mattias-p:
I ran into a bug when trying to use `zmtest` in the smoke test.
For unknown reasons the message strings in the JSON output end with "\n". In `zmtest` we process the output through echo and jq. Echo interprets escape sequences by default. New-line characters aren't allowed in JSON. And jq throws up on invalid JSON. So zmtest crashes.

This PR:
1. Makes zmtest not interpret escape sequences in the JSON output.
2. Makes zmtest print the testid earlier to make sure it gets printed even the program crashes during result processing.
3. Makes the installation instruction smoke test use zmtest.

An old copyright statement is also removed.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -52,18 +52,18 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
 > **Note:** In addition to the normal dependencies, the post-installation
 > smoke test instruction assumes that you have curl installed.
 
-Optionally install jq (only needed for the post-installation smoke test):
-
-```sh
-sudo yum install jq
-```
-
 ## 3. Installation on CentOS
 
 ### 3.1 Install Zonemaster::Backend and related dependencies (CentOS)
 
 > **Note:** Zonemaster::LDNS and Zonemaster::Engine are not listed here as they
 > are dealt with in the [prerequisites](#prerequisites) section.
+
+Optionally install jq (only needed for the post-installation smoke test):
+
+```sh
+sudo yum install jq
+```
 
 Install dependencies available from binary packages:
 
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -49,9 +49,6 @@ Prerequisite for FreeBSD is that the package system is upadated and activated
 For details on supported versions of Perl, database engine and operating system
 for Zonemaster::Backend, see the [declaration of prerequisites].
 
-> **Note:** In addition to the normal dependencies, the post-installation
-> smoke test instruction assumes that you have curl installed.
-
 ## 3. Installation on CentOS
 
 ### 3.1 Install Zonemaster::Backend and related dependencies (CentOS)
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -684,8 +684,7 @@ API on localhost port 5000 as below. The command requires that `curl` and `jq` a
 
 
 ```sh
-cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
-./zmtest zonemaster.net
+`perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`/zmtest zonemaster.net
 ```
 
 The command is expected to immediately print out a testid,
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -687,11 +687,9 @@ cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backen
 ./zmtest zonemaster.net
 ```
 
-The command is expected to give an immediate JSON response similiar to:
-
-```json
-{"id":"1","jsonrpc":"2.0","result":{"zonemaster_backend":"2.0.2","zonemaster_engine":"v2.0.6"}}
-```
+The command is expected to immediately print out a testid,
+followed by a percentage ticking up from 0% to 100%.
+Once the number reaches 100% a JSON object is printed and zmtest terminates.
 
 
 ### 7.2. What to do next?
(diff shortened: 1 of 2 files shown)


docs/Installation.md
```diff
@@ -677,14 +683,13 @@ you should be able to use the
 API on localhost port 5000 as below. The command requires that `curl` is installed.
```
pnax, line 683:
```suggestion
API on localhost port 5000 as below. The command requires that `curl` and `jq` are installed.

```

pnax (dismissed):

docs/Installation.md
```diff
@@ -677,14 +683,13 @@ you should be able to use the
 API on localhost port 5000 as below. The command requires that `curl` is installed.
 
 ```sh
-curl -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"version_info","id":"1"}' http://localhost:5000/ && echo
+./zmtest zonemaster.net
```
matsduf, line 687: I suggest
```
$(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`)/zmtest zonemaster.net
```
instead so that you do not have to jump around in the file system.

matsduf:
> For unknown reasons the message strings in the JSON output end with "\n". In `zmtest` we process the output through echo and jq. Echo interprets escape sequences by default.

I guess it depends on `echo`. I guess that `printf` is more portable.

share/zmtest
```diff
@@ -4,35 +4,36 @@ bindir="$(dirname "$0")"
...
 
 zmb () {
     server_url="$1"; shift
     json="$("${ZMB}" --server="${server_url}" "$@" 2>&1)" || error 1 "method $1: ${json}"
-    json="$(echo "${json}" | "${JQ}" -S . 2>/dev/null)" || error 1 "method $1 did not return valid JSON output"
-    error="$(echo "${json}" | "${JQ}" -e .error 2>/dev/null)" && error 1 "method $1: ${error}"
-    echo "${json}"
+    json="$("${ECHO}" -E "${json}" | "${JQ}" -S . 2>/dev/null)" || error 1 "method $1 did not return valid JSON output"
+    error="$("${ECHO}" -E "${json}" | "${JQ}" -e .error 2>/dev/null)" && error 1 "method $1: ${error}"
+    "${ECHO}" -E "${json}"
```
matsduf, line 36: `-E` is not supported as an option by `/bin/echo` in FreeBSD. "\n" is nothing special either. 
```
$ /bin/echo -E text
-E text
$ /bin/echo "text\n"
text\n
```

matsduf (changes requested):
I think it is great to include `zm` in the smoke test. It makes it so much better. I think you should consider installing it (and `zmtest`) in the same bin directory as `zonemaster-cli` is installed in. If you think the names are too short and obscure, then consider `zonemaster-backend` and `zonemaster-backend-test`, respectively.

sandoche2k (approved):

docs/Installation.md
mattias-p: I implemented this.

share/zmtest
mattias-p: I replaced /bin/echo with printf.

mattias-p:
@matsduf Can you re-review?

docs/Installation.md
```diff
@@ -52,6 +52,12 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
...
 
+Optionally install jq (only needed for the post-installation smoke test):
+
+```sh
+sudo yum install jq
+```
+
 ## 3. Installation on CentOS
 
 ### 3.1 Install Zonemaster::Backend and related dependencies (CentOS)
```
matsduf, line 63: To be consistent with Debian and FreeBSD, under this headline, just after the "note" about Zonemaster::LDNS and Zonemaster::Engine add:

Optionally install Curl and jq (only needed for the post-installation smoke test):

```sh
sudo yum install curl jq
```

docs/Installation.md
```diff
@@ -52,6 +52,12 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
 > **Note:** In addition to the normal dependencies, the post-installation
 > smoke test instruction assumes that you have curl installed.
```
matsduf, line 53: These two lines are not needed since that is repeated for each OS.

docs/Installation.md
```diff
@@ -52,6 +52,12 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
 > **Note:** In addition to the normal dependencies, the post-installation
 > smoke test instruction assumes that you have curl installed.
 
+Optionally install jq (only needed for the post-installation smoke test):
+
+```sh
+sudo yum install jq
+```
+
```
matsduf, line 60: These six lines (CentOS specific) do not belong to the general Prerequisites section. Suggested to be replaced by the lines below.

matsduf (changes requested):

docs/Installation.md
mattias-p: The scope of this PR is to update the smoke test. It's not about fixing all problems we can find in the installation instruction.

docs/Installation.md
mattias-p: Right. I'm moving them.

docs/Installation.md
mattias-p: The scope of this PR is to update the smoke test. It's not about fixing all problems we can find in the installation instruction.

docs/Installation.md
mattias-p: I moved the misplaced jq installation step here. I'm not adding curl here because it isn't a new dependency. I haven't checked but I'm assuming it's installed by default on CentOS.

matsduf (approved):

PR #663, opened 2020-11-02
Remove MySQL and add MariaDB installation instructions

sandoche2k:


docs/Installation.md
```diff
@@ -64,6 +64,11 @@ Install dependencies available from binary packages:
 ```sh
 sudo yum install perl-Module-Install perl-IO-CaptureOutput perl-String-ShellQuote
 ```
+*  Only for CentOS 7 *
+
+```sh
+sudo yum install perl-Plack-Test
+```
```
matsduf, line 71: Plack::Test is not a direct dependency. If it is an indirect dependency we should not have to list it here. Furthermore, is it really true that it is only needed for CentOS 7?

mattias-p: Plack::Test is part of the Plack CPAN distribution which we do depend on. For some reason the CentOS packaging of the Plack CPAN distribution is split into two packages: perl-Plack and perl-Plack-Test. So I'm not surprised to see perl-Plack-Test among our dependencies. But Plack seems to be split the same way on both CentOS 7 and 8, and it's not obvious to me why perl-Plack-Test is only needed on one of them.

sandoche2k: Yes, needed only for CentOS7

mattias-p (approved):

matsduf (approved):

matsduf:
@sandoche2k, please merge.

PR #665, opened 2020-11-05
Update installation instruction

matsduf:
Some updates primarily for FreeBSD

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -631,7 +631,7 @@ sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= PostgreSQL/' /usr/local/etc/zonemaster
 Install, configure and start database engine (and Perl bindings):
 
 ```sh
-pkg install databases/postgresql12-server databases/p5-DBD-Pg
+pkg install postgresql12-server p5-DBD-Pg
 sysrc postgresql_enable="YES"
 service postgresql initdb
 service postgresql start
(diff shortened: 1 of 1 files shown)


sandoche2k (approved):

docs/Installation.md
```diff
@@ -744,6 +740,9 @@ Zonemaster-Backend and keep the database, then you might have to upgrade the
 database to use it with the new version of Zonemaster-Backend. Please see the
 [upgrade][README.md-upgrade] information.
 
+If you keep the database but upgrade the database engine, read the instructions
+for the database engine if any steps have to be take first.
+
```
pnax, line 745: To be honest, I had to read this paragraph several times to understand it. I was wondering where those « instructions for the database engine » were until I realized they were not part of Zonemaster's documentation.
Maybe this is just me, but a section title might help to grasp what the idea is here, or rephrasing with something like « refer to your database engine instructions to upgrade it before proceeding with ... (the thing to proceed with) »

matsduf: Well, I could just remove it also. It is not necessary. Maybe more confusing than helpful. On my private installation of Zonemaster with Postgresql on FreeBSD I lost the database when FreeBSD upgraded Postgresql from 11 to 12.

matsduf: I removed that text.

mattias-p (approved):

PR #681, opened 2020-12-08
Fix issue 179

### LINKED ISSUE #179: Ensure that the SQLite backend can be used as a fully featured backend for Zonemaster

mtoma:
First version of full support for SQLite fixes #179

### DIFF
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -9,6 +9,7 @@ use Encode;
 use DBI qw(:utils);
 use JSON::PP;
 use Digest::MD5 qw(md5_hex);
+use Data::Dumper;
 
 use Zonemaster::Backend::Config;
 
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -65,6 +65,8 @@ sub create_db {
     $self->dbh->do(
         'CREATE TABLE test_results (
                          id integer PRIMARY KEY AUTOINCREMENT,
+                         hash_id VARCHAR(16) DEFAULT NULL,
+                         domain VARCHAR(255) NOT NULL,
                          batch_id integer NULL,
                          creation_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
                          test_start_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL,
@@ -74,7 +76,9 @@ sub create_db {
                          progress integer DEFAULT 0,
                          params_deterministic_hash character varying(32),
                          params text NOT NULL,
-                         results text DEFAULT NULL
+                         results text DEFAULT NULL,
+                         undelegated boolean NOT NULL DEFAULT false,
+                         nb_retries integer NOT NULL DEFAULT 0
                )
      '
     ) or die "SQLite Fatal error: " . $self->dbh->errstr() . "\n";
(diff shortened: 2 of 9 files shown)


mtoma:
To test this pull request the script/create_db_sqllite.pl needs to be called to initialize the database.
It expects to find the full path (including the file name) to the database in the 
```database_name``` parameter onf the backend_config.ini file
A suggestion is:
```zonemaster = /var/lib/zonemaster/db.sqlite```
or
```zonemaster = /tmp/db.sqlite``` for temporary testing purposes.
The user and password are unused for the SQLite backend.

**Comment by matsduf** [comment]
In FreeBSD the database directory should be `/var/db/zonemaster` and owned by the zonemaster user. The file name `db.sqlite` looks fine as the default name. Short and clear.

> This file needs to be readable and writable by both the starman RPCAPI backend and the zonemaster_test_agent script

That is the zonemaster user, isn't it?

**Comment by mtoma** [comment]
> That is the zonemaster user, isn't it?

This entirely depends on the content of the startup script.

**Comment by matsduf** [comment]
> > That is the zonemaster user, isn't it?
> 
> This entirely depends on the content of the startup script.

And by default that is user:group "zonemaster:zonemaster".

**Comment by matsduf** [comment]
Not in MANIFEST: script/add-batch-job.pl
Not in MANIFEST: script/create_db_sqlite.pl

**Comment by matsduf** [comment]
> The user and password are unused for the SQLite backend.

@mtoma, what about `database_host`?

**Comment by mtoma** [comment]
> @mtoma, what about `database_host`?

Also unused

**Comment by matsduf** [comment]
In FreeBSD the command to create the Sqlite database is (after updating the `ini`  file):
```
ZONEMASTER_BACKEND_CONFIG_FILE="/usr/local/etc/zonemaster/backend_config.ini" perl create_db_sqlite.pl
```

**Comment by matsduf** [comment]
I let the directory be owned by `zonemaster:zonemaster`, but after the creation of the database I had to change ownership of the database file:
```
chown zonemaster:zonemaster /var/db/zonemaster/db.sqlite
```

**Comment by matsduf** [comment]
This seems to work as it should. It is just the documentation that is missing. If the Configuration.pm is updated with what should be in the `ini` file, then the installation instructions could be added in a second PR.

File: lib/Zonemaster/Backend/DB/SQLite.pm [hunk 1]
```diff
@@ -164,97 +165,124 @@ sub create_new_batch_job {
...
+            $batch_id,
+            $priority,
+            $queue,
+            $test_params_deterministic_hash,
+            $encoded_params,
+            $test_params->{domain},
+            ($test_params->{nameservers})?(1):(0),
+        );
+        
+        $dbh->do(q[UPDATE test_results SET hash_id = ? WHERE params_deterministic_hash = ?], undef, substr(md5_hex(time().rand()), 0, 16), $test_params_deterministic_hash);
```
mattias-p, line 215: I don't see why this update is separated from the insert statement above. Am I missing something?

script/add-batch-job.pl
```diff
@@ -0,0 +1,40 @@
...
+use warnings;
+
+use Data::Dumper;
+use Encode qw[decode_utf8];
+use Zonemaster::Backend::RPCAPI;
+use Digest::MD5 qw(md5_hex);
+
+binmode STDOUT, ':utf8';
+
+my $e = Zonemaster::Backend::RPCAPI->new;
```
mattias-p, line 13: Could you update this variable name to be something meaningful?

script/add-batch-job.pl
```diff
@@ -0,0 +1,40 @@
...
+
+$e->add_api_user({ username => 'test_user', api_key => 'API_KEY_01'});
+
+$e->add_batch_job(
+	{
+		client_id      => 'Add Script',
+		client_version => '1.0',
+		username       => 'test_user',
+		api_key        => 'API_KEY_01',
+		test_params    => {
```
mattias-p, line 31: Could you fix the indentation here?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -9,6 +9,7 @@ use Encode;
 use DBI qw(:utils);
 use JSON::PP;
 use Digest::MD5 qw(md5_hex);
+use Data::Dumper;
```
mattias-p, line 12: Could you remove this unused import?

There are more of these below.

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -164,97 +165,124 @@ sub create_new_batch_job {
...
 
-    my ( $params_json ) = $self->dbh->selectrow_array( "SELECT params FROM test_results WHERE id=$test_id" );
-    eval { $result = decode_json( $params_json ); };
-    die "$@ \n" if $@;
+    my $result;
+    eval {
+        $result = decode_json( $params_json );
+    };
+    
+    warn "decoding of params_json failed (testi_id: [$test_id]):".Dumper($params_json) if $@;
```
mattias-p, line 262: `test_id` is misspelled.

Also, if you change `warn` to use `Log::Any::warn()` instead, the log message gets prepended with a nice header.

mattias-p (commented):
I have a couple of nits

matsduf (changes requested):
When I use the Sqlite backend tests are never considered too old. If I test "zonemaster.net" and test again after, say 30 minutes, I will get the old result, not a new test.

It is set in the `ini` file, but how is it used?

lib/Zonemaster/Backend/DB/SQLite.pm
mtoma: It is separated from the insert (unlike mysql and postgresql) because SQLite doesn't have the required hashing functions to be able to initialize the hash_id column properly.

lib/Zonemaster/Backend/DB/MySQL.pm
mtoma: I added this one @mattias-p because surprisingly it is used in the generation of error logs (search to file for Dumper)

matsduf: @mtoma, you have added the library to `MySQL.pm`. What has changed to make it necessary to add it? MySQL backend is not the target of this PR. And if it should be there, shouldn't it be added for PostgreSQL too?

mattias-p: @mtoma, right. I can see it's needed in SQLite.pm as well. I didn't expect that, though it does make sense.

matsduf (changes requested):
`script/create_db_sqlite.pl` does not get installed. It should be installed at the same directory as `initial-mysql.sql` and `initial-postgres.sql`. I also suggest that it is renamed to `initial-sqlite.pl` to have consistent naming.

matsduf:
> When I use the Sqlite backend tests are never considered too old. If I test "zonemaster.net" and test again after, say 30 minutes, I will get the old result, not a new test.

This issue seems to be resolved.

matsduf (changes requested):
I see some issue when I test. I see the following in the log file:
```
Internal error 012: Unexpected error in the get_test_results API call: [malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /usr/local/lib/perl5/site_perl/Zonemaster/Backend/DB/SQLite.pm line 281.] 
```

matsduf (changes requested):
With other databases, `creation_time` is UTC even if processes are not started with UTC. With SQLite, `creation_time` is the local time.

matsduf:
> I see some issue when I test. I see the following in the log file:
> 
> ```
> Internal error 012: Unexpected error in the get_test_results API call: [malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /usr/local/lib/perl5/site_perl/Zonemaster/Backend/DB/SQLite.pm line 281.] 
> ```

I have investigated a little bit further and have found some bad glitches.

If you start a test (`start_domain_test`), you get a test ID. If you then run `test_progress` it will take a while before "result" is 100 (100%) and you can get the result (`get_test_results`). This is normal.

If you try to start a new test on the same domain within 10 minutes, you get the same test ID as before and "result" is already 100%. This is also normal.

If you then try start a test again on the same domain after those 10 minutes, you get a new test ID, i.e. a new test is preformed. But what is wrong is that when you run `test_progress` on that you get "result" with 100% even though the test is **NOT** completed. `get_test_results` will return "message: Internal error 012". But if you retry `get_test_results` a few times, then finally you get the result, the test has completed.

matsduf:
Replaced by #690.

mattias-p:
Should this be closed if it's been replaced?

matsduf:
This was automatically considered to be merged when #690 was merged since its commits are a subset of #690.

PR #684, opened 2020-12-15
Document all backend config keys

pnax:
Add documentation to all backend keys as found in the `/lib/Zonemaster/Backend/Config.pm` file.

It appears that some keys are not used anywhere in the code.

This addresses #520 

This is a work in progress, some used keys are not documented yet.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -1,8 +1,9 @@
 # Configuration
 
-Zonemaster *Backend* is configured as a whole from `/etc/zonemaster/backend_config.ini`
-(CentOS, Debian and Ubuntu) or `/usr/local/etc/zonemaster/backend_config.ini`
-(FreeBSD).
+Zonemaster *Backend* is configured in
+`/etc/zonemaster/backend_config.ini` (CentOS, Debian and Ubuntu) or
+`/usr/local/etc/zonemaster/backend_config.ini` (FreeBSD). Following
+[Installation instructions] will create the file with factory settings.
 
 Each section in `backend_config.ini` is documented below.
 
@@ -31,34 +32,29 @@ SQLite            | `SQLite`
 ### user
 
 The name of the user with sufficient permission to access the database.
-Default value: `zonemaster`.
 
 Ignored by the SQLite database engine.
 
 ### password
 
-The password of the configured user. Default value: `zonemaster`.
+The password of the configured user.
 
 Ignored by the SQLite database engine.
 
 ### database_host
 
-The host name of the machine on which the engine is running. Default
-value: `localhost`.
+The host name of the machine on which the engine is running.
 
 Ignored by the SQLite database engine.
 
 ### database_name
 
 The name of the database to use, except for SQLite database engine where
-it holds the full path to the SQLite database file. Default value:
-`zonemaster` or `/var/lib/zonemaster/db.sqlite`
-(`/var/db/zonemaster/db.sqlite` on FreeBSD).
+it holds the full path to the SQLite database file.
 
 ### polling_interval
 
-Time in seconds between database lookups by Test Agent. Default value:
-`0.5`.
+Time in seconds between database lookups by Test Agent.
 
 
 ## LANGUAGE section
@@ -223,6 +219,7 @@ Integer working as a label to associate a test to a specific Test Agent.
 
 --------
 
+[Installation instructions]:          Installation.md
 [Default JSON profile file]:          https://github.com/zonemaster/zonemaster-engine/blob/master/share/profile.json
 [ISO 3166-1 alpha-2]:                 https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
 [ISO 639-1]:                          https://en.wikipedia.org/wiki/ISO_639-1
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -31,22 +31,34 @@ SQLite            | `SQLite`
 ### user
 
 The name of the user with sufficient permission to access the database.
+Default value: `zonemaster`.
+
+Ignored by the SQLite database engine.
 
 ### password
 
-The password of the configured user.
+The password of the configured user. Default value: `zonemaster`.
+
+Ignored by the SQLite database engine.
 
 ### database_host
 
-The host name of the machine on which the engine is running.
+The host name of the machine on which the engine is running. Default
+value: `localhost`.
+
+Ignored by the SQLite database engine.
 
 ### database_name
 
-The name of the database to use.
+The name of the database to use, except for SQLite database engine where
+it holds the full path to the SQLite database file. Default value:
+`zonemaster` or `/var/lib/zonemaster/db.sqlite`
+(`/var/db/zonemaster/db.sqlite` on FreeBSD).
 
 ### polling_interval
 
-not used anywhere besides in script/zonemaster_backend_testagent
+Time in seconds between database lookups by Test Agent. Default value:
+`0.5`.
 
 
 ## LANGUAGE section
@@ -185,16 +197,13 @@ The ZONEMASTER section has several keys :
 
 ### max_zonemaster_execution_time
 
-This key allows to define the interval in seconds to reload unfinished
-queries.
-This assume that after `max_zonemaster_execution_time`, if the test is
-not finished, it has fail. And it will be run again up to
-`maximal_number_of_retries`.
+Time in seconds before reporting an unfinished test as failed. Default
+value: `600`.
 
 ### maximal_number_of_retries
 
-Specifies the number of time a test is allowed to be run again if
-unfinished after `max_zonemaster_execution_time`.
+Number of time a test is allowed to be run again if unfinished after
+`max_zonemaster_execution_time`. Default value: `0`.
 
 ### number_of_professes_for_frontend_testing
 
@@ -210,7 +219,7 @@ used -> todo
 
 ### lock_on_queue
 
-used -> todo
+Integer working as a label to associate a test to a specific Test Agent.
 
 --------
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -193,7 +193,13 @@ used -> todo
 
 ### force_hash_id_use_in_API_starting_from_id
 
-used -> todo
+This parameter determines if it is possible to get the test results
+using an incremental numeric id (allowing easy enumeration of the whole
+database of results) or if a half-md5 hash id should be used instead. If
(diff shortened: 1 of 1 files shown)


mattias-p:
This looks great!
You're going well beyond what's required for #520 with this PR, but of course that's highly appreciated!
I also appreciate that you're including the unused keys and marking them as such.

There are two things I'd like to see added. At least for the DB section.
1. Some keys are only used in some contexts. E.g. `database_host` is only used if `engine` is either MySQL or PostgreSQL. It would be great if that was clarified in the documentation.
2. It would be great if the documentation clarified what values are legal for each of the keys.

docs/Configuration.md
```diff
@@ -146,7 +176,41 @@ to specifying a profile JSON file containing the entire
...
+`maximal_number_of_retries`.
+
+### maximal_number_of_retries
+
+Specifies the number of time a test is allowed to be run again if
+unfinished after `max_zonemaster_execution_time`.
+
+### number_of_professes_for_frontend_testing
+
+not used in the code
```
matsduf, line 201: Suggest:

Removed from all config files (normal and for Travis) and Config.pm. I do not think that will break anything.

Same thing for "number_of_processes_for_frontend_testing",

docs/Configuration.md
```diff
@@ -146,7 +176,41 @@ to specifying a profile JSON file containing the entire
...
+
+### number_of_professes_for_batch_testing
+
+not used in the code
```
matsduf, line 205: Suggest:

Removed from all config files (normal and for Travis) and Config.pm. I do not think that will break anything.

Same thing for "number_of_processes_for_batch_testing".

docs/Configuration.md
```diff
@@ -146,7 +176,41 @@ to specifying a profile JSON file containing the entire
...
+
+### force_hash_id_use_in_API_starting_from_id
+
+used -> todo
```
matsduf, line 209: @mtoma, can you give a description and legal values for this?

docs/Configuration.md
```diff
@@ -146,7 +176,41 @@ to specifying a profile JSON file containing the entire
...
+
+### lock_on_queue
+
+used -> todo
```
matsduf, line 213: @mtoma, can you give a description and legal values for this?

matsduf (changes requested):
This is great. Finally we get things documented.

docs/Configuration.md
```diff
@@ -109,14 +130,23 @@ Each locale set in the configuration file, including the implied
...
 
+
 ## LOG section
 
-TBD
+The LOG section has a unique key, `log_dir`.
+
+### log_dir
+
+not used in the code
```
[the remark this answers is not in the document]
pnax: Is your suggestion about saying that the key has been removed ? Or to actually remove the code related to this key in Config.pm ?

docs/Configuration.md
pnax: I realize there is a misspelling in the key name which is present in the code: `professes` -> `processes`

docs/Configuration.md
pnax: It looks like there is a description of the `queue` attribute in the [API documentation](https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#queue)

As I understand it, it is possible to use one database with different Test Agent, so this value works as a label to associate each test to its Test Agent.

pnax (commented):
And about the default values for the DB section, are they hard coded somewhere or only extracted from the `share/backend_config.ini` file ?

docs/Configuration.md
```diff
@@ -27,6 +28,26 @@ MySQL             | `MySQL`
 PostgreSQL        | `PostgreSQL`
 SQLite            | `SQLite`
 
+### user
+
+The name of the user with sufficient permission to access the database.
```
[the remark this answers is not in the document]
pnax: see https://github.com/zonemaster/zonemaster-backend/pull/684/commits/3110d32d76d60958a54de2e71d5dbc257eebb782

docs/Configuration.md
```diff
@@ -27,6 +28,38 @@ MySQL             | `MySQL`
...
+Default value: `zonemaster`.
```
mattias-p, line 34: I don't think it's right to say that "zonemaster" is the default value of the "user" property. In my view "zonemaster" is just the value we're using in the provided template config file.

Instead I'd say that the "default value" is the value a property takes on if it's omitted from the config file. The presence of a default value for a property indicates that the property is optional.

Technically Z::B::Config::DB_user returns `undef` if "user" is omitted. But this isn't really the default value for the database user, as that causes [DBI::connect](https://metacpan.org/pod/DBI#connect) to take its username value from the `DBI_USER` environment variable. The documentation doesn't specify what default value is used if `DBI_USER` is undefined.

I'm not sure how I think we ought to treat the "user" property with regard to optionality. One possibility is to make it required if "engine" is "MySQL" or "PostgreSQL" and forbidden if "engine" is "SQLite". But I guess that's a topic for another day.

docs/Configuration.md
matsduf: OK, so then we can possibly refer to the API document, or copy from there.

matsduf:
> And about the default values for the DB section, are they hard coded somewhere or only extracted from the `share/backend_config.ini` file ?

They are not hard coded in the code, as far as I can see. They are in the default `backend_config.ini` and they are used in the installation instructions.

docs/Configuration.md
matsduf: I agree that they are not default values in the code, but they are default values when using the default `backend_config.ini`. I think it would be fine to say that they are examples and refer to the default `backend_config.ini` and the installation instructions.

matsduf:
> It looks like there is a description of the queue attribute in the API documentation
>
> As I understand it, it is possible to use one database with different Test Agent, so this value works as a label to associate each test to its Test Agent.

Yes, but the scenario is not documented.

docs/Configuration.md
mattias-p: Can't we just assume that people interested in the factory settings would look them up in the template config file in the dist directory? Perhaps we could just spell out a command somewhere for how to find the file? I don't think we should go any further than that.

matsduf: It could be enough to refer to the factory settings in the default `backend_config.ini`.

pnax: With commit https://github.com/zonemaster/zonemaster-backend/pull/684/commits/efeed67bd2105ddf58fcd0433acaa5eb40289d25 I have added a mention at the beginning of the file about the factory settings, and removed the default values.

docs/Configuration.md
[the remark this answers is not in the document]
matsduf: I see no reason to keep any of it, but if you do not want to change the code in this PR, then "not used" could be fine. Same with all non-used entries.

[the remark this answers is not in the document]
pnax: Ok. I'll remove them from the documentation and from the Config.pm file then.

docs/Configuration.md
pnax: There is a comment about this property in a [Travis .ini file](https://github.com/zonemaster/zonemaster-backend/blob/159e80bc3510582f0d3f028a60a2d44773f19419/share/travis_postgresql_backend_config.ini#L24..L29)

So it looks like a condition to have a database with `id` made of integers and hashes. The value of this key is the boundary between each range.

docs/Configuration.md
```diff
@@ -146,10 +166,48 @@ to specifying a profile JSON file containing the entire
 
 ## ZONEMASTER section
 
-TBD
+The ZONEMASTER section has several keys :
+`max_zonemaster_execution_time`,
+`number_of_professes_for_frontend_testing`,
+`number_of_professes_for_batch_testing`,
```
matsduf, line 172: The two typos can be removed.

docs/Configuration.md
```diff
@@ -146,10 +166,48 @@ to specifying a profile JSON file containing the entire
...
+
+Time in seconds before reporting an unfinished test as failed. Default
+value: `600`.
+
+### maximal_number_of_retries
+
+Number of time a test is allowed to be run again if unfinished after
+`max_zonemaster_execution_time`. Default value: `0`.
+
+### number_of_professes_for_frontend_testing
```
matsduf, line 186: number_of_processes_for_frontend_testing

docs/Configuration.md
```diff
@@ -146,10 +166,48 @@ to specifying a profile JSON file containing the entire
...
+
+used -> todo
+
+### number_of_professes_for_batch_testing
```
matsduf, line 190: number_of_processes_for_batch_testing

docs/Configuration.md
pnax: done with https://github.com/zonemaster/zonemaster-backend/pull/684/commits/894d74be4a0f7abb2bf5323fa992723a50ec17f3

docs/Configuration.md
pnax: done with https://github.com/zonemaster/zonemaster-backend/pull/684/commits/894d74be4a0f7abb2bf5323fa992723a50ec17f3

docs/Configuration.md
pnax: done with https://github.com/zonemaster/zonemaster-backend/pull/684/commits/894d74be4a0f7abb2bf5323fa992723a50ec17f3

pnax:
For the keys `number_of_processes_for_frontend_testing` and `number_of_processes_for_batch_testing`, I understand that together, they define the total number of allowed processes. However, digging into the code, I cannot see how they are used on their own exactly.

It looks like theses parameters are linked to [the `priority` value of the test](https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#priority). I was then expecting to find portions of code to limit the number of used processes linked to the priority of the test. Am I missing something here ?

docs/Configuration.md
```diff
@@ -27,6 +29,33 @@ MySQL             | `MySQL`
...
+Ignored by the SQLite database engine.
+
+### database_name
+
+The name of the database to use, except for SQLite database engine where
+it holds the full path to the SQLite database file.
+
+### polling_interval
+
+Time in seconds between database lookups by Test Agent.
```
matsduf, line 57: Just a side comment, the "polling_interval" key does not really belong to the "DB" section (but that is how it is).

matsduf (approved):
Great. I think this is fine for now.

mattias-p (approved):

matsduf:
@pnax, please merge.

Issue #685, opened 2020-12-16
Validate backend config


mattias-p:
The backend config should be validated when it is loaded.
This should only be implemented after #651 is solved.
This is a breaking change (e.g. at present we silently ignore misspelled section names). If there is something else we want to change about the config file format it would probably be a good idea to make sure we can include those changes in the same release.

The documentation in Configuration.md should be updated so that:
* It is clear which properties are required, optional and possibly forbidden. If it depends on some context, those rules should also be clear.
* Every optional property should have a specified default value.
* It is clear what values are valid for each property.

The Config.md module should be updated so that:
* The section and property names should be checked so that only valid names are allowed.
* The set of properties in the config file should be checked so that all required properties are present and no forbidden properties are present.
* Each property value should be validated against its datatype.

When either daemon fails to start because of an invalid config file, an error message should be emitted containing the path of the offending file and the line number of the offending line, and some kind of hint as to what's wrong should also be provided. The error message should be presented to the user either directly in the console or in the log file, or both. Both daemons should have the same behavior in this respect.

matsduf:
I added a milestone to it, v2021.2. Is that reasonable?

pnax:
Would you say that this issue has been fully addressed @mattias-p ?

mattias-p:
Yeah, I think it's good.

PR #692, opened 2021-01-04
 Make age for reused test configurable 

matsduf:
When creating a new test through RPCAPI it will first check if there is a recent test with the same test parameters. If so, and that test is not too old, a new test will not be created, instead the test result for that recent test will be reused.

As it is now, "not too old" is hard-coded to 10 minutes. That is probably a good choice in most cases, but sometimes that is too short and sometimes too long. This PR makes that configurable.

As a side effect, this PR sorts the methods in the "[ZONEMASTER]" section from the ini file in `lib/Zonemaster/Backend/Config.pm` and also adds POD documentation to those methods.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -167,11 +167,13 @@ to specifying a profile JSON file containing the entire
 ## ZONEMASTER section
 
 The ZONEMASTER section has several keys :
-`max_zonemaster_execution_time`,
-`number_of_processes_for_frontend_testing`,
-`number_of_processes_for_batch_testing`,
-`force_hash_id_use_in_API_starting_from_id`, `lock_on_queue`,
-`maximal_number_of_retries`.
+* max_zonemaster_execution_time
+* number_of_processes_for_frontend_testing
+* number_of_processes_for_batch_testing
+* force_hash_id_use_in_API_starting_from_id
+* lock_on_queue
+* maximal_number_of_retries
+* age_reuse_previous_test
 
 ### max_zonemaster_execution_time
 
@@ -204,6 +206,13 @@ simple id, for everything above hash_id will be forced). Default value:
 ### lock_on_queue
 
 Integer working as a label to associate a test to a specific Test Agent.
+
+### age_reuse_previous_test
+
+Positiv integer (in seconds) for how old a previous test of the same
+zone name and parameters must be before we start a new test. Internally
+the value is converted to whole minutes. If the conversion results in
+zero minutes, then the default value (600 seconds) is used.
 
 --------
 
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -190,7 +190,8 @@ sub PollingInterval {
 
 =head3 INPUT
 
-'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file.
+'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file. See
+L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
 
 =head3 RETURNS
 
@@ -211,7 +212,8 @@ sub MaxZonemasterExecutionTime {
 
 =head3 INPUT
 
-'number_of_processes_for_frontend_testing' from [ZONEMASTER] section in ini file.
+'number_of_processes_for_frontend_testing' from [ZONEMASTER] section in ini file. See
+L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#number_of_processes_for_frontend_testing>.
 
 =head3 RETURNS
 
@@ -233,7 +235,8 @@ sub NumberOfProcessesForFrontendTesting {
 
 =head3 INPUT
 
-'number_of_processes_for_batch_testing' from [ZONEMASTER] section in ini file.
+'number_of_processes_for_batch_testing' from [ZONEMASTER] section in ini file. See
+L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#number_of_processes_for_batch_testing>.
 
 =head3 RETURNS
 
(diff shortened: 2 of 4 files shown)


docs/Configuration.md
```diff
@@ -205,6 +207,13 @@ simple id, for everything above hash_id will be forced). Default value:
 
 Integer working as a label to associate a test to a specific Test Agent.
 
+### age_reuse_previous_test
+
+Positiv integer (in seconds) for how old a previous test of the same
+zone name and parameters must be before we start a new test. Internally
+the value is converted to whole minutes. If the conversion results in
+zero minutes, then the default value (600 seconds) is used.
```
mattias-p, line 215: Why is the unit seconds here, rather than minutes?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -185,32 +185,84 @@ sub PollingInterval {
     return $self->{cfg}->val( 'DB', 'polling_interval' );
 }
 
+
+=head2 MaxZonemasterExecutionTime
+
+=head3 INPUT
+
+'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file.
```
mattias-p, line 193: Maybe link to the property in Configuration.md?

docs/Configuration.md
matsduf: I felt that seconds was a better unit, especially since other keys take seconds. I would rather turn it around and make it seconds all the way through. In the MySQL adapter it is converted to seconds before use.

If this is merged I will create a second PR where the time unit is seconds all the way through.

But I will not insist. If minutes is better I can take that.

lib/Zonemaster/Backend/Config.pm
matsduf: Good suggestion, but those links will be broken until the updated Configuration.pm has been merged to master branch. I guess that is OK.

docs/Configuration.md
mattias-p: I see. Uniform units sure is a nice feature. Seconds it is. If we could also get rid of the rounding, that would be great!

matsduf:
@mattias-p and @pnax, can you review?

share/backend_config.ini
```diff
@@ -19,7 +19,7 @@ interpreter       = perl
 max_zonemaster_execution_time            = 300
 number_of_processes_for_frontend_testing = 20
 number_of_processes_for_batch_testing    = 20
-
+#age_reuse_previous_test=600
 #
```
pnax, line 23: I think this line should be blank to make a clear separation with the following WARNING clause that is related to `maximal_number_of_retries`.

pnax (commented):
See my comments.
I also ran some tests, and this seems to work.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -282,6 +321,64 @@ sub maximal_number_of_retries {
...
+10). Note that it returns in minutes, whereas the input is in seconds.
+
+=cut
+
+sub age_reuse_previous_test {
+    my ($self) = @_;
+    my $val = $self->{cfg}->val( 'ZONEMASTER', 'age_reuse_previous_test' );
+    $val = ($val > 0) ? $val : 0;
+    $val = int ( ($val / 60) + 0.5); # in minutes
+    return ($val)? $val : 10;
```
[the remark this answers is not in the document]
matsduf: Fixed. Some other PR can make it consistent in the module.

share/backend_config.ini
matsduf: Fixed and made the warning clearer. For how long should `maximal_number_of_retries` be considered to be experimental?

matsduf:
@pnax, please re-review.

share/backend_config.ini
pnax: No idea... I found PR #544 where this parameter has been added. I understand that this option has not been fully tested (as the warning says) and that there might be issues with parallelism.

pnax (approved):

lib/Zonemaster/Backend/Config.pm
```diff
@@ -185,32 +185,88 @@ sub PollingInterval {
...
+=head2 MaxZonemasterExecutionTime
+
+=head3 INPUT
+
+'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file. See
+L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
+
+=head3 RETURNS
+
+Integer (number of seconds)
```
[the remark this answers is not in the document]
matsduf: It is out of scope for this PR to add validation for anything execept `age_reuse_previous_test`, but that is a good suggestion. For `age_reuse_previous_test` the validation is done such as if the value is not numeric and positive, it will be ignored. Should it croak instead?

I do not understand how to used `load_config` method for validation.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -185,32 +185,88 @@ sub PollingInterval {
...
+=head2 NumberOfProcessesForFrontendTesting
+
+=head3 INPUT
+
+'number_of_processes_for_frontend_testing' from [ZONEMASTER] section in ini file. See
+L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#number_of_processes_for_frontend_testing>.
+
+=head3 RETURNS
+
+Positive integer (default 600).
```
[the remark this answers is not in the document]
matsduf: Good catch. I will fix.

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
mattias-p: On second thought it's probably better to leave the validation out for now and deal with it in #685.

mattias-p (approved):

mattias-p:

I verified this using the commands below on CentOS 7 and CentOS 8.

```sh
sudo sed -i '/\bage_reuse_previous_test\b/ s/=.*/= 20/' /etc/zonemaster/backend_config.ini
sudo systemctl restart zm-testagent
zmtest example > /dev/null
sleep 10
zmtest example > /dev/null
sleep 20
zmtest example > /dev/null
# Verify that the two first zmtest commands report the same testid, but the third one reports a different one
```

Issue #687, opened 2020-12-17
Install zmb and zmtest in PATH


matsduf:
`zmb` and `zmtest` are useful tools for backend testing. Now it is installed in "share dist" directory, which makes it hard to use them. Instead they should be installed in the `/usr/bin` (Linux) or `/usr/local/bin` (FreeBSD).

matsduf:
Resolved by #693.

PR #693, opened 2021-01-07
Proper installation of zmtest and zmb

mattias-p:
Solves #687.

Since the tools are being properly installed the dependencies that needs to be reflected in the dependencies. AFAICT we don't depend on curl anymore, so I'm removing that from the installation instruction. The GettingStarted document still uses curl but it has a separate note about installing it so we shouldn't need to consider that in the installation instruction.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -22,6 +22,7 @@ requires
   'Log::Any'                    => 0,
   'Log::Any::Adapter::Dispatch' => 0,
   'Log::Dispatch'               => 0,
+  'LWP::UserAgent'              => 0,
   'Moose'                       => 2.04,
   'Parallel::ForkManager'       => 1.12,
   'Plack::Builder'              => 0,
@@ -44,6 +45,8 @@ install_share;
 
 install_script 'zonemaster_backend_rpcapi.psgi';
 install_script 'zonemaster_backend_testagent';
+install_script 'zmtest';
+install_script 'zmb';
 
 no_index directory => 'CodeSnippets';
 no_index directory => 'Doc';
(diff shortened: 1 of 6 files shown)


Makefile.PL
```diff
@@ -22,6 +22,7 @@ requires
   'Log::Any'                    => 0,
   'Log::Any::Adapter::Dispatch' => 0,
   'Log::Dispatch'               => 0,
+  'LWP::UserAgent'              => 0,
```
matsduf, line 25: Do you have a comment for this change? I do not see that it is related to the rest.

matsduf (approved):

Makefile.PL
mattias-p: `zmb` nees it.

docs/Installation.md
```diff
@@ -691,12 +673,10 @@ Use the procedure for installation on [Debian](#4-installation-on-debian).
...
 
 If you have followed the installation instructions for Zonemaster::Backend above,
-you should be able to use the
-API on localhost port 5000 as below. The command requires that `curl` and `jq` are installed.
-
+you should be able to use the API on localhost port 5000 as below.
 
 ```sh
-`perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`/zmtest zonemaster.net
+`zmtest zonemaster.net
```
pnax, line 679: I think you left a leading back quote.

mattias-p: Good catch! Fixing.

matsduf (approved):

PR #696, opened 2021-01-21
Update CentOS installation instruction

mattias-p:
These are corrections and simplifications I made during installation testing for MariaDB and PostgreSQL. See commit messages for


matsduf (approved):
I see no issue with it.

PR #697, opened 2021-01-22
Fix SQLite DB file ownership

mattias-p:
With this change I can successfully run the smoke test with SQLite.


matsduf (approved):
If this is needed for CentOS I guess it applies for Debian/Ubuntu and FreeBSD too.

PR #698, opened 2021-01-25
Make Backend start automatically when booting

mattias-p:
Make the services come up after reboot on CentOS.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -230,7 +230,7 @@ Make sure our tmpfiles configuration takes effect:
 sudo systemd-tmpfiles --create /usr/lib/tmpfiles.d/zonemaster.conf
 ```
 
-Start the services:
+Enable services at boot time and start them:
 
 ```sh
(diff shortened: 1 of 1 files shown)


matsduf (dismissed):

docs/Installation.md
```diff
@@ -233,15 +233,10 @@ sudo systemd-tmpfiles --create /usr/lib/tmpfiles.d/zonemaster.conf
 Start the services:
```
pnax, line 233: I would add the fact that the following commands will also enable the services with something like : «Enable the services at boot time and start them:»

pnax (commented):
This is a good idea. I think these commands can also be added to the Debian section.

matsduf (approved):

mattias-p:
> I think these commands can also be added to the Debian section.

I believe so too. But I'm only testing CentOS so I don't want to make updates elsewhere. Feel free to copy them.

PR #700, opened 2021-01-27
Update Debian installation instructions

pnax:
This PR:
* aligns locale generation with CLI installation document
* apply relevant changes from Centos section to Debian (inspired from https://github.com/zonemaster/zonemaster-backend/pull/696)
* install databases before configuring Zonemaster
* fix incorrect option
* fix database ownership

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -255,12 +255,17 @@ See the [post-installation] section for post-installation matters.
 Install required locales:
 
 ```sh
-locale -a | grep da_DK.utf8 || echo da_DK.UTF-8 UTF-8 | sudo tee -a /etc/locale.gen
-locale -a | grep en_US.utf8 || echo en_US.UTF-8 UTF-8 | sudo tee -a /etc/locale.gen
-locale -a | grep fr_FR.utf8 || echo fr_FR.UTF-8 UTF-8 | sudo tee -a /etc/locale.gen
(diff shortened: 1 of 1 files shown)


matsduf (dismissed):
It looks good.

docs/Installation.md
```diff
@@ -376,9 +384,6 @@ sudo -u postgres psql -f $(perl -MFile::ShareDir -le 'print File::ShareDir::dist
 
 #### 4.2.3 Instructions for SQLite (Debian)
 
-> **Note:** Zonemaster with SQLite backend is not yet considered stable and anyway
-> not meant for an installation with heavy load.
```
matsduf, line 380: I guess we should have the same statement -- or no statement -- for all OS. Should the statement that "Zonemaster with SQLite backend is not yet considered stable and anyway not meant for an installation 
with heavy load" be put somewhere else or not at all?

pnax: Good point. I removed it because this release brings a stable SQLite support. But indeed the second part of the note is still relevant.
I'll revert this specific commit and make a new PR to update this note in all sections.

pnax:
Rebased on top of `develop` and removed commit https://github.com/zonemaster/zonemaster-backend/pull/700/commits/cbb6ee624c4f5b2907f005ac6538116fcdf6e2e8

matsduf (approved):

PR #705, opened 2021-02-03
Accept empty IPs in nameserver objects

### LINKED ISSUE #702: RPCAPI does not accept empty IP
This issue is based on the findings in issue zonemaster/zonemaster#925.

In method `start_domain_test` a name server object optionally be specified. If specified, and non-empty, it contains a list of name servers. For each name server a name, `ns`, must be specified. One IP address, `ip` can optionally be specified. Backend correctly accepts `ns` without `ip` but it incorrectly rejects empty `ip`. Those two cases should be considered to be equal. E.g.

```
    "nameservers": [
      {
        "ip": "",
        "ns": "ns3.nic.se"
      }
    ],
```
and

```
    "nameservers": [
      {
        "ns": "ns3.nic.se"
      }
    ],
```

Example from the API document.

```
    "nameservers": [
      {
        "ip": "2001:67c:124c:2007::45",
        "ns": "ns3.nic.se"
      },
      {
        "ip": "192.93.0.4",
        "ns": "ns2.nic.fr"
      }
    ],
```

[API]: https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#api-method-start_domain_test

matsduf:
Resolves #702

### DIFF
--- a/lib/Zonemaster/Backend/Validator.pm
+++ b/lib/Zonemaster/Backend/Validator.pm
@@ -48,7 +48,7 @@ sub ds_info {
     );
 }
 sub ip_address {
-    return joi->string->regex($ipv4_regex."|".$ipv6_regex);
+    return joi->string->regex('^$|'.$ipv4_regex."|".$ipv6_regex);
 }
 sub nameserver {
     return joi->object->strict->props(


matsduf:
@pnax and @mattias-p, please review!

pnax (approved):
I have tested it with the GUI directly and this looks good. No need to do anything on the GUI, this solves the issue with empty "ip" fields sent by the GUI.

edit: just a note, you used single quotes but the rest of the line uses double quotes, any reason for that choice ?

mattias-p (approved):

mattias-p:
Technically this change is in violation with the API documentation, but I accept it as a quick fix.

We should also update GUI to omit the "ip" parameter if the corresponding field is empty. After GUI (and any other clients) have been fixed, this PR should be reverted to bring Backend back into compliance with the documentation.

matsduf:
> I have tested it with the GUI directly and this looks good. No need to do anything on the GUI, this solves the issue with empty "ip" fields sent by the GUI.

Great!

> edit: just a note, you used single quotes but the rest of the line uses double quotes, any reason for that choice ?

I use single quote on the string I added because no interpolation is done in it. I did not change the other part, but I was tempted to. I use single when possible and it does not add extra backslash.

PR #713, opened 2021-02-27
Fixes correct start order at start-up

matsduf:
When backend uses SQLite the rc system places zm_testagent to be started before the network, which does not work well, and testagent does not get started. This PR requires the network to be up before zm_testagent is run.


mattias-p (approved):

pnax (approved):

matsduf:

Works as expected.

PR #714, opened 2021-03-05
Clean up obsolete command

mattias-p:
This directory creation is handled by tmpfiles.

Kudos to @matsstralbergiis for spotting this.


matsduf (approved):

pnax (commented):
I don't know much about tmpfiles.d. From my understanding, this service is run on boot and at periodic interval. In this case, how to be sure that the directory will be there at the end of the installation process ?

mattias-p:
@pnax I share your understanding of systemd-tmpfiles. We take care of the immediate need by running `sudo systemd-tmpfiles --create /usr/lib/tmpfiles.d/zonemaster.conf` further down in the "Service configuration and startup" section.

pnax:
Oh the command is indeed already there. Thank you for pointing this.

pnax (approved):

Issue #691, opened 2021-01-03
Any use case for "force_hash_id_use_in_API_starting_from_id"?


matsduf:
The records in the databases have two unique identifiers, a normal integer that is incremented by one each time a test is created and a hash that is just unique. The hash is what is always used unless `force_hash_id_use_in_API_starting_from_id` is set in the configuration file to some value higher than zero and then straight integer IDs can be used for those lower than that value.

Is there really a valid use case for using the integer IDs? If not, the code could be simplified.

Proposal:
* Remove all code dependent on `force_hash_id_use_in_API_starting_from_id` and force the use of the hash IDs in all cases.
* Remove `force_hash_id_use_in_API_starting_from_id` from `Config.pm`, the ini file and all documentation.

mtoma:
I'd say this is reasonable proposal as the only use of this parameter is for very old installations that started by using the id without hash_id and wanted/needed to enforce the hash_id starting from some specific id. As far as I can tell the zonemaster.net installation was the only one in this case. Not sure if there are some IDs without hash_ids or if the database was wiped at some point making this parameter useless even there.

mattias-p:
So we should look for test_results in the zonemaster.net database that don't have hash_ids, and if there are any, generate new random hash_ids for them. Since `force_hash_id_use_in_API_starting_from_id` is undocumented in the latest release (2020.1) it should be safe to assume that nobody outside of the Zonemaster team has been using it.

Right?

matsduf:
Resolved by #726.

PR #715, opened 2021-03-09
SQLite: harmonize get_test_history with MySQL and PostgreSQL

pnax:
Currently SQLite `get_test_history` does not behave exactly as the MySQL and PostgreSQL equivalent functions.

This PR fixes that :

 * the return `id` is now either the `hash_id` or the `id` based on `force_hash_id_use_in_API_starting_from_id` parameter in the config file
 * `overall_result` is now present in the output

There is an issue #691 currently discussing on a possible removal of `force_hash_id_use_in_API_starting_from_id`. This PR adds it in SQLite to offer the same behavior as with other databases (and because of another PR to add SQLite to `t/test_DB_backend.pl` - #716).

### DIFF
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -311,7 +311,8 @@ sub get_test_history {
                     id,
                     hash_id,
                     creation_time,
-                    params
+                    params,
+                    results
                  FROM
                     test_results
                  WHERE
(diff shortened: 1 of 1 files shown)


matsduf (approved):

pnax:
About the `force_hash_id_use_in_API_starting_from_id`, just linking PR #694 that deprecates it.

mattias-p (approved):

pnax:

I've verified this test procedure on Debian 10, Ubuntu 18.04 and Ubuntu 20.04. There is an `overall_result` entry with SQLite, MySQL and PostgreSQL.

PR #716, opened 2021-03-09
Support SQLite in t/test_DB_backend.pl

pnax:
Allow SQLite as database in the `test_DB_backend.pl` script. This requires some small fixes in the code.
In order to make all test pass, this PR requires #715 (and the use of parameter `force_hash_id_use_in_API_starting_from_id`).

_Remark:_ this PR contains a commit (51de4b1) that reformats the code (replace tabs with spaces and remove trailing spaces).

### DIFF
--- a/t/test_DB_backend.pl
+++ b/t/test_DB_backend.pl
@@ -56,7 +56,7 @@ sub run_zonemaster_test_with_backend_API {
     foreach my $i ( 1 .. 12 ) {
         sleep( 5 );
         my $progress = $engine->test_progress( $api_test_id );
-        diag "pregress: $progress";
+        diag "progress: $progress";
         last if ( $progress == 100 );
     }
     ok( $engine->test_progress( $api_test_id ) == 100 , 'API test_progress -> Test finished' );
--- a/t/test_DB_backend.pl
+++ b/t/test_DB_backend.pl
@@ -7,7 +7,7 @@ use JSON::PP;
 use Test::Exception;
 
 my $db_backend = $ARGV[0] // BAIL_OUT( "No database backend specified" );
-( $db_backend eq 'PostgreSQL' || $db_backend eq 'MySQL' ) or BAIL_OUT( "Unsupported database backend: $db_backend" );
+( $db_backend eq 'SQLite' || $db_backend eq 'PostgreSQL' || $db_backend eq 'MySQL' ) or BAIL_OUT( "Unsupported database backend: $db_backend" );
 
 my $frontend_params_1 = {
     client_id      => "$db_backend Unit Test",         # free string
@@ -83,7 +83,7 @@ my $user_check_query;
 if ($db_backend eq 'PostgreSQL') {
     $user_check_query = q/SELECT * FROM users WHERE user_info->>'username' = 'zonemaster_test'/;
 }
-elsif ($db_backend eq 'MySQL') {
+else {
     $user_check_query = q/SELECT * FROM users WHERE username = 'zonemaster_test'/;
 }
 
--- a/t/test_DB_backend.pl
+++ b/t/test_DB_backend.pl
@@ -10,17 +10,17 @@ my $db_backend = $ARGV[0] // BAIL_OUT( "No database backend specified" );
 ( $db_backend eq 'PostgreSQL' || $db_backend eq 'MySQL' ) or BAIL_OUT( "Unsupported database backend: $db_backend" );
 
 my $frontend_params_1 = {
-	client_id      => "$db_backend Unit Test",         # free string
-	client_version => '1.0',               # free version like string
-	domain         => 'afnic.fr',          # content of the domain text field
-	ipv4           => JSON::PP::true,                   # 0 or 1, is the ipv4 checkbox checked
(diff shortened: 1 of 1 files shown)


matsduf:
This looks fine, but SQLite database is already tested in t/test01.t. Will this PR create double testing? Should the testing t/test01.t be removed?

pnax:
Oh to be honest, I haven't opened `t/test01.t` ^^'. You're right, this might create double testing indeed.

To share how I came to propose this PR, I was looking on how to run the tests and I looked at how Travis was configured. And I saw that MySQL and PostreSQL had the same way to invoke tests, but no trace of SQLite. So I wanted to add support for SQLite. But this was wrong of me, SQLite is already tested with `make test`.

Now I understand that there is a duplication with 2 different ways of calling the test (correct me if I am missing something) :

* one way that lies on `make test` and that tests the SQLite database
* another one that should be called via a command such as `perl t/test_DB_backend.pl <DB>` where we pass the database to use (MySQL or PostgreSQL)

I think it would be easier to have one unique way to call the tests. Moreover the tests duplication might make it harder to maintain and add tests that test all databases. Especially, there are tests in `t/test_DB_backend.pl` that are not present in `t/test01.t` (this is why I opened #715).

So I think it would be nice to avoid duplication of tests between databases. If you agree I can merge the files into one that can be used for all databases.
That said, how do we differentiate between databases ? I don't know much about testing in Perl but it looks like :
* using `make test`, we cannot pass an argument, so we need to have one `backend_config.ini` file per database with the right `engine` key
* using a perl script, we can pass the database as an argument as it is currently done in `t/test_DB_backend.pl`, but we can't rely on `make test` anymore

Any comments, thoughts ?

mattias-p:
I completely agree with you that having all databases run through the same tests would be an improvement.

I think it would be neat if the test file could autodetect which databases are installed and [SKIP](https://perldoc.perl.org/Test::More#Conditional-tests) the ones that are unavailable.

mattias-p:
Also, it's possible to use environment variables to pass arguments through `make test`.

And I think it would be neat if the Config module could be configured programmatically. That way we wouldn't have to deal with different backend_config.ini files for the unit tests.

mattias-p:
And.. `make test` is run when you install using `cpanm` (unless you specify `--notest`).

matsduf:
We have two use cases for tests, one are tests that are run by "make test", and they can only use SQLite. The other are tests that are tests that are run by travis, and then all databases can be used.

I think it would be an improvement if it is specified what the purpose are of the different test scripts so it is easy to see that this is a script for Travis only, and this is a script for "make test".

pnax:
@mattias-p: thank you for your answers and pointers.

@matsduf : looking at the [Travis config (v6.1.0)](https://github.com/zonemaster/zonemaster-backend/blob/v6.1.0/.travis.yml#L122..L124), it looks like Travis also runs the `make test` command (and therefore tests the SQLite database).

So merging your answers, I would like to propose something like this :
* keep only the `make test` command to run the tests
* by default use the SQLite database (assert SQLite is installed)
* use environment variables to change/force the database used for the tests
* use the SKIP stanza to avoid running tests specific to one database (because it is not installed or because the user does not want to)

I think this handles all the cases (Travis and cpanm). And from my point of view, this would make easier to implement new tests, maintain tests and run the tests locally.

What do you think ?

matsduf:
The idea is good, but I am not sure that running tests on MariaDB/MySQL or PostgreSQL by default if they are installed is unproblematic. I suggest that that "make test"
* by default runs tests on SQLite, as today, but make it possible to deselect it by an environment variable,
* by default not runs tests on MariaDB/Mysql or PostgreSQL, but make it possible to select it by an environment variable.

Travis could then these tests three times, one time each for each database.

pnax:
> but I am not sure that running tests on MariaDB/MySQL or PostgreSQL by default if they are installed is unproblematic

I have no idea how easy or hard this can be to implement. But I think we should let the user choose what database the tests should be run with and not create tables everywhere in the system. Except strong opinion on this, I'll go with SQLite by default and a way to specify another database.

It looks like we agree on the idea to merge the files. I'll start working on this. Should I close this PR and start a new one ?

mattias-p:
Yeah, it sounds like a good model to make SQLite the default and use an environment variable to allow overriding to another use different database.

The easiest way to do this at the moment is probably to let the environment variable override the backend_config.ini - this allows us to provide different database credentials for the different databases. In fact there is already such a variable, though I don't like how it's implemented.

A better model for the database configuration would be to have a separate section in the config file for each database type, a property in the configuration file to select database type, as well as a programmatic interface to for overriding at least the database type property. And finally have the unit test take the database type from an environment variable and override the config property. But this is a fairly big change in itself.

mattias-p:
On the whether to stay with this PR or open another one, I'd say go with whichever you feel more comfortable with. If you switch to a new PR it's probably a good idea to link to this one to keep track of the discussion.

pnax:
> Is this fully replaced by #731? If so, can this be closed?

Yes, I forgot to close it when I opened #731 since this is the following from this discussion.

Issue #674, opened 2020-11-16
Inconsistent behavior of profiles


matsduf:
If the "PUBLIC PROFILE" section of `backend_config.ini` is set up to

```
[PUBLIC PROFILES]
default=/etc/zonemaster/profile.json
nisse=/etc/zonemaster/nisse-profile.json
```
then RPCAPI sees the change immediately, i.e. by method `profile_names`, but the Test Agent requires a restart to see the changes.

matsduf:
Resolved by #717.

Issue #651, opened 2020-09-30
Reload ini file at startup only


matsduf:
Today the ini file is reloaded for every test. Change that behavior so that the ini file is only reloaded when the daemons (RPCAPI and Test Agent, respectively) are started. They should also verify the ini file and refuse to run if any error is found.

matsduf:
Resolved by #717.

PR #717, opened 2021-03-10
Load ini file at RPCAPI and TestAgent startup only

pnax:
This PR is about loading the `backend_config.ini` file only once, at daemon startup for RPCAPI and TestAgent.

* it makes mandatory to pass a `config ref` as parameter to each  RPCAPI and TestAgent constructor (die otherwise)
* use this `config ref` everywhere instead of loading a new config
* remove the possibility to pass and load a module from constructor parameter, add a layer to check the validity of the input and then import the module (or die)
* do some refactoring and code formatting

Addresses #651, #674

_Remark:_ one commit (fd7a7f3 ) is only about reformatting (replacing tabs by 4 spaces and removing trailing spaces) 

## How to test the PR

Before, the RPCAPI was loading the ini file on every API calls. This can be highlighted by running the following steps :

1. create a valid `backend_config.ini` file
2. start the daemons `rpcapi` and `testagent`
3. use `zmb` to load the available profiles
   ```
   $ zmb profile_names
   {"result":["default"],"jsonrpc":"2.0","id":1}
   ```
4. add a new profile to the `backend_config.ini` file :
   ```
   ...
   [PUBLIC PROFILES]
   dummy=/tmp/dummy-profile.json
   ...
   ```
5. run `zmb` again, and see this profile being listed
   ```
   $ zmb profile_names
   {"result":["default","dummy"],"jsonrpc":"2.0","id":1}
   ```

Now with this PR, steps 1 to 4 are the same, however the output from step 5 should be the same as the output in 3.
1. create a valid `backend_config.ini` file
2. **restart** the daemons `rpcapi` and `testagent`
3. use `zmb` to load the available profiles
4. add a new profile to the `backend_config.ini` file
5. run `zmb` again, and see the same output as in 3.
   ```
   $ zmb profile_names
   {"result":["default"],"jsonrpc":"2.0","id":1}
   ```
This can be tested with other parameters than a new profile entry (for instance removing a `locale`). Any modification of the `backend_config.ini` file should remain invisible to the TestAgent and the RPCAPI. Only restarting the daemons will apply the changes.

---
Edit :

* it is possible to pass the config reference to the constructor, see https://github.com/zonemaster/zonemaster-backend/pull/717#discussion_r594380960 and 06ef769 
* rebased and force pushed with [commit 6a12dea](#commits-pushed-6a12dea)
* removed the _work in progress_ statement
* updated PR description
* update commit ref after rebase
* added examples on how to test this PR

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -57,18 +57,24 @@ sub new {
         }
     }
     else {
-        eval {
-            my $backend_module = "Zonemaster::Backend::DB::" . $self->{config}->BackendDBType();
-            eval "require $backend_module";
-            die "$@ \n" if $@;
-            $self->{db} = $backend_module->new( { config => $self->{config} } );
-        };
-        if ($@) {
-            handle_exception('new', "Failed to initialize the [$params->{db}] database backend module: [$@]", '002');
-        }
+        $self->_init_db($self->{config}->BackendDBType());
     }
 
     return ( $self );
+}
+
+sub _init_db {
+    my ( $self, $dbtype ) = @_;
+
+    eval {
+        my $backend_module = "Zonemaster::Backend::DB::" . $dbtype;
+        eval "require $backend_module";
+        die "$@ \n" if $@;
+        $self->{db} = $backend_module->new( { config => $self->{config} } );
+    };
+    if ($@) {
+        handle_exception('_init_db', "Failed to initialize the [$dbtype] database backend module: [$@]", '002');
+    }
 }
 
 sub handle_exception {
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -46,7 +46,7 @@ sub new {
 
     $self->{config} = $params->{config};
 
-    if ( $params && $params->{db} ) {
+    if ( $params->{db} ) {
         eval {
             eval "require $params->{db}";
             die "$@ \n" if $@;
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -40,13 +40,13 @@ sub new {
     my $self = {};
     bless( $self, $type );
 
-    my $config = Zonemaster::Backend::Config->load_config();
+    $self->{config} = Zonemaster::Backend::Config->load_config();
 
     if ( $params && $params->{db} ) {
         eval {
             eval "require $params->{db}";
             die "$@ \n" if $@;
-            $self->{db} = "$params->{db}"->new( { config => $config } );
+            $self->{db} = "$params->{db}"->new( { config => $self->{config} } );
         };
         if ($@) {
             handle_exception('new', "Failed to initialize the [$params->{db}] database backend module: [$@]", '001');
@@ -54,10 +54,10 @@ sub new {
     }
     else {
         eval {
-            my $backend_module = "Zonemaster::Backend::DB::" . $config->BackendDBType();
+            my $backend_module = "Zonemaster::Backend::DB::" . $self->{config}->BackendDBType();
             eval "require $backend_module";
             die "$@ \n" if $@;
-            $self->{db} = $backend_module->new( { config => $config } );
+            $self->{db} = $backend_module->new( { config => $self->{config} } );
         };
         if ($@) {
             handle_exception('new', "Failed to initialize the [$params->{db}] database backend module: [$@]", '002');
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -46,14 +46,15 @@ sub new {
 
     $self->{config} = $params->{config};
 
-    if ( $params->{db} ) {
+    if ( $params->{dbtype} ) {
         eval {
-            eval "require $params->{db}";
+            my $backend_module = "Zonemaster::Backend::DB::" . $params->{dbtype};
+            eval "require $backend_module";
             die "$@ \n" if $@;
-            $self->{db} = "$params->{db}"->new( { config => $self->{config} } );
+            $self->{db} = $backend_module->new( { config => $self->{config} } );
         };
         if ($@) {
-            handle_exception('new', "Failed to initialize the [$params->{db}] database backend module: [$@]", '001');
+            handle_exception('new', "Failed to initialize the [$params->{dbtype}] database backend module: [$@]", '001');
         }
     }
     else {
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -89,7 +89,7 @@ sub version_info {
     if ($@) {
         handle_exception('version_info', $@, '003');
     }
-    
+
     return \%ver;
 }
 
@@ -105,7 +105,7 @@ sub profile_names {
     if ($@) {
         handle_exception('profile_names', $@, '004');
     }
-    
+
     return \@profiles;
 }
 
@@ -136,7 +136,7 @@ sub get_host_by_name {
     if ($@) {
         handle_exception('get_host_by_name', $@, '005');
     }
-    
+
     return \@adresses;
 }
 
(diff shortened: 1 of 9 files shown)


pnax:
@matsduf, I think this is ready for testing. I said this is a work in progress because I haven't tested all API calls with all DBs.

Yes good point on documenting that both daemons should be restarted. I'll have a look to add documentation to the PR.

pnax:
Some updates.

* RPCAPI only takes the database type and then load the right module accordingly.
* I fixed a missing overwritting of the database value in the `_init_db` call.
* I finally removed all call to `Zonemaster::Backend::Config->load_config` in the RPCAPI. The configuration should be passed as parameter while instantiating the RPCAPI or in each dispatch call.

I realize the TestAgent has a similar behavior regarding database module and config. I'll clean this a little bit in upcoming commits. I think this will only be some code refactoring.

The documentation is still missing.

Feel free to comment.

pnax:
Oopsy, I think commit c794bdc  broke Travis. It removes instantiation of `$self->{db}` that is required in `t/test01.t`.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -40,33 +40,53 @@ sub new {
...
+    }
+    else {
+        handle_exception('_check_db', "Unknown dbtype '$dbtype', should be one of SQLite, MySQL or PostgreSQL", '001');
+    }
+}
+
+sub _init_db {
+    my $self = shift;
+    my $config = $self->{config};
+    $self->{dbtype} //= _check_db($config->BackendDBType());
```
mattias-p, line 77: Eventually I'd like to put the responsibility to perform this validation on the Config module. It's okay to have it here for now though.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -89,33 +109,35 @@ sub version_info {
...
-    
+
     return \%ver;
 }
 
 $json_schemas{profile_names} = joi->object->strict;
 sub profile_names {
-    my ($self) = @_;
+    my ( $self, undef, undef, $config ) = @_;
+    $self->{config} //= $config;
```
mattias-p, line 119: What's the point of having a config member if we're getting the config through arguments anyway?

lib/Zonemaster/Backend/RPCAPI.pm
pnax: I agree, this is not the place of such a test. I might move it the the Config module if I implement the same process in TestAgent.pm.

lib/Zonemaster/Backend/RPCAPI.pm
pnax: The config member is required to use the `_init_db` method. You are right this method is not called here, this is a leftover from my numerous copy/paste.
However it is required in other methods calling `_init_db`. This is because there are places where we specify the database to use even if there is one in the .ini (in the t/*.t files for instance).

I think it would be even better to pass the $config to the constructor and not to the dispatch calls.

I'll remove this line here and in `get_language_tags` and `get_data_from_parent_zone` as there is no use for it.

mattias-p: I see, that makes sense. We should be able to provide the Z::B::RPCAPI constructor with arguments like so: https://metacpan.org/pod/JSON::RPC#BASIC-USAGE.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -78,7 +97,7 @@ sub handle_exception {
 
 $json_schemas{version_info} = joi->object->strict;
 sub version_info {
-    my ( $self ) = @_;
+    my $self = shift;
```
mattias-p, line 100: Could you change these back to `my ( $self ) = @_;` for consistency?

mattias-p (commented):
I've got one last nit from looking at the changes. I haven't actually tested it though.

What's your take on the state of this PR? Do you feel it's ready fro merge?

lib/Zonemaster/Backend/RPCAPI.pm
pnax: done in 52852d5

pnax:
With these new commits, I :

* moved the check of the database to the Config module
* applied the changes to TestAgent.pm (I put all the commits on a specific branch, you can use `git log --oneline --graph` to see it)
* made some refactoring

I realize, there are a lot of commits. Some are more or less reverting changes added at the beginning of the work on this PR. Should I rebase and try to clean the git history ?

I haven't run edge cases yet and have not tried all databases. So I don't feel comfortable with merging yet. But you can test on your side, I think this is a quite stable PR now.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -40,33 +40,51 @@ sub new {
...
+
+    $self->{config} = $params->{config};
+
+    my $dbtype;
+    if ( $params->{db} ) {
+        $dbtype = $params->{db};
+    } else {
+        $dbtype = $self->{config}->BackendDBType();
     }
+    $dbtype = $self->_check_db($dbtype);
```
mattias-p, line 55: Couldn't the call to _check_db() be replaced with something like `$dbtype = $self->{config}->check_db( $params->{db} );` a few lines up? The call to handle_exception() made sense back when we reloaded the config file all the time, but this PR fixes that. After this PR the only time _check_db() would call handle_exception() is in unit tests, and I don't think handle_exception is very helpful in that context.

Also don't you think `dbtype` is a better name than `db`? Since we don't have proper type annotations in Perl I try to convey nuances like that in the names themselves.

mattias-p (commented):
I found another last nit.. Sorry 'bout that. But it's a good one!

lib/Zonemaster/Backend/RPCAPI.pm
pnax: I see what you mean for the `_check_db` call. I'll remove it.

> Also don't you think dbtype is a better name than db?

Do you mean in the hash passed to the constructor as in `params->{dbtype}` ?

mattias-p:
> I moved the check of the database to the Config module

Very nice!

> I realize, there are a lot of commits. Some are more or less reverting changes added at the beginning of the work on this PR. Should I rebase and try to clean the git history ?

If it's not too much work that would make git blame much more helpful. I believe our future selves might appreciate it.

> I haven't run edge cases yet and have not tried all databases. So I don't feel comfortable with merging yet. But you can test on your side, I think this is a quite stable PR now.

Sounds good! I'll see if I can fit in some testing too.

pnax:
> I believe our future selves might appreciate it.

I totally agree with you on this !

matsduf:
I agree that cleaning up the commits would both help reviewing and future inspection.

pnax:
I have cleaned the commits and force pushed them. You can review them.

I haven't done more testing yet.

matsduf:
@pnax, I will review and test. Could you review the PR description and update if needed to make it easier to review and understand what to expect from testing?

pnax:
@matsduf I updated the PR description

matsduf:
> @matsduf I updated the PR description

Thanks! I will review and test. Great that you have done this.

mattias-p:
@pnax Could you resolve the conflicts of this PR? It would be great if we could have this tested and merged.

pnax:
Oh I haven't seen that there were conflicts. Thanks for pointing them out. I'm on it.

pnax:
I rebased it and fixed the conflicts. Just waiting for Travis and this will be good for review @matsduf @mattias-p

pnax:
I updated the PR description with an example on how to test this PR.

mattias-p (approved):
I was able to reproduce the reloading of the RPCAPI in the develop branch, and I was able to verify that the config loading happens only at startup.

For the TestAgent I observed loading of the config only at startup for both the develop branch and this PR.

mattias-p:

Using the test procedure describe above I verified that zm-rpcapi only loads its config at startup on CentOS 7 and CentOS 8.

To verify that zm-testagent also only loads its config at startup I did the following:

```
sudo sed -i '/\bengine\b/ s/=.*/= SQLite/' /etc/zonemaster/backend_config.ini
sudo systemctl restart zm-testagent
zmtest example > /dev/null
sudo sed -i '/\bengine\b/ s/=.*/= PostgreSQL/' /etc/zonemaster/backend_config.ini
zmtest example > /dev/null
sudo systemctl restart zm-testagent
zmtest example > /dev/null
# Verify that the two first zmtest commands report the same testid, but the third one reports a different one
```

PR #718, opened 2021-03-15
Delete unused test_runner.pl

pnax:
The `test_runner.pl` looks like a leftover from an ancient time. It looks like it is not used anywhere anymore, so I suggest to delete it from the code base. If I missed something about its usage, please correct me.


matsduf (approved):
It cannot be meaningful as is. Yes, remove.

mattias-p (approved):

PR #723, opened 2021-03-19
Properly initialize SQLite handle

### LINKED ISSUE #722: SQLite adapter fails to raise error
When I configured Backend to use SQLite and did this:

```
$ zmb start_domain_test --domain example.com | jq .
{
  "jsonrpc": "2.0",
  "result": null,
  "id": 1
}
```

We should never be getting `"result": null` here. Something is wrong.

```
$ tail -n4 /home/mattiasp/.log/zonemaster/zm-rpcapi.log
DBD::SQLite::db selectrow_array failed: no such table: test_results at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/DB/SQLite.pm line 190.
DBD::SQLite::db do failed: no such table: test_results at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/DB/SQLite.pm line 223.
DBD::SQLite::db selectrow_array failed: no such table: test_results at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/DB/SQLite.pm line 226.
Use of uninitialized value $id in numeric gt (>) at /home/mattiasp/zonemaster-backend/lib/Zonemaster/Backend/DB/SQLite.pm line 228.
```

So I forgot to initialize my database. That's obviously wrong. But it should result in an error response. We're getting a successful response with `"result": null`. Also we're making three failed database queries. We should be aborting after the first failure.

mattias-p:
This properly initialized the SQLite handle to throw exceptions on error instead of just setting $DBI::errstr and returning an error value.

Fixes #722.

### DIFF
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -36,6 +36,7 @@ sub BUILD {
 
         my $dbh = DBI->connect(
             $connection_string,
+            undef, undef,
             {
                 AutoCommit => 1,
                 RaiseError => 1,


matsduf (approved):

pnax (approved):

pnax:

Based on #722, I've verified that using SQLite, there is always a valid `result`.
I've also verified that with an uninitialized database, an error is returned :

1. change the `database_file` to `/var/lib/zonemaster/db-EMPTY.sqlite`
2. restart the daemons
3. `zmb start_domain_test --domain example.com | jq .`

Tested on Debian 10, Ubuntu 18.04 and Ubuntu 20.04.

PR #725, opened 2021-03-19
Fix logical type errors

### LINKED ISSUE #722: SQLite adapter fails to raise error

### LINKED ISSUE #724: Broken domain name validation
Domain name validation was broken in release 6.0.1.

In version 5.0.2 I get:
```
$ zmb start_domain_test --domain in/va/lid.se
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32603,
    "message": "The domain name contains unauthorized character(s) \n"
  }
}
```

In version 6.0.1 I get:
```
$ zmb start_domain_test --domain in/va/lid.se
{
  "id": 1,
  "result": "8943ee6faf1f9af6",
  "jsonrpc": "2.0"
}
```

Note that MySQL and PostgreSQL were the only supported databases back in version 5.0.2.

mattias-p:
This re-enables the domain name validation we had back in 5.0.2.

With these changes the error response for invalid domain names caught by _check_domain is "Internal server error". This is less than great, but at least it's much better than accepting invalid strings.

Fixes #724.

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -198,53 +198,42 @@ sub get_data_from_parent_zone {
 sub _check_domain {
     my ( $self, $dn, $type ) = @_;
 
-    ( $dn, my $result ) = eval {
-        if ( !defined( $dn ) ) {
-            return ( $dn, { status => 'nok', message => encode_entities( "$type required" ) } );
-        }
+    if ( !defined( $dn ) ) {
+        return ( $dn, { status => 'nok', message => encode_entities( "$type required" ) } );
+    }
 
-        if ( $dn =~ m/[^[:ascii:]]+/ ) {
-            if ( Zonemaster::LDNS::has_idn() ) {
-                eval { $dn = Zonemaster::LDNS::to_idn( $dn ); };
-                if ( $@ ) {
-                    return (
-                        $dn,
-                        {
-                            status  => 'nok',
-                            message => encode_entities( "The domain name is not a valid IDNA string and cannot be converted to an A-label" )
-                        }
-                    );
-                }
-            }
-            else {
+    if ( $dn =~ m/[^[:ascii:]]+/ ) {
+        if ( Zonemaster::LDNS::has_idn() ) {
+            eval { $dn = Zonemaster::LDNS::to_idn( $dn ); };
+            if ( $@ ) {
                 return (
                     $dn,
                     {
-                        status => 'nok',
-                        message =>
-                        encode_entities( "$type contains non-ascii characters and IDNA is not installed" )
+                        status  => 'nok',
+                        message => encode_entities( "The domain name is not a valid IDNA string and cannot be converted to an A-label" )
                     }
                 );
             }
         }
-
-        if( $dn !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
+        else {
             return (
                 $dn,
                 {
-                status  => 'nok',
-                message => encode_entities( "The domain name character(s) not supported")
+                    status  => 'nok',
+                    message => encode_entities( "$type contains non-ascii characters and IDNA is not installed" )
                 }
             );
         }
-
-        return ( $dn, undef );
-    };
-    if ($@) {
-        handle_exception('_check_domain', $@, '007');
     }
-    elsif ($result) {
-		return ( $dn, $result );
+
+    if ( $dn !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
+        return (
+            $dn,
+            {
+                status  => 'nok',
+                message => encode_entities( "The domain name character(s) not supported" )
+            }
+        );
     }
 
     my %levels = Zonemaster::Engine::Logger::Entry::levels();
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -238,6 +238,7 @@ sub _check_domain {
             );
         }
 
+        return ( $dn, undef );
     };
     if ($@) {
         handle_exception('_check_domain', $@, '007');
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -198,7 +198,7 @@ sub get_data_from_parent_zone {
 sub _check_domain {
     my ( $self, $dn, $type ) = @_;
 
-    my $result = eval {
+    ( $dn, my $result ) = eval {
         if ( !defined( $dn ) ) {
             return ( $dn, { status => 'nok', message => encode_entities( "$type required" ) } );
         }
@@ -243,7 +243,7 @@ sub _check_domain {
         handle_exception('_check_domain', $@, '007');
     }
     elsif ($result) {
-        return $result;
+		return ( $dn, $result );
     }
 
     my %levels = Zonemaster::Engine::Logger::Entry::levels();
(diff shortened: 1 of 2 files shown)


matsduf (approved):

mattias-p:
I'd forgotten to take one of the branches in _check_domain into account, so I added a commit to fix that.

In order to isolate the error more easily I updated some of the tests in the failing test file to more clearly isolate tests from each other, to more clearly isolate execution phases from verification phases, to give slightly more feedback, and to not modify the global $frontend_params variable.

matsduf (approved):

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -193,7 +193,7 @@ sub get_data_from_parent_zone {
 sub _check_domain {
     my ( $self, $dn, $type ) = @_;
 
-    my $result = eval {
+    ( $dn, my $result ) = eval {
```
pnax, line 196: Just wondering if this variable name `dn` has a meaning ? I know this is not a new name added in this PR, I was just wondering if someone knew what `dn` stands for. And if this is not explicit enough, why not update the name to something more obvious.

pnax (commented):
I like the refactoring of the `test_validate_syntax.t` file. Just 2 questions on the code. Beside those this looks good to me.

lib/Zonemaster/Backend/RPCAPI.pm
mattias-p: To me it's clearly an abbreviation of "domain name". But I agree that it's not great. I'll change it.

mattias-p:
When I had another look at this I realized the eval/handle_exception construct in _check_domain is useless. It only makes sense to have eval/handle_exception in top-level procedures. So I threw it out from _check_domain.

Initially I was going to get rid of all instances of `$dn`. But then `get_data_from_parent_zone` is doing weird things and for this PR I don't feel like digging into it, so I'm leaving it as is.

lib/Zonemaster/Backend/RPCAPI.pm
matsduf: I think it would help to be consistent on what variable name to use when referring to zone name -- or domain name, if you wish -- to be tested. And to use that variable every time it works, and to reserve it for that.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
...
+            if ( $@ ) {
                 return (
-                    $dn,
+                    $domain,
                     {
-                        status => 'nok',
-                        message =>
-                        encode_entities( "$type contains non-ascii characters and IDNA is not installed" )
+                        status  => 'nok',
+                        message => encode_entities( "The domain name is not a valid IDNA string and cannot be converted to an A-label" )
```
matsduf, line 208: To be correct, a domain name is not an A-label. An A-label is a label within a domain name.

"The domain name has one or more non-ASCII label that cannot be converted to an A-label"

mattias-p:
@pnax, @matsduf Please review again.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
 
 
 sub _check_domain {
-    my ( $self, $dn, $type ) = @_;
+    my ( $self, $domain, $type ) = @_;
```
matsduf, line 194: I think it would help if the API of this method is defined. Especially spell out what "$type" is for.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
...
-        if( $dn !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
+        else {
             return (
-                $dn,
+                $domain,
                 {
-                status  => 'nok',
-                message => encode_entities( "The domain name character(s) not supported")
+                    status  => 'nok',
+                    message => encode_entities( "$type contains non-ascii characters and IDNA is not installed" )
```
matsduf, line 218: "$type contains non-ascii label but support for conversion from U-label to A-label is not installed"

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
...
     }
-    elsif ($result) {
-		return $result;
+
+    if ( $domain !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
+        return (
+            $domain,
+            {
+                status  => 'nok',
+                message => encode_entities( "The domain name character(s) not supported" )
```
matsduf, line 229: "The domain name contains one or more characters not supprted"

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
...
     
     my %levels = Zonemaster::Engine::Logger::Entry::levels();
     my @res;
-    @res = Zonemaster::Engine::Test::Basic->basic00($dn);
+    @res = Zonemaster::Engine::Test::Basic->basic00( $domain );
     @res = grep { $_->numeric_level >= $levels{ERROR} } @res;
-    if (@res != 0) {
-        return ( $dn, { status => 'nok', message => encode_entities( "$type name or label is too long" ) } );
+    if ( @res != 0 ) {
+        return ( $domain, { status => 'nok', message => encode_entities( "$type name or label is too long" ) } );
```
matsduf, line 239: "$type name or at least one of its labels is too long"

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -191,65 +191,55 @@ sub get_data_from_parent_zone {
...
     }
 
-    return ( $dn, { status => 'ok', message => 'Syntax ok' } );
+    return ( $domain, { status => 'ok', message => 'Syntax ok' } );
```
matsduf, line 242: "Valid name"

matsduf (changes requested):

matsduf (approved):
An issue of the incorrect messages has been created in #735.

mattias-p:

I've verified this using the command in #724 on CentOS 7 and CentOS 8.

PR #726, opened 2021-03-19
Remove configuration that allows non-hash test ID

matsduf:
This PR resolves issue #691. The use of configuration key "force_hash_id_use_in_API_starting_from_id" has been deprecated by #694.

This PR simplifies the database engine codes considerable and removes a feature that we see no real use of.

### DIFF
--- a/lib/Zonemaster/Backend/DB/PostgreSQL.pm
+++ b/lib/Zonemaster/Backend/DB/PostgreSQL.pm
@@ -77,18 +77,17 @@ sub user_authorized {
 sub test_progress {
     my ( $self, $test_id, $progress ) = @_;
 
-    my $id_field = $self->_get_allowed_id_field_name($test_id);
     my $dbh = $self->dbh;
     if ( $progress ) {
         if ($progress == 1) {
-            $dbh->do( "UPDATE test_results SET progress=?, test_start_time=NOW() WHERE $id_field=? AND progress <> 100", undef, $progress, $test_id );
+            $dbh->do( "UPDATE test_results SET progress=?, test_start_time=NOW() WHERE hash_id=? AND progress <> 100", undef, $progress, $test_id );
         }
         else {
-            $dbh->do( "UPDATE test_results SET progress=? WHERE $id_field=? AND progress <> 100", undef, $progress, $test_id );
+            $dbh->do( "UPDATE test_results SET progress=? WHERE hash_id=? AND progress <> 100", undef, $progress, $test_id );
         }
     }
     
-    my ( $result ) = $dbh->selectrow_array( "SELECT progress FROM test_results WHERE $id_field=?", undef, $test_id );
+    my ( $result ) = $dbh->selectrow_array( "SELECT progress FROM test_results WHERE hash_id=?", undef, $test_id );
 
     return $result;
 }
@@ -120,7 +119,6 @@ sub create_new_batch_job {
 
 sub create_new_test {
     my ( $self, $domain, $test_params, $minutes_between_tests_with_same_params, $batch_id ) = @_;
-    my $result;
     my $dbh = $self->dbh;
 
     $test_params->{domain} = $domain;
@@ -143,17 +141,10 @@ sub create_new_test {
 
     my $nb_inserted = $dbh->do( $query );
 
-    my ( $id, $hash_id ) = $dbh->selectrow_array(
-        "SELECT id, hash_id FROM test_results WHERE params_deterministic_hash='$test_params_deterministic_hash' ORDER BY id DESC LIMIT 1" );
-        
-    if ( $id > $self->config->force_hash_id_use_in_API_starting_from_id() ) {
-        $result = $hash_id;
-    }
-    else {
-        $result = $id;
-    }
+    my ( undef, $hash_id ) = $dbh->selectrow_array(
+        "SELECT id,hash_id FROM test_results WHERE params_deterministic_hash=? ORDER BY id DESC LIMIT 1", undef, $test_params_deterministic_hash );
 
-    return $result;
+    return $hash_id;
 }
 
 sub get_test_params {
@@ -162,8 +153,7 @@ sub get_test_params {
     my $result;
 
     my $dbh = $self->dbh;
-    my $id_field = $self->_get_allowed_id_field_name($test_id);
-    my ( $params_json ) = $dbh->selectrow_array( "SELECT params FROM test_results WHERE $id_field=?", undef, $test_id );
+    my ( $params_json ) = $dbh->selectrow_array( "SELECT params FROM test_results WHERE hash_id=?", undef, $test_id );
     eval { $result = decode_json( encode_utf8( $params_json ) ); };
     die "$@ \n" if $@;
 
@@ -174,14 +164,13 @@ sub test_results {
     my ( $self, $test_id, $results ) = @_;
 
     my $dbh = $self->dbh;
-    my $id_field = $self->_get_allowed_id_field_name($test_id);
-    $dbh->do( "UPDATE test_results SET progress=100, test_end_time=NOW(), results = ? WHERE $id_field=? AND progress < 100",
+    $dbh->do( "UPDATE test_results SET progress=100, test_end_time=NOW(), results = ? WHERE hash_id=? AND progress < 100",
         undef, $results, $test_id )
       if ( $results );
 
     my $result;
     eval {
-        my ( $hrefs ) = $dbh->selectall_hashref( "SELECT id, hash_id, creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time, params, results FROM test_results WHERE $id_field=?", $id_field, undef, $test_id );
+        my ( $hrefs ) = $dbh->selectall_hashref( "SELECT id, hash_id, creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time, params, results FROM test_results WHERE hash_id=?", 'hash_id', undef, $test_id );
         $result            = $hrefs->{$test_id};
         
         # This workaround is needed to properly handle all versions of perl and the DBD::Pg module 
@@ -210,7 +199,6 @@ sub get_test_history {
 
     my $dbh = $self->dbh;
 
-    my $use_hash_id_from_id = $self->config->force_hash_id_use_in_API_starting_from_id();
     my $undelegated = "";
     if ($p->{filter} eq "undelegated") {
         $undelegated = "AND (params->'nameservers') IS NOT NULL";
@@ -248,7 +236,7 @@ sub get_test_history {
         push(
             @results,
             {
-                id               => ($h->{id} > $use_hash_id_from_id)?($h->{hash_id}):($h->{id}),
+                id               => $h->{hash_id},
                 creation_time    => $h->{creation_time},
                 overall_result   => $overall_result,
             }
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -188,20 +188,15 @@ sub create_new_test {
 
     # Search for recent test result with the test same parameters, where "$minutes"
     # gives the time limit for how old test result that is accepted.
-    my ( $recent_id, $recent_hash_id ) = $dbh->selectrow_array(
-        "SELECT id, hash_id FROM test_results WHERE params_deterministic_hash = ? AND test_start_time > DATETIME('now', '-$minutes minutes')",
+    my ( $recent_hash_id ) = $dbh->selectrow_array(
+        "SELECT hash_id FROM test_results WHERE params_deterministic_hash = ? AND test_start_time > DATETIME('now', '-$minutes minutes')",
         undef,
         $test_params_deterministic_hash,
     );
 
-    if ( $recent_id ) {
+    if ( $recent_hash_id ) {
         # A recent entry exists, so return its id
-        if ( $recent_id > $self->config->force_hash_id_use_in_API_starting_from_id() ) {
-            $result_id = $recent_hash_id;
-        }
-        else {
-            $result_id = $recent_id;
-        }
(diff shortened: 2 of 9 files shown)


t/test_DB_backend.pl
```diff
@@ -103,9 +102,10 @@ sub run_zonemaster_test_with_backend_API {
 my $test_history =
 	$engine->get_test_history( { frontend_params => $frontend_params_1, offset => $offset, limit => $limit } );
 diag explain( $test_history );
-ok( scalar( @$test_history ) == 2 );
-ok( $test_history->[0]->{id} eq '1' || $test_history->[1]->{id} eq '1' );
-ok( length($test_history->[0]->{id}) == 16 || length($test_history->[1]->{id}) == 16 );
+ok( scalar( @$test_history ) == 2 ), 'Two tests created';
+
+ok( length($test_history->[0]->{id}) == 16 && length($test_history->[1]->{id}) == 16 ),
+    'Created tests has 16 character lenght hash ID';
```
mattias-p, line 108: If you split this up into two separate calls to ok() it'll be easier to know exactly which condition failed.

mattias-p (commented):
I just skimmed the code changes and they look pretty good so far.
Have you though about the compatibility consequences?

matsduf:
> Have you though about the compatibility consequences?

Can you clarify what compatibility consequence that you are thinking of?

mattias-p:
I'm thinking of how people with existing installations will be affected by this change.

Is the new version a drop-in replacement for the old one with respect to this change? I guess Backend won't complain about the unknown option in the config file since we're don't really have any validation, but does the behavior change in any user-observable way for people with this option set to different values? If there is a change in behavior we should probably defer this removal until we have proper validation of the config file.

mattias-p:
Also, #715 mentions this option.

matsduf:
> I'm thinking of how people with existing installations will be affected by this change.

Referring to an old test will only work with hash_id. Otherwise the test history is available as before.

mattias-p:
So in case any old users use this format they'll need to extract a translation table mapping integer id to hash id when upgrading. Chances are we don't have any such users, right? At the very least it should be mentioned in the release notes.

And in effect test ids matching `'^[0-9]$|^[1-9][0-9]{1,8}$` are still technically valid but they never match anything? We should probably update API.md and Validation.pm to reject those values.

But the numeric ids in the databases are just single-column primary keys? Those should probably be kept as is.

matsduf:
> So in case any old users use this format they'll need to extract a translation table mapping integer id to hash id when upgrading. Chances are we don't have any such users, right? At the very least it should be mentioned in the release notes.

I should be mentioned, and I find it unlikely that we have those users.

> And in effect test ids matching `'^[0-9]$|^[1-9][0-9]{1,8}$` are still technically valid but they never match anything? We should probably update API.md and Validation.pm to reject those values.

We could validate, but I do not think it is necessary.

> But the numeric ids in the databases are just single-column primary keys? Those should probably be kept as is.

Yes, the numeric id should be kept as the unique identifier and is used for sorting.

mattias-p:
> We could validate, but I do not think it is necessary.

As a matter of hygiene we really should. The only thing that comes out of accepting test ids in `^[0-9]$|^[1-9][0-9]{1,8}$` is people spending time trying to understand and support it, for no good reason.

matsduf:
> As a matter of hygiene we really should. The only thing that comes out of accepting test ids in `^[0-9]$|^[1-9][0-9]{1,8}$` is people spending time trying to understand and support it, for no good reason.

I am not against adding such a validation. It should then be done after this has been merged.

mattias-p:
> It should then be done after this has been merged.

That's fine with me.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -143,17 +141,10 @@ sub create_new_test {
...
-        "SELECT id, hash_id FROM test_results WHERE params_deterministic_hash='$test_params_deterministic_hash' ORDER BY id DESC LIMIT 1" );
-        
-    if ( $id > $self->config->force_hash_id_use_in_API_starting_from_id() ) {
-        $result = $hash_id;
-    }
-    else {
-        $result = $id;
-    }
+    my ( undef, $hash_id ) = $dbh->selectrow_array(
+        "SELECT id,hash_id FROM test_results WHERE params_deterministic_hash='$test_params_deterministic_hash' ORDER BY id DESC LIMIT 1" );
```
mattias-p, line 145: Have you considered updating this to use `?` instead of string interpolation?

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -174,14 +164,13 @@ sub test_results {
...
-    my $id_field = $self->_get_allowed_id_field_name($test_id);
-    $dbh->do( "UPDATE test_results SET progress=100, test_end_time=NOW(), results = ? WHERE $id_field=? AND progress < 100",
+    $dbh->do( "UPDATE test_results SET progress=100, test_end_time=NOW(), results = ? WHERE hash_id=? AND progress < 100",
         undef, $results, $test_id )
       if ( $results );
 
     my $result;
     eval {
-        my ( $hrefs ) = $dbh->selectall_hashref( "SELECT id, hash_id, creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time, params, results FROM test_results WHERE $id_field=?", $id_field, undef, $test_id );
+        my ( $hrefs ) = $dbh->selectall_hashref( "SELECT id, hash_id, creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time, params, results FROM test_results WHERE hash_id=?", 'hash_id', undef, $test_id );
```
mattias-p, line 173: This looks strange to me. What does the `'hash_id'` argument do?

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -187,20 +187,15 @@ sub create_new_test {
 
     # Search for recent test result with the test same parameters, where "$minutes"
     # gives the time limit for how old test result that is accepted.
-    my ( $recent_id, $recent_hash_id ) = $dbh->selectrow_array(
-        "SELECT id, hash_id FROM test_results WHERE params_deterministic_hash = ? AND test_start_time > DATETIME('now', '-$minutes minutes')",
+    my ( $recent_hash_id ) = $dbh->selectrow_array(
+        "SELECT hash_id FROM test_results WHERE params_deterministic_hash = ? AND test_start_time > DATETIME('now', '-$minutes minutes')",
```
mattias-p, line 191: Have you considered updating this to use ? instead of string interpolation?

mattias-p (commented):
This clean up is such an improvement.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
matsduf: Not really because I wanted to limit the change to what is needed. Would that be a considerable improvement? If so, I could do that change.

lib/Zonemaster/Backend/DB/SQLite.pm
matsduf: Not really because I wanted to limit the change to what is needed. Would that be a considerable improvement? If so, I could do that change.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
mattias-p: Yes, it would definitely be an improvement. It makes it much easier to make code reviews with regard to SQL injections.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
mattias-p: Right, thanks. I was looking at the wrong definition.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
matsduf: In the updated commits the change is already implemented.

lib/Zonemaster/Backend/DB/SQLite.pm
matsduf: In the updated commits the requested change is already implemented.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
matsduf: I also wondered what is was, and had to look it up, so that is why I had a link to the documentation available.

matsduf:
@mattias-p and @pnax, please review.

mattias-p (dismissed):

pnax:
I haven't review this PR yet, but looking at the the description, I agree that I'll need to review #715 once this is merged.

mattias-p:
> I haven't review this PR yet, but looking at the the description, I agree that I'll need to review #715 once this is merged.

Either that or we merge #715 first and resolve the conflict as part of this PR instead.

t/test_DB_backend.pl
```diff
@@ -103,9 +102,10 @@ sub run_zonemaster_test_with_backend_API {
 my $test_history =
 	$engine->get_test_history( { frontend_params => $frontend_params_1, offset => $offset, limit => $limit } );
 diag explain( $test_history );
-ok( scalar( @$test_history ) == 2 );
-ok( $test_history->[0]->{id} eq '1' || $test_history->[1]->{id} eq '1' );
-ok( length($test_history->[0]->{id}) == 16 || length($test_history->[1]->{id}) == 16 );
+ok( scalar( @$test_history ) == 2 ), 'Two tests created';
+
+ok( length($test_history->[0]->{id}) == 16 ),'Test 0 has 16 character lenght hash ID';
+ok( length($test_history->[1]->{id}) == 16 ),'Test 1 has 16 character lenght hash ID';
```
pnax, line 108: `lenght` -> `length`

pnax (commented):
Beside the typo, I like how this PR lightens the code.

pnax:
> Either that or we merge #715 first and resolve the conflict as part of this PR instead.

This looks like a good idea, so that this PR does the same for every database which might produce a better code history. What do you think @matsduf ?

t/test_DB_backend.pl
matsduf: Fixed.

matsduf:
@mattias-p and @pnax, please approve again, if you approve.

matsduf:
> > I haven't review this PR yet, but looking at the the description, I agree that I'll need to review #715 once this is merged.
> 
> Either that or we merge #715 first and resolve the conflict as part of this PR instead.

Either @matsduf or @mattias-p has to fix a PR. I am open for either way.

matsduf:
Or else I could "pick" the code in #715 that adds missing "overall" to SQLite and add that to this PR.

pnax:
If it does not bother you, I think it would be nice to harmonize and then remove so that this removal is consistent across all database.

pnax:
I merged #715.

matsduf:
@mattias-p and @pnax, moved to updated develop branch. There were several conflicts in `SQLite.pm` that I had to fix manually. I found no issues when doing test with it, but please inspect that again with your eyes.

t/test_DB_backend.pl
```diff
@@ -103,9 +102,10 @@ sub run_zonemaster_test_with_backend_API {
 my $test_history =
 	$engine->get_test_history( { frontend_params => $frontend_params_1, offset => $offset, limit => $limit } );
 diag explain( $test_history );
-ok( scalar( @$test_history ) == 2 );
-ok( $test_history->[0]->{id} eq '1' || $test_history->[1]->{id} eq '1' );
-ok( length($test_history->[0]->{id}) == 16 || length($test_history->[1]->{id}) == 16 );
+ok( scalar( @$test_history ) == 2 ), 'Two tests created';
```
pnax, line 105: Parenthesis mismatch here.

t/test_DB_backend.pl
```diff
@@ -103,9 +102,10 @@ sub run_zonemaster_test_with_backend_API {
...
+
+ok( length($test_history->[0]->{id}) == 16 ),'Test 0 has 16 character length hash ID';
+ok( length($test_history->[1]->{id}) == 16 ),'Test 1 has 16 character length hash ID';
```
pnax, line 108: same parenthesis problem

pnax (commented):
Thank you for the fixes with SQLite.pm. Just a last small adjustment with some parentheses.

t/test_DB_backend.pl
matsduf: I would say *mis*placed rather than *mis*match. Corrected. The code was legal, but what came after comma was meaningless.

t/test_DB_backend.pl
matsduf: As above.

t/test_DB_backend.pl
pnax: yes you are totally right, I couldn't find that word when commenting yesterday ;)

pnax (approved):

lib/Zonemaster/Backend/DB/SQLite.pm
mattias-p: Did you accidentally revert this when resolving the conflicts?

mattias-p (approved):

lib/Zonemaster/Backend/DB/SQLite.pm
matsduf: I probably never replaced "$minutes" with "?". That can be fixed later.

mattias-p:

I verified this using the commands below on CentOS 7 and CentOS 8.

```
 # Under [ZONEMASTER] add the line "force_hash_id_use_in_API_starting_from_id = 10":
sudoedit /etc/zonemaster/backend_config.ini
journalctl -xe | grep force_hash_id_use_in_API_starting_from_id
# Verify that you got error messages about
# ZONEMASTER.force_hash_id_use_in_API_starting_from_id being unrecognize
# from both zm-rpcapi and zm-testagent
```

PR #730, opened 2021-03-26
Separate config section for each db engine

### LINKED ISSUE #719: Separate config section for each db engine
Today all supported database engines are configured using the same properties in the same section of the backend_config ini file. If you have multiple database engines installed and want to switch between them you need to either update multiple properties in your config to make the switch, or you need to have several config files differing only in the database details. This adds an unnecessary hoop you have to jump through in all sorts of testing related contexts. Also this overloading of properties makes the configuration format documentation more opaque to the reader.

The configuration format should be updated to have a separate section for each supported database engine.
The Config module should support overriding the database engine type programmatically. (This would come in handy for unit tests at the very least.)

The old properties should be deprecated, their semantics should be updated so that they work as fallbacks for the properties in all of the database engine-specific sections. The deprecation message should instruct the user to use the corresponding database specific property instead.

mattias-p:
## Purpose

This PR is about having separate a separate config section for each db engine.

## Context

Fixes #719.

As @matsduf pointed out, PR #717 must be merged before this is merged or else there is a risk of filling up the log files.

## Changes

Config format:
* New sections are added. The new section names are upper-case-only for consistency with previous sections.
* The old properties are deprecated.

Implementation:
* New accessor methods for the properties are added to Config.pm. They're named after the config section and property name, with an underscore in between. There is some precedence for this naming scheme in Config.pm.
* If the new properties are absent in the config file, the old ones are used as fallbacks to make sure old config files keep working.
* The accessors for the old properties are removed.
* The DB_connection_string() method didn't really belong in the Config module so I inlined it into the DB modules.

Installation:
* Newly obsolete steps in the installation instruction are cleaned up.
* The included config file is updated to the new format.

CI:
* The backend config files for travis are updated to the new format.

Database initialization:
* The initialization scripts now use the adapter class - they used to read accessors on the config object and construct its own handle. (Note that the SQLite initialization script is only used in the installation instruction while the MySQL and PostgreSQL dittos are only used for CI.)

Database upgrade:
* The database upgrade scripts are updated the same way as the database initialization scripts.

Logging:
* It's now possible to override ZM_BACKEND_RPCAPI_LOGLEVEL and ZM_BACKEND_TESTAGENT_LOGLEVEL when starting the daemons using the supplied init scripts.
* Logging of database connections is now more informative.
* Log messages emitted during the loading for the config file during startup is now output to the terminal (instead of being silently ignored).

## How to test this PR

This is a new feature, so older versions of Zonemaster need not be tested.

For more convenient testing of the database settings, set ZM_BACKEND_RPCAPI_LOGLEVEL and ZM_BACKEND_TESTAGENT_LOGLEVEL to notice or lower.

For each database engine:
1. Get the source from this PR.
   ```sh
   git clone --origin mattias-p --branch 719-db-sections https://github.com/mattias-p/zonemaster-backend.git
   cd zonemaster-backend
   ```
2. Verify that both daemons can be installed and that the SQLite database initialization works.
   1. Install Backend for according to the installation instruction, but instead of `cpanm Zonemaster::Backend` run `cpanm .`.
3. Verify that the installation uses the new configuration format.
      ```sh
      cat /etc/zonemaster/backend_config.ini
      ```
4. Verify that both daemons can be started using the old configuration format.
   1. Copy the share/backend_config.ini from the current develop branch into /etc/zonemaster.
      ```sh
      git remote add origin https://github.com/zonemaster/zonemaster-backend.git
      git fetch origin
      git checkout origin/develop
      sudo install -v -m 640 -g zonemaster share/backend_config.ini /etc/zonemaster/
      ```
   2. Update the DB.engine property for the current database engine according to the installation instruction.
   3. Stop the daemons.
      ```sh
      sudo service stop zm-rpcapi
      sudo sercive status zm-rpcapi
      sudo service stop zm-testagent
      sudo sercive status zm-testagent
      ```
   4. Start the daemons.
      ```sh
      sudo service start zm-rpcapi
      sudo sercive status zm-rpcapi
      sudo service start zm-testagent
      sudo sercive status zm-testagent
      ```
5. Verify that deprecation warnings are emitted to the log files of both rpcapi and testagent if the old configurations properties are present when the configuration file is loaded.
   ```sh
   cat /var/log/zonemaster/zm-testagent.log | grep deprecated
   journalctl -u zm-rpcapi.service | grep deprecated
   ```

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -35,19 +35,11 @@ SQLite            | `SQLite`
 
 **Deprecated.** Use MYSQL.user or POSTGRESQL.user instead.
 
-The name of the user with sufficient permission to access the database.
-
-Ignored by the SQLite database engine.
-
 The MYSQL.user and POSTGRESQL.user properties take precedense over this.
 
 ### password
 
 **Deprecated.** Use MYSQL.password or POSTGRESQL.password instead.
-
-The password of the configured user.
-
-Ignored by the SQLite database engine.
 
 The MYSQL.password and POSTGRESQL.password properties take precedense over this.
 
@@ -55,18 +47,11 @@ The MYSQL.password and POSTGRESQL.password properties take precedense over this.
 
 **Deprecated.** Use MYSQL.host or POSTGRESQL.host instead.
 
-The host name of the machine on which the engine is running.
-
-Ignored by the SQLite database engine.
-
 The MYSQL.host and POSTGRESQL.host properties take precedense over this.
 
 ### database_name
 
 **Deprecated.** Use MYSQL.database, POSTGRESQL.database or SQLite.file instead.
-
-The name of the database to use, except for SQLite database engine where
-it holds the full path to the SQLite database file.
 
 The MYSQL.database, POSTGRESQL.database, SQLite.file properties take precedense
 over this.
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -51,9 +51,9 @@ The [MYSQL.host] and [POSTGRESQL.host] properties take precedence over this.
 
 ### database_name
 
-**Deprecated.** Use [MYSQL.database], [POSTGRESQL.database] or [SQLite.file] instead.
+**Deprecated.** Use [MYSQL.database], [POSTGRESQL.database] or [SQLITE.database_file] instead.
 
-The [MYSQL.database], [POSTGRESQL.database], [SQLite.file] properties take precedence
+The [MYSQL.database], [POSTGRESQL.database], [SQLITE.database_file] properties take precedence
 over this.
 
 ### polling_interval
@@ -298,7 +298,7 @@ zero minutes, then the default value (600 seconds) is used.
 [Profile JSON files]:                 https://github.com/zonemaster/zonemaster-engine/blob/master/docs/Profiles.md
 [Profile names]:                      API.md#profile-name
 [Profiles]:                           Architecture.md#profile
-[SQLite.file]:                        #file
+[SQLITE.database_file]:               #database_file
 [Zonemaster-Engine share directory]:  https://github.com/zonemaster/zonemaster-engine/tree/master/share
 [Zonemaster::Engine::Profile]:        https://metacpan.org/pod/Zonemaster::Engine::Profile#PROFILE-PROPERTIES
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -33,34 +33,42 @@ SQLite            | `SQLite`
 
 ### user
 
+**Deprecated.** Use MYSQL.user or POSTGRESQL.user instead.
+
 The name of the user with sufficient permission to access the database.
 
 Ignored by the SQLite database engine.
 
-The MySQL.user and PostgreSQL.user properties take precedense over this.
+The MYSQL.user and POSTGRESQL.user properties take precedense over this.
 
 ### password
+
+**Deprecated.** Use MYSQL.password or POSTGRESQL.password instead.
 
 The password of the configured user.
 
 Ignored by the SQLite database engine.
 
-The MySQL.password and PostgreSQL.password properties take precedense over this.
+The MYSQL.password and POSTGRESQL.password properties take precedense over this.
 
 ### database_host
+
+**Deprecated.** Use MYSQL.host or POSTGRESQL.host instead.
 
 The host name of the machine on which the engine is running.
 
 Ignored by the SQLite database engine.
 
-The MySQL.host and PostgreSQL.host properties take precedense over this.
+The MYSQL.host and POSTGRESQL.host properties take precedense over this.
 
 ### database_name
+
+**Deprecated.** Use MYSQL.database, POSTGRESQL.database or SQLite.file instead.
 
 The name of the database to use, except for SQLite database engine where
 it holds the full path to the SQLite database file.
 
-The MySQL.database, PostgreSQL.database, SQLite.file properties take precedense
+The MYSQL.database, POSTGRESQL.database, SQLite.file properties take precedense
 over this.
 
 ### polling_interval
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -121,11 +121,11 @@ If this property is unspecified, the value of [DB.database_name] is used instead
 
 ## SQLITE section
 
-Available keys : `file`.
+Available keys : `database_file`.
 
-### file
+### database_file
 
-The full path to the SQLite database file.
+The full path to the SQLite main database file.
 
 If this property is unspecified, the value of [DB.database_name] is used instead.
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -12,7 +12,6 @@ made to the `backend_config.ini` file.
 
 ## DB section
 
-The DB section has a number of keys.
 Available keys : `engine`, `user`, `password`, `database_name`,
 `database_host`, `polling_interval`.
 
@@ -38,11 +37,15 @@ The name of the user with sufficient permission to access the database.
 
 Ignored by the SQLite database engine.
 
+The MySQL.user and PostgreSQL.user properties take precedense over this.
+
 ### password
 
 The password of the configured user.
 
 Ignored by the SQLite database engine.
+
+The MySQL.password and PostgreSQL.password properties take precedense over this.
 
 ### database_host
 
@@ -50,14 +53,88 @@ The host name of the machine on which the engine is running.
 
 Ignored by the SQLite database engine.
 
+The MySQL.host and PostgreSQL.host properties take precedense over this.
+
 ### database_name
 
 The name of the database to use, except for SQLite database engine where
 it holds the full path to the SQLite database file.
 
+The MySQL.database, PostgreSQL.database, SQLite.file properties take precedense
+over this.
+
 ### polling_interval
 
 Time in seconds between database lookups by Test Agent.
+
+
+## MYSQL section
+
+Available keys : `host`, `user`, `password`, `database`.
+
+### host
+
+The host name of the machine on which the MYSQL server is running.
+
+If this property is unspecified, the value of DB.database_host is used instead.
+
+### user
+
+The name of the user with sufficient permission to access the database.
+
+If this property is unspecified, the value of DB.user is used instead.
+
+### password
+
+The password of the configured user.
+
+If this property is unspecified, the value of DB.password is used instead.
+
+### database
+
+The name of the database to use.
+
+If this property is unspecified, the value of DB.database_name is used instead.
+
+
+## POSTGRESQL section
+
+Available keys : `host`, `user`, `password`, `database`.
+
+### host
+
+The host name of the machine on which the PostgreSQL server is running.
+
+If this property is unspecified, the value of DB.database_host is used instead.
+
+### user
+
+The name of the user with sufficient permission to access the database.
+
+If this property is unspecified, the value of DB.user is used instead.
+
+### password
+
+The password of the configured user.
+
+If this property is unspecified, the value of DB.password is used instead.
+
+### database
+
+The name of the database to use.
+
+If this property is unspecified, the value of DB.database_name is used instead.
+
+
+## SQLITE section
+
+Available keys : `file`.
+
+### file
+
+The full path to the SQLite database file.
+
+If this property is unspecified, the value of DB.database_name is used instead.
 
 
 ## LANGUAGE section
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -278,27 +278,27 @@ zero minutes, then the default value (600 seconds) is used.
 
 --------
 
-[DB.database_host]:                   #db-section
-[DB.database_name]:                   #db-section
-[DB.password]:                        #db-section
-[DB.user]:                            #db-section
+[DB.database_host]:                   #database_host
+[DB.database_name]:                   #database_name
+[DB.password]:                        #password
+[DB.user]:                            #user
 [Default JSON profile file]:          https://github.com/zonemaster/zonemaster-engine/blob/master/share/profile.json
 [ISO 3166-1 alpha-2]:                 https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
 [ISO 639-1]:                          https://en.wikipedia.org/wiki/ISO_639-1
 [Installation instructions]:          Installation.md
 [Language tag]:                       API.md#language-tag
-[MYSQL.database]:                     #mysql-section
-[MYSQL.host]:                         #mysql-section
-[MYSQL.password]:                     #mysql-section
-[MYSQL.user]:                         #mysql-section
-[POSTGRESQL.database]:                #postgresql-section
-[POSTGRESQL.host]:                    #postgresql-section
-[POSTGRESQL.password]:                #postgresql-section
-[POSTGRESQL.user]:                    #postgresql-section
+[MYSQL.database]:                     #database
+[MYSQL.host]:                         #host
+[MYSQL.password]:                     #password-1
+[MYSQL.user]:                         #user-1
+[POSTGRESQL.database]:                #database-1
+[POSTGRESQL.host]:                    #host-1
+[POSTGRESQL.password]:                #password-2
+[POSTGRESQL.user]:                    #user-2
 [Profile JSON files]:                 https://github.com/zonemaster/zonemaster-engine/blob/master/docs/Profiles.md
 [Profile names]:                      API.md#profile-name
 [Profiles]:                           Architecture.md#profile
-[SQLite.file]:                        #sqlite-section
+[SQLite.file]:                        #file
 [Zonemaster-Engine share directory]:  https://github.com/zonemaster/zonemaster-engine/tree/master/share
 [Zonemaster::Engine::Profile]:        https://metacpan.org/pod/Zonemaster::Engine::Profile#PROFILE-PROPERTIES
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -33,27 +33,27 @@ SQLite            | `SQLite`
 
 ### user
 
-**Deprecated.** Use MYSQL.user or POSTGRESQL.user instead.
+**Deprecated.** Use [MYSQL.user] or [POSTGRESQL.user] instead.
 
-The MYSQL.user and POSTGRESQL.user properties take precedense over this.
+The [MYSQL.user] and [POSTGRESQL.user] properties take precedense over this.
 
 ### password
 
-**Deprecated.** Use MYSQL.password or POSTGRESQL.password instead.
+**Deprecated.** Use [MYSQL.password] or [POSTGRESQL.password] instead.
 
-The MYSQL.password and POSTGRESQL.password properties take precedense over this.
+The [MYSQL.password] and [POSTGRESQL.password] properties take precedense over this.
 
 ### database_host
 
-**Deprecated.** Use MYSQL.host or POSTGRESQL.host instead.
+**Deprecated.** Use [MYSQL.host] or [POSTGRESQL.host] instead.
 
-The MYSQL.host and POSTGRESQL.host properties take precedense over this.
+The [MYSQL.host] and [POSTGRESQL.host] properties take precedense over this.
 
 ### database_name
 
-**Deprecated.** Use MYSQL.database, POSTGRESQL.database or SQLite.file instead.
+**Deprecated.** Use [MYSQL.database], [POSTGRESQL.database] or [SQLite.file] instead.
 
-The MYSQL.database, POSTGRESQL.database, SQLite.file properties take precedense
+The [MYSQL.database], [POSTGRESQL.database], [SQLite.file] properties take precedense
 over this.
 
 ### polling_interval
@@ -69,25 +69,25 @@ Available keys : `host`, `user`, `password`, `database`.
 
 The host name of the machine on which the MYSQL server is running.
 
-If this property is unspecified, the value of DB.database_host is used instead.
+If this property is unspecified, the value of [DB.database_host] is used instead.
 
 ### user
 
 The name of the user with sufficient permission to access the database.
 
-If this property is unspecified, the value of DB.user is used instead.
+If this property is unspecified, the value of [DB.user] is used instead.
 
 ### password
 
 The password of the configured user.
 
-If this property is unspecified, the value of DB.password is used instead.
+If this property is unspecified, the value of [DB.password] is used instead.
 
 ### database
 
 The name of the database to use.
 
-If this property is unspecified, the value of DB.database_name is used instead.
+If this property is unspecified, the value of [DB.database_name] is used instead.
 
 
 ## POSTGRESQL section
@@ -98,25 +98,25 @@ Available keys : `host`, `user`, `password`, `database`.
 
 The host name of the machine on which the PostgreSQL server is running.
 
-If this property is unspecified, the value of DB.database_host is used instead.
+If this property is unspecified, the value of [DB.database_host] is used instead.
 
 ### user
 
 The name of the user with sufficient permission to access the database.
 
-If this property is unspecified, the value of DB.user is used instead.
+If this property is unspecified, the value of [DB.user] is used instead.
 
 ### password
 
 The password of the configured user.
 
-If this property is unspecified, the value of DB.password is used instead.
+If this property is unspecified, the value of [DB.password] is used instead.
 
 ### database
 
 The name of the database to use.
 
-If this property is unspecified, the value of DB.database_name is used instead.
+If this property is unspecified, the value of [DB.database_name] is used instead.
 
 
 ## SQLITE section
@@ -127,7 +127,7 @@ Available keys : `file`.
 
 The full path to the SQLite database file.
 
-If this property is unspecified, the value of DB.database_name is used instead.
+If this property is unspecified, the value of [DB.database_name] is used instead.
 
 
 ## LANGUAGE section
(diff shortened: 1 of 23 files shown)


docs/Configuration.md
```diff
@@ -31,32 +30,118 @@ SQLite            | `SQLite`
 
 ### user
 
+**Deprecated.** Use MYSQL.user or POSTGRESQL.user instead.
+
 The name of the user with sufficient permission to access the database.
 
 Ignored by the SQLite database engine.
```
matsduf, line 38: I think the four lines above can be removed to keep it shorter. The references to the new configuration makes it very clear.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -37,6 +37,20 @@ sub load_config {
...
+    }
+    if ( defined $self->{cfg}->val( 'DB', 'user' ) ) {
+        $log->warning( "Use of deprecated config property DB.user. Use MYSQL.user or POSTGRESQL.user instead." );
+    }
+    if ( defined $self->{cfg}->val( 'DB', 'password' ) ) {
+        $log->warning( "Use of deprecated config property DB.password. Use MYSQL.password or POSTGRESQL.password instead." );
+    }
+    if ( defined $self->{cfg}->val( 'DB', 'database_name' ) ) {
+        $log->warning( "Use of deprecated config property DB.database_name. Use MYSQL.database, POSTGRESQL.database or SQLITE.file instead." );
+    }
```
matsduf, line 52: PR #717 must be merged before this is merged or else there is a risk of filling up the log files.

matsduf:
This looks good.

lib/Zonemaster/Backend/Config.pm
mattias-p: I included this comment in the PR description.

docs/Configuration.md
mattias-p: I like to keep the documentation of deprecated APIs even if we provide a migration path away from them. But sure, I'll remove it this time. I think it's a good thing that we stop and think about how we deprecate things.

docs/Configuration.md
```diff
@@ -31,32 +30,103 @@ SQLite            | `SQLite`
 
 ### user
 
-The name of the user with sufficient permission to access the database.
+**Deprecated.** Use [MYSQL.user] or [POSTGRESQL.user] instead.
 
-Ignored by the SQLite database engine.
+The [MYSQL.user] and [POSTGRESQL.user] properties take precedense over this.
```
pnax, line 35: `precedense` -> `precedence`

docs/Configuration.md
```diff
@@ -31,32 +30,103 @@ SQLite            | `SQLite`
...
 
 ### password
 
-The password of the configured user.
+**Deprecated.** Use [MYSQL.password] or [POSTGRESQL.password] instead.
 
-Ignored by the SQLite database engine.
+The [MYSQL.password] and [POSTGRESQL.password] properties take precedense over this.
```
pnax, line 41: `precedense` -> `precedence`

docs/Configuration.md
```diff
@@ -31,32 +30,103 @@ SQLite            | `SQLite`
...
 
 ### database_host
 
-The host name of the machine on which the engine is running.
+**Deprecated.** Use [MYSQL.host] or [POSTGRESQL.host] instead.
 
-Ignored by the SQLite database engine.
+The [MYSQL.host] and [POSTGRESQL.host] properties take precedense over this.
```
pnax, line 47: `precedense` -> `precedence`

docs/Configuration.md
```diff
@@ -31,32 +30,103 @@ SQLite            | `SQLite`
...
 
 ### database_name
 
-The name of the database to use, except for SQLite database engine where
-it holds the full path to the SQLite database file.
+**Deprecated.** Use [MYSQL.database], [POSTGRESQL.database] or [SQLite.file] instead.
+
+The [MYSQL.database], [POSTGRESQL.database], [SQLite.file] properties take precedense
```
pnax, line 53: `precedense` -> `precedence`

docs/Configuration.md
```diff
@@ -31,32 +30,103 @@ SQLite            | `SQLite`
...
 Time in seconds between database lookups by Test Agent.
 
 
+## MYSQL section
+
+Available keys : `host`, `user`, `password`, `database`.
+
+### host
+
+The host name of the machine on which the MYSQL server is running.
```
pnax, line 67: This one can have a lowercase `y`, `MYSQL` -> `MySQL` ;-)

docs/Configuration.md
```diff
@@ -205,15 +275,28 @@ zero minutes, then the default value (600 seconds) is used.
 
 --------
 
-[Installation instructions]:          Installation.md
+[DB.database_host]:                   #db-section
+[DB.database_name]:                   #db-section
+[DB.password]:                        #db-section
+[DB.user]:                            #db-section
```
pnax, line 281: Why not pointing to the right subsection directly ?

docs/Configuration.md
```diff
@@ -205,15 +275,28 @@ zero minutes, then the default value (600 seconds) is used.
...
+[Installation instructions]:          Installation.md
+[Language tag]:                       API.md#language-tag
+[MYSQL.database]:                     #mysql-section
+[MYSQL.host]:                         #mysql-section
+[MYSQL.password]:                     #mysql-section
+[MYSQL.user]:                         #mysql-section
+[POSTGRESQL.database]:                #postgresql-section
+[POSTGRESQL.host]:                    #postgresql-section
+[POSTGRESQL.password]:                #postgresql-section
+[POSTGRESQL.user]:                    #postgresql-section
```
pnax, line 294: same

docs/Configuration.md
```diff
@@ -205,15 +275,28 @@ zero minutes, then the default value (600 seconds) is used.
...
 [Profile JSON files]:                 https://github.com/zonemaster/zonemaster-engine/blob/master/docs/Profiles.md
 [Profile names]:                      API.md#profile-name
 [Profiles]:                           Architecture.md#profile
+[SQLite.file]:                        #sqlite-section
```
pnax, line 298: same

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
     }
 }
 
-sub DB_user {
-    my ($self) = @_;
+=head2 MYSQL_database
+
+Returns the L<MYSQL.database|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#mysql>
```
pnax, line 77: The fragment should be `#mysql-section`, it is even possible to point directly to the right section with `#database`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+property from the loaded config, or the L<DB.database_name|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 78: The fragments should be `#db-section`, it is even possible to point directly to the right section with `#database_name`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+property if it is unspecified.
+
+=cut
 
-    return $self->{cfg}->val( 'DB', 'user' );
+sub MYSQL_database {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'MYSQL', 'database' ) // $self->{cfg}->val( 'DB', 'database_name' );
```
pnax, line 86: Will it fail if `DB.database_name` is unspecified ?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+
+    return $self->{cfg}->val( 'MYSQL', 'host' ) // $self->{cfg}->val( 'DB', 'database_host' );
 }
 
-sub DB_name {
-    my ($self) = @_;
+=head2 MYSQL_password
+
+Returns the L<MYSQL.password|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#mysql>
+property from the loaded config, or the L<DB.password|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 106: `#mysql-section` or `#password-1`
`#db-section` or `#password`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+sub MYSQL_password {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'MYSQL', 'password' ) // $self->{cfg}->val( 'DB', 'password' );
+}
+
+=head2 MYSQL_user
+
+Returns the L<MYSQL.user|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#mysql>
+property from the loaded config, or the L<DB.user|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 120: `#mysql-section` or `#user-1`
`#db-section` or `#user`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+sub MYSQL_user {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'MYSQL', 'user' ) // $self->{cfg}->val( 'DB', 'user' );
+}
+
+=head2 POSTGRESQL_database
+
+Returns the L<POSTGRESQL.database|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#postgresql>
+property from the loaded config, or the L<DB.database_name|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 134: `#postgresql-section` or `#database-1`
`#db-section` or `#database_name`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'POSTGRESQL', 'database' ) // $self->{cfg}->val( 'DB', 'database_name' );
+}
 
-    return $self->{cfg}->val( 'DB', 'database_name' );
+=head2 POSTGRESQL_host
+
+Returns the L<POSTGRESQL.host|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#postgresql>
+property from the loaded config, or the L<DB.database_host|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 148: `#postgresql-section` or `#host-1`
`#db-section` or `#database_host`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+sub POSTGRESQL_host {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'POSTGRESQL', 'host' ) // $self->{cfg}->val( 'DB', 'database_host' );
+}
+
+=head2 POSTGRESQL_password
+
+Returns the L<POSTGRESQL.password|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#postgresql>
+property from the loaded config, or the L<DB.password|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 162: `#postgresql-section` or `#password-1`
`#db-section` or `#password`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+sub POSTGRESQL_password {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'POSTGRESQL', 'password' ) // $self->{cfg}->val( 'DB', 'password' );
+}
+
+=head2 POSTGRESQL_user
+
+Returns the L<POSTGRESQL.user|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#postgresql>
+property from the loaded config, or the L<DB.user|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 176: `#postgresql-section` or `#user-2`
`#db-section` or `#user`

lib/Zonemaster/Backend/Config.pm
```diff
@@ -58,22 +72,130 @@ sub BackendDBType {
...
+sub POSTGRESQL_user {
+    my ( $self ) = @_;
+
+    return $self->{cfg}->val( 'POSTGRESQL', 'user' ) // $self->{cfg}->val( 'DB', 'user' );
+}
+
+=head2 SQLITE_file
+
+Returns the L<SQLITE.file|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#sqlite>
+property from the loaded config, or the L<DB.database_name|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#db>
```
pnax, line 190: `#postgresql-section` or `#file`
`#db-section` or `#database_name`

pnax (commented):
This looks really good !

docs/Configuration.md
mattias-p: I was thinking that the password and user headings were ambiguous and I wanted to keep things consistent for the others too. I'd forgotten that Github has a way of disambiguating the headings.

Thanks for pointing this out!

lib/Zonemaster/Backend/Config.pm
mattias-p: It's quite error prone to link to the auto-suffixed anchors. But let's try it out. If it's too painful we'll go back to just linking to the headings.

lib/Zonemaster/Backend/Config.pm
mattias-p: It'll return undef, and the caller will most likely fail at some point. But it's like that for all the properties. I plan to take care of this in #685.

docs/Configuration.md
pnax: Yes I totally agree that this is not the easiest to maintain.

lib/Zonemaster/Backend/Config.pm
pnax: Ok, we'll see how it goes, to be honest I don't have a strong opinion on this. I just wanted to know if you made this choice on purpose and I now totally understand your first move. Now that you updated it, let's go for a try ;)

mattias-p:
I've rebased this PR onto the current develop branch and updated the testing instructions.

It kindof sucks that the rpcapi logs only go the systemd journal. I haven't checked where they go on FreeBSD. Maybe /var/log/messages? The reason is that the rpcapi currently doesn't integrate well with starman when it comes to logging. If you like I can update this PR to also include an improvement to the log integration, or if you'd prefer to have this merged first and have the log integration fixed in a follow-up PR, that's okay with me too.

matsduf:
> It kindof sucks that the rpcapi logs only go the systemd journal. I haven't checked where they go on FreeBSD. Maybe /var/log/messages? 

No, there is no trace of those log messages in FreeBSD as far as I can see.

> The reason is that the rpcapi currently doesn't integrate well with starman when it comes to logging. If you like I can update this PR to also include an improvement to the log integration, or if you'd prefer to have this merged first and have the log integration fixed in a follow-up PR, that's okay with me too.

I prefer to have logging updated in a separate PR. It is better to have several, but smaller PRs when possible.

mattias-p:
I created #736 to addresses the mishandling of the log messages.

@matsduf, @pnax Please, review.

docs/Installation.md
```diff
@@ -591,7 +583,6 @@ path:
 
 ```sh
 sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= SQLite/' /usr/local/etc/zonemaster/backend_config.ini
-sed -i '' '/[[:<:]]database_name[[:>:]]/ s:=.*:= /var/db/zonemaster/db.sqlite:' /usr/local/etc/zonemaster/backend_config.ini
```
matsduf, line 594: In FreeBSD, the location of the SQLite database file should be /var/db/zonemaster, not /var/lib/zonemaster.

matsduf (changes requested):

docs/Installation.md
mattias-p: Good catch! I'll add that line back.

matsduf (dismissed):

docs/Installation.md
```diff
@@ -591,7 +583,7 @@ path:
 
 ```sh
 sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= SQLite/' /usr/local/etc/zonemaster/backend_config.ini
-sed -i '' '/[[:<:]]database_name[[:>:]]/ s:=.*:= /var/db/zonemaster/db.sqlite:' /usr/local/etc/zonemaster/backend_config.ini
+sed -i '' '/[[:<:]]file[[:>:]]/ s:=.*:= /var/db/zonemaster/db.sqlite:' /usr/local/etc/zonemaster/backend_config.ini
```
matsduf, line 586: For this replacement to work, there cannot be any other key "file", which is not unlikely since that is a generic term. Maybe "databasefile" would be better?

mattias-p: Apparently SQLite can use more than one file. https://sqlite.org/fileformat.html
I'll rename it to database_file.

docs/Configuration.md
```diff
@@ -34,32 +33,103 @@ SQLite            | `SQLite`
...
+**Deprecated.** Use [MYSQL.host] or [POSTGRESQL.host] instead.
 
-Ignored by the SQLite database engine.
+The [MYSQL.host] and [POSTGRESQL.host] properties take precedence over this.
 
 ### database_name
 
-The name of the database to use, except for SQLite database engine where
-it holds the full path to the SQLite database file.
+**Deprecated.** Use [MYSQL.database], [POSTGRESQL.database] or [SQLite.file] instead.
```
matsduf, line 54: "SQLite.file" => "SQLite.database_file"

docs/Configuration.md
```diff
@@ -34,32 +33,103 @@ SQLite            | `SQLite`
...
+
+The [MYSQL.database], [POSTGRESQL.database], [SQLite.file] properties take precedence
```
matsduf, line 56: "SQLite.file" => "SQLite.database_file"

docs/Configuration.md
```diff
@@ -208,15 +278,28 @@ zero minutes, then the default value (600 seconds) is used.
...
+[MYSQL.password]:                     #password-1
+[MYSQL.user]:                         #user-1
+[POSTGRESQL.database]:                #database-1
+[POSTGRESQL.host]:                    #host-1
+[POSTGRESQL.password]:                #password-2
+[POSTGRESQL.user]:                    #user-2
 [Profile JSON files]:                 https://github.com/zonemaster/zonemaster-engine/blob/master/docs/Profiles.md
 [Profile names]:                      API.md#profile-name
 [Profiles]:                           Architecture.md#profile
+[SQLite.file]:                        #file
```
matsduf, line 301:
```
[SQLite.database_file]:     #database_file
```

share/backend_config.ini
```diff
@@ -3,12 +3,23 @@
...
+database = zonemaster
+
+[POSTGRESQL]
+host     = localhost
+user     = zonemaster
+password = zonemaster
+database = zonemaster
+
+[SQLITE]
+file = /var/lib/zonemaster/db.sqlite
```
matsduf, line 21: "file" => "database_file"

share/travis_sqlite_backend_config.ini
```diff
@@ -1,14 +1,11 @@
...
-user=travis_zonemaster
-password=travis_zonemaster
-database_host=localhost
-# If using SQLite, database_name is the full filename for the db
-database_name=/tmp/zonemaster
 polling_interval=0.5
 #seconds
 
+[SQLITE]
+file=/tmp/zonemaster
```
matsduf, line 7: "file" => "database_file"

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -31,18 +31,17 @@ sub BUILD {
     my ( $self ) = @_;
 
     if ( !defined $self->dbh ) {
-        my $connection_string = $self->config->DB_connection_string( 'sqlite' );
-        $log->debug( "Connection string: " . $connection_string );
+        my $file = $self->config->SQLITE_file;
```
matsduf, line 34:
```
my $file = $self->config->SQLITE_database_file;
```

matsduf (changes requested):

docs/Configuration.md
mattias-p: My God I missed a lot of them.

matsduf (approved):
I assume that all "file" <=> "database_file" have been found.

mattias-p:
@pnax Could you review this again?

pnax:
Yes, sorry, I'll do it.

share/zm-testagent.lsb
```diff
@@ -20,6 +20,8 @@ PIDFILE=${ZM_BACKEND_PIDFILE:-/var/run/zonemaster/zm-testagent.pid}
 USER=${ZM_BACKEND_USER:-zonemaster}
 GROUP=${ZM_BACKEND_GROUP:-zonemaster}
 
+#export ZM_BACKEND_TESTAGENT_LOGLEVEL='warning'  # Set this variable to override the default log level
```
pnax, line 23: `ZM_BACKEND_TESTAGENT_LOGLEVEL` is not used in `zonemaster_backend_testagent`. Maybe an omission.

Furthermore this might be in conflict with the option `--loglevel`.

share/zm_testagent-bsd
```diff
@@ -16,6 +16,7 @@ load_rc_config $name
 : ${zm_testagent_pidfile="/var/run/zonemaster/${name}.pid"}
 
 export ZONEMASTER_BACKEND_CONFIG_FILE="/usr/local/etc/zonemaster/backend_config.ini"
+#export ZM_BACKEND_TESTAGENT_LOGLEVEL='warning'  # Set this variable to override the default log level
```
pnax, line 19: same remark on `ZM_BACKEND_TESTAGENT_LOGLEVEL`.

pnax (commented):
See my remark on `ZM_BACKEND_TESTAGENT_LOGLEVEL`.

Otherwise, I tested it and I was able :

* to launch the daemons with a new and an old config file
* with the old config file, I can see the deprecation warnings
* with the new config file, I am able to create a test on a domain

About the logs:

* it seems that the `Opening SQLite` message is issued twice (maybe because we test the connection first, then we open it)
* trying with PostgreSQL, each time I send a query to the API, there is the `Connecting to PostgreSQL` message appended (it looks like we reconnect to the database each time, is that so?)

share/zm-testagent.lsb
mattias-p: Good spotting! I made an update to make sure the log level value is propagated.

share/zm_testagent-bsd
mattias-p: I fixed this the same way.

mattias-p:
> About the logs (level NOTICE) :
>
> * it seems that the Opening SQLite message is issued twice (maybe because we test the connection first, then we open it)
> * trying with PostgreSQL, each time I send a query to the API, there is the Connecting to PostgreSQL message appended (it looks like we reconnect to the database each time, is that so?)

Those are good observations! Could you create issues for those? I'd like to limit the scope of this PR. It's already grown way beyond my initial intentions.

share/zm-testagent.lsb
```diff
@@ -20,9 +20,9 @@ PIDFILE=${ZM_BACKEND_PIDFILE:-/var/run/zonemaster/zm-testagent.pid}
 USER=${ZM_BACKEND_USER:-zonemaster}
 GROUP=${ZM_BACKEND_GROUP:-zonemaster}
 
-#export ZM_BACKEND_TESTAGENT_LOGLEVEL='warning'  # Set this variable to override the default log level
+#export ZM_BACKEND_TESTAGENT_LOGLEVEL='info'  # Set this variable to override the default log level
 
-testagent_args="--logfile=$LOGFILE --outfile=$OUTFILE --pidfile=$PIDFILE --user=$USER --group=$GROUP"
+testagent_args="--loglevel=$ZM_BACKEND_TESTAGENT_LOGLEVEL --logfile=$LOGFILE --outfile=$OUTFILE --pidfile=$PIDFILE --user=$USER --group=$GROUP"
```
pnax: If you pass it directly to the arguments, shouldn't the variable be exported then ?

share/zm_testagent-bsd
```diff
@@ -16,10 +16,10 @@ load_rc_config $name
 : ${zm_testagent_pidfile="/var/run/zonemaster/${name}.pid"}
 
 export ZONEMASTER_BACKEND_CONFIG_FILE="/usr/local/etc/zonemaster/backend_config.ini"
-#export ZM_BACKEND_TESTAGENT_LOGLEVEL='warning'  # Set this variable to override the default log level
+#export ZM_BACKEND_TESTAGENT_LOGLEVEL='info'  # Set this variable to override the default log level
```
pnax, line 19: same

pnax (commented):
Sorry to be punctilious, shouldn't the variable `ZM_BACKEND_TESTAGENT_LOGLEVEL` be set then ?

pnax:
> Those are good observations! Could you create issues for those? I'd like to limit the scope of this PR. It's already grown way beyond my initial intentions.

Having a quick look on this, it seems to be the expected behavior. We can keep it like this for now. (Sorry to overload the PR with my remarks)

share/zm-testagent.lsb
mattias-p: Since the variable is only used locally it doesn't need to be exported. And since the default value is defined in zonemaster_backend_testagent I don't want to define it again in the init scripts - for the sake of having a single source of truth.

But you're completely right that there is something fishy going on here! I'm setting --loglevel to the empty string when I ZM_BACKEND_TESTAGENT_LOGLEVEL is undefined. I'll fix that.

mattias-p:
> Sorry to overload the PR with my remarks

In that case I'm sorry for making so many silly mistakes.

pnax (approved):
This looks good, thanks :)

matsduf (approved):

mattias-p:

(However to get it to work I had to replace the `service` commands with `systemctl` dittos.)

PR #731, opened 2021-03-30
Unify tests

pnax:
This is in continuity to what has been discussed in #716. To summarize, the idea is to merge files `t/test_DB_backend.pl` and `t/test01.t` with the following goals in mind :

* call the tests through `make test`
* SQLite should be the default database to use
* to test with another database use an environment variable
* if tests are specific to a database, use [SKIP](https://perldoc.perl.org/Test::More#Conditional-tests) blocks

With this in mind I merged `test_DB_backend.pl` in `test01.t`.
I found that Travis uses the TARGET environment variable, so I reused it here.


t/test01.t
```diff
@@ -83,14 +100,17 @@ sub run_zonemaster_test_with_backend_API {
     }
 
     use_ok( 'Zonemaster::Backend::TestAgent' );
-    Zonemaster::Backend::TestAgent->new( { dbtype => "SQLite", config => $config } )->run( $hash_id );
+    Zonemaster::Backend::TestAgent->new( { dbtype => "$db_backend", config => $config } )->run( $hash_id );
 
     Zonemaster::Backend::TestAgent->reset() unless ( $ENV{ZONEMASTER_RECORD} );
 
+    sleep( 5 ) unless ( $ENV{ZONEMASTER_RECORD} );
```
mattias-p, line 107: I tried to think about this call to sleep. My guess is that the sleep here is intended to give the TestAgent time to pick up the request. But if that is the case, I can't see why it's only needed if we aren't recording. Am I getting something wrong?

t/test01.t
```diff
@@ -68,57 +86,72 @@ my $frontend_params_1 = {
...
+    Zonemaster::Backend::TestAgent->new( { dbtype => "$db_backend", config => $config } )->run( $hash_id );
 
     Zonemaster::Backend::TestAgent->reset() unless ( $ENV{ZONEMASTER_RECORD} );
 
-    ok( $engine->test_progress( $hash_id ) > 0 );
+    sleep( 5 ) unless ( $ENV{ZONEMASTER_RECORD} );
+    cmp_ok( $engine->test_progress( $hash_id ), '>',  0 , 'API test_progress -> Test started');
 
     foreach my $i ( 1 .. 12 ) {
+        sleep( 5 ) unless ( $ENV{ZONEMASTER_RECORD} );
```
mattias-p, line 111: I tried to think about this sleep call too. If we're recording we'll poll twelve times back-to-back, and if we're using a preloaded cache we poll for a minute. Shouldn't it be the other way around?

t/test01.t
```diff
@@ -8,6 +8,15 @@ use Zonemaster::Engine;
...
 use File::ShareDir qw[dist_file];
 
+# Use the TARGET environment variable to set the database to use
+# default to SQLite
+my $db_backend = $ENV{TARGET};
+if ( not $db_backend ) {
+    $db_backend = 'SQLite';
+} elsif ( $db_backend !~ /SQLite|MySQL|PostgreSQL/ ) {
+    BAIL_OUT( "Unsupported database backend: $db_backend" );
+}
```
mattias-p, line 18: Could you add a diagnostic here so it's visible from the TAP output which database engine is beign used?

mattias-p (commented):
I have a suggestion and a couple of questions. Otherwise this looks really good!

t/test01.t
```diff
@@ -8,6 +8,15 @@ use Zonemaster::Engine;
 use JSON::PP;
 use File::ShareDir qw[dist_file];
 
+# Use the TARGET environment variable to set the database to use
+# default to SQLite
+my $db_backend = $ENV{TARGET};
+if ( not $db_backend ) {
+    $db_backend = 'SQLite';
+} elsif ( $db_backend !~ /SQLite|MySQL|PostgreSQL/ ) {
```
matsduf, line 16: To be more precise:
```
$db_backend !~ /^(?:SQLite|MySQL|PostgreSQL)$/
```
or
```
$db_backend ne 'SQLite'  and $db_backend ne 'MySQL' and $db_backend ne 'PostgreSQL'
```

matsduf (dismissed):

t/test01.t
pnax: Oh, I think I misunderstand what and how this `ZONEMASTER_RECORD` environment variable worked.

Now, from my understanding of the code if this variable is unset or egal to `0`, the test will use the data in `t/test01.data`. Otherwise, it will run all the tests as done in production. Is that so ?

And a subsidiary question, when setting `ZONEMASTER_RECORD=1`, will it actually « record » something ? (or should it « record » something ?

matsduf: Yes, 
```
if ( $ENV{ZONEMASTER_RECORD} ) {
	Zonemaster::Engine->save_cache( $datafile );
}
```

mattias-p: Yes, the idea is to make the unit tests run with recorded data by default, so as to make them independent of network connectivity and changes to the internet. When you update Engine so it makes new queries that isn't covered by the recorded data, you need to rerecord it. It's a pretty neat feature!

For completeness I might also add that ZONEMASTER_RECORD also counts as false when it's set to the empty string.

pnax: Ok. I have a better view now. Thanks.

These `sleep()` calls look like they are needed to wait for the tests to finish. However running the code with and without the `ZONEMASTER_RECORD` set to `1`, it looks like the `TestAgent->run()` is blocking. So there might be no need to call these `sleep()`. Unless the code is updated to run the TestAgent in a subprocess (and therefore may be killed if it hangs).

t/test01.t
pnax: Done in 3896241

t/test01.t
pnax: See 1ff939e

t/test01.t
mattias-p: Right. It makes sense that TestAgent is single-threaded. We're using Parallel::ForkManager to manage a set of worker processes, each one of which is using TestAgent.

If we can do without the sleeping now I say good riddance. If we need it later we should probably first look for alternatives before adding it back.

pnax:
I updated with your remarks and suggestions.

I haven't modified the logic around the `sleep` calls because I feel they are unnecessary with the current call the `TestAgent->run()`.
However there is a part of code in the test that I'd like to discuss in [`t/test01.t` (v6.1.0)](https://github.com/zonemaster/zonemaster-backend/blob/master/t/test01.t#L92..L95) :

```
foreach my $i ( 1 .. 12 ) {
    my $progress = $engine->test_progress( $hash_id );
    last if ( $progress == 100 );
}
```

This looks like we want to output the progress of the tests. However `TestAgent->run()` is blocking (please correct me if I missed something). So this loop will only by crossed once. Therefore I suggest to choose either :

1. we wait for the `TestAgent->run()` to return and then we remove this loop
2. we make `TestAgent->run()` non blocking and can then keep the loop (and potentially kill the TestAgent if it takes too long, if this may ever happen, I don't know if the TestAgent has a way to prevent the tests to take forever)

mattias-p:
It's better to remove the loop. We don't gain anything by just making TestAgent->run() non-blocking.

matsduf:
>     2. we make `TestAgent->run()` non blocking and can then keep the loop (and potentially kill the TestAgent if it takes too long, if this may ever happen, I don't know if the TestAgent has a way to prevent the tests to take forever)

In the script `zonemaster_backend_testagent` there is a funktion to kill children:
```
        $self->pm->on_wait();                   # Sends SIGKILL to overdue child processes
```

pnax:
I removed the loop and calls to `sleep`. I also removed a `cmp_ok()` test as the test should be finished when TestAgent->run() returns.

A side question: do you have a way to distinguish between the usage of the word «test» her as it can be a _test_ to test the code or a _test_ to test a domain. Do you have an idea on how to be more explicit when using this word, or even suggestion on better words to use ?

Finally, we start by adding a user, however this user is never used in the following API calls. My guess is that we want to try that the API methods to add a user is working. If so, this will leave the database in a dirty state. If we run the tests again (with the same environment), the user creation test will fail because the user already exists. Should we remove the user (a new method might be needed...), or start by testing that the user does not already exists (and creating if needed) ?

t/test01.t
pnax: The loop has been removed in 741b75b

t/test01.t
pnax: Removed in 741b75b

matsduf:
> A side question: do you have a way to distinguish between the usage of the word «test» her as it can be a _test_ to test the code or a _test_ to test a domain. Do you have an idea on how to be more explicit when using this word, or even suggestion on better words to use ?

I have no better suggestion than to be explicit to write "domain test" and "code test", respectively.

> Finally, we start by adding a user, however this user is never used in the following API calls. My guess is that we want to try that the API methods to add a user is working. If so, this will leave the database in a dirty state. If we run the tests again (with the same environment), the user creation test will fail because the user already exists. Should we remove the user (a new method might be needed...), or start by testing that the user does not already exists (and creating it if needed) ?

I am not sure what you mean with "the same environment". As I understand it, the database will go away when the test script has finished, but it would not be wrong to make sure that you can remove a user if that is supported by the API.

pnax:
> I am not sure what you mean with "the same environment". As I understand it, the database will go away when the test script has finished, but it would not be wrong to make sure that you can remove a user if that is supported by the API.

I am testing the code on my local environment before pushing the branch. Travis creates a completely new environment for each test. This is not the case for me. With SQLite, the database is dropped at the end (it it was created by the test code itself)*. For PostgreSQL and MySQL, there is nothing to "clean" the database, so you cannot run the tests twice with the same database without getting an error on user creation.

* I realize the code for the removal of the SQLite database might not work, because it is looking for a `zonemaster` file and not for the `database_name` entry of the config.

matsduf (approved):

mattias-p (approved):

PR #738, opened 2021-04-12
Cleaning and fixing in the continuity of #731 and #734

pnax:
In #731 when using SQLite, the code suggests that the database file should be deleted if it was created for the tests. I forgot to update the path to the database file. This fixes that.

In #734 , I discover some old code, I decided to not touch it at that time as `test01.t` was using it. This removes this old code and updates `test01.t` accordingly.
There are calls to `test_progress` and to `get_data_from_parent_zone` using the `SCALAR` parameter in [`CodeSnippets/client.pl` (v6.1.0)](https://github.com/zonemaster/zonemaster-backend/blob/v6.1.0/CodeSnippets/client.pl). I'm not sure how up to date this snippet is, so I haven't updated it yet. I totally can if you think it is worth it.

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -151,15 +151,11 @@ $json_schemas{get_data_from_parent_zone} = joi->object->strict->props(
 );
 sub get_data_from_parent_zone {
     my ( $self, $params ) = @_;
-    my $domain = "";
 
     my $result = eval {
         my %result;
-        if (ref \$params eq "SCALAR") {
-            $domain = $params;
-        } else {
-            $domain = $params->{domain};
-        }
+
+        my $domain = $params->{domain};
 
         my ( $dn, $dn_syntax ) = $self->_check_domain( $domain, 'Domain name' );
         return $dn_syntax if ( $dn_syntax->{status} eq 'nok' );
@@ -408,13 +404,7 @@ sub test_progress {
 
     my $result = 0;
     eval {
-        my $test_id = "";
-        if (ref \$params eq "SCALAR") {
-            $test_id = $params;
-        } else {
-            $test_id = $params->{test_id};
-        }
-
+        my $test_id = $params->{test_id};
         $result = $self->{db}->test_progress( $test_id );
     };
     if ($@) {
--- a/t/test01.t
+++ b/t/test01.t
@@ -97,7 +97,7 @@ sub run_zonemaster_test_with_backend_API {
     is( scalar( $engine->{db}->dbh->selectrow_array( qq/SELECT id FROM test_results WHERE id=$test_id/ ) ), $test_id , 'API start_domain_test -> Test inserted in the DB' );
 
     # test test_progress API
-    is( $engine->test_progress( $hash_id ), 0 , 'API test_progress -> OK');
+    is( $engine->test_progress( { test_id => $hash_id } ), 0 , 'API test_progress -> OK');
 
     if ( not $ENV{ZONEMASTER_RECORD} ) {
         Zonemaster::Engine->preload_cache( $datafile );
@@ -113,7 +113,7 @@ sub run_zonemaster_test_with_backend_API {
 
     Zonemaster::Backend::TestAgent->reset() unless ( $ENV{ZONEMASTER_RECORD} );
 
-    is( $engine->test_progress( $hash_id ), 100 , 'API test_progress -> Test finished' );
+    is( $engine->test_progress( { test_id => $hash_id } ), 100 , 'API test_progress -> Test finished' );
 
     my $test_results = $engine->get_test_results( { id => $hash_id, language => 'fr_FR' } );
     ok( defined $test_results->{id},                 'TEST1 $test_results->{id} defined' );
--- a/t/test01.t
+++ b/t/test01.t
@@ -150,7 +150,7 @@ if ( $ENV{ZONEMASTER_RECORD} ) {
 done_testing();
 
 if ( $db_backend eq 'SQLite' ) {
-    my $dbfile = 'zonemaster';
+    my $dbfile = $engine->{db}->dbh->sqlite_db_filename;
     if ( -e $dbfile and -M $dbfile < 0 and -o $dbfile ) {
         unlink $dbfile;
     }


mattias-p:
This looks good!

Though I was thinking about another thing - the TARGET variable. Do we really need it? Can't we just take what's in the DB.engine property?

mattias-p:
Oh, and I believe the CodeSnippets have been obsoleted by zmb.

matsduf:
> There are calls to `test_progress` and to `get_data_from_parent_zone` using the `SCALAR` parameter in [`CodeSnippets/client.pl` (v6.1.0)](https://github.com/zonemaster/zonemaster-backend/blob/v6.1.0/CodeSnippets/client.pl). I'm not sure how up to date this snippet is, so I haven't updated it yet. I totally can if you think it is worth it.

Are the files found in [`CodeSnippets`](https://github.com/zonemaster/zonemaster-backend/blob/v6.1.0/CodeSnippets) used by anyone? They are never included in a Zonemaster-Backend installation. Most of it has not been updated during the last 6-7 years. The set is completely undocumented.

Today we have, as @mattias-p commented, the `zmb` and `zmtest` that are documented, installed and used. They replace the earlier purpose of `CodeSnippets`.

As far as I can see the entire set can be removed instead.

pnax:
> Though I was thinking about another thing - the TARGET variable. Do we really need it? Can't we just take what's in the DB.engine property?

Good point. I don't know. As I see it, there are 2 scenarios:

1. tests run by Travis: in that case everything is already scripted, and will be launched by Travis
2. tests run by a developer or anyone on its own machine: in that case I think it would be nice to be able to easily switch between the database engines, so we should keep the switch as light as possible

Having said that, I see the following options :

* use 3 different config files, with a different `DB.engine` key in each one
* use 1 config file and programmatically override the `DB.engine` key
* use the TARGET variable to wrap either one of the previous point to easily switch between database engine


> The set is completely undocumented.

I think it is documentation by itself. However, I think that the `API` file does that job perfectly well. And as you both said, the `zmb` script can be used to supersede this snippets.

matsduf:
Travis fails to use perl 5.32. It is not limited to this PR. Travis tries to download a file from Travis, but the file does not exist.

matsduf (approved):
To me this lookes fine, in spite of the problem with Perl 5.32 and Travis.

mattias-p:
> tests run by a developer or anyone on its own machine: in that case I think it would be nice to be able to easily switch between the database engines, so we should keep the switch as light as possible

I agree.

> Having said that, I see the following options :
>
> * use 3 different config files, with a different `DB.engine` key in each one
> * use 1 config file and programmatically override the `DB.engine` key
> * use the TARGET variable to wrap either one of the previous point to easily switch between database engine

My take (building on your points) is to build on this like so:
* use 3 different config files, with a different `DB.engine` key in each one
* inline all three config files as strings in the unit tests
* add `Z::B::Config->parse($text)` - a variation on `Z::B::Config->load_config()` that bypasses environment variables and file IO
* use the TARGET variable to easily switch between database engine

That way it'll be easy to add unit tests for config file validation too, which is what I'm planning for #685. If you want to resolve the DB.engine selection as part of this PR that's fine with me. Otherwise I'll happily do it as part of #685.

mattias-p (approved):

pnax:
> Travis fails to use perl 5.32. It is not limited to this PR.

Woah, it looks like this is the case since 5.32 has been added to our Travis config in #645 ^^'
Travis jobs run in #645 (all green) : https://travis-ci.org/github/zonemaster/zonemaster-backend/builds/731258752
for instance the job with MySQL and Perl 5.32 : https://travis-ci.org/github/zonemaster/zonemaster-backend/jobs/731258753 (it is using Perl 5.22)

Luckily the code works like a charm with Perl 5.32 even though it has always been tested against Perl 5.22 ^^'

Good catch @matsduf !

>My take (building on your points) is to build on this like so:
>   * use 3 different config files, with a different DB.engine key in each one

I am wondering, what is the point to keep 3 different files now that we can have one file with a section for each database and that we just need to update the `DB.engine` entry ? Rereading #719 « _you need to have several config files differing only in the database details[, t]his adds an unnecessary loop you have to jump through in all sorts of testing related contexts_ », so it seems one goal was to remove the need to keep multiple config files. If you agree, I think we can go that way by merging the Travis config file, and this will be handled in another PR.


> If you want to resolve the DB.engine selection as part of this PR that's fine with me. Otherwise I'll happily do it as part of #685.

I'll merge this PR as it is for now (see previous paragraph).

matsduf:
> > Travis fails to use perl 5.32. It is not limited to this PR.
> 
> Woah, it looks like this is the case since 5.32 has been added to our Travis config in #645 ^^'
> Travis jobs run in #645 (all green) : https://travis-ci.org/github/zonemaster/zonemaster-backend/builds/731258752
> for instance the job with MySQL and Perl 5.32 : https://travis-ci.org/github/zonemaster/zonemaster-backend/jobs/731258753 (it is using Perl 5.22)
> 
> Luckily the code works like a charm with Perl 5.32 even though it has always been tested against Perl 5.22 ^^'

I have created an issue on this, #740.

mattias-p:
> I am wondering, what is the point to keep 3 different files now that we can have one file with a section for each database and that we just need to update the DB.engine entry ? Rereading #719 « you need to have several config files differing only in the database details[, t]his adds an unnecessary loop you have to jump through in all sorts of testing related contexts », so it seems one goal was to remove the need to keep multiple config files.

The way I'm thinking about this is that we want config format validation, and we can't just validate each parameter individually. We need validation rules that stretch across more than one property. E.g. the SQLite.database_file property is required if and only if DB.engine=SQLite. But then we also want the ability to override individual properties. At least DB.engine, but maybe others too in the future.

At first I thought we needed setter methods to support overriding. And this would necessarily would make the validation layer more complex and less straight forward to review against the validation rules in the config format documentation (given the multi-property validation rules). In this light and with the ability to parse config strings I thought that maybe it wouldn't be so bad after all with a couple of config files sitting next to each other embedded in the test code, especially compared to having them in separate files outside of the test case file.

But your question made me realize that we can just use string interpolation to inject the DB.engine value into the config text in the test file. In fact this would work for any configuration property.

pnax:
> E.g. the SQLite.database_file property is required if and only if DB.engine=SQLite

I don't agree on the « _if and only if_ », I think that all databases sections can be present even if it is not related to the current DB.engine value (because I feel this makes life easier if you need to switch between databases).

> But your question made me realize that we can just use string interpolation to inject the DB.engine value into the config text in the test file. In fact this would work for any configuration property.

I don't know if we need to interpolate all properties... As you feel it.

> Unless you this is a bad idea for some reason I can't see right now I'll go on implementing this.

On the contrary, I think this is a good idea if we can use the fact that switching between database is just as easy as modifying the DB.engine value :)

mattias-p:
> > E.g. the SQLite.database_file property is required if and only if DB.engine=SQLite
>
> I don't agree on the « if and only if », I think that all databases sections can be present even if it is not related to the current DB.engine value (because I feel this makes life easier if you need to switch between databases).

But we do agree! I meant that if ( DB.engine == SQLite ) then { SQLite.database_file is required } else { SQLite.database_file is optional (and ignored) }.

> I don't know if we need to interpolate all properties... As you feel it.

Actually the interpolation is performed in the calling code, so everyone can interpolate as much or as little as they wish.

Issue #740, opened 2021-04-13
Verify that Travis supports configured Perl versions


matsduf:
At the time of writing the [.travis.yml] (develop branch) includes
```
  include:
  # Cover MySQL and PostgreSQL with the latest supported Perl version
  - perl: "5.32"
    env: TARGET=MySQL ZONEMASTER_RECORD=0 ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_mysql_backend_config.ini
    services: mysql
```
but Travis does not support Perl 5.32. It says
```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/perl/binaries/ubuntu/16.04/x86_64/perl-5.32.tar.bz2
curl: (22) The requested URL returned error: 404 Not Found
$ sudo tar xjf perl-5.32.tar.bz2 --directory /
tar (child): perl-5.32.tar.bz2: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
rm: cannot remove 'perl-5.32.tar.bz2': No such file or directory
$ travis_apt_get_update
$ sudo systemctl start mysql
$ git clone --depth=50 https://github.com/zonemaster/zonemaster-backend.git zonemaster/zonemaster-backend

Setting environment variables from .travis.yml
$ export TARGET=MySQL
$ export ZONEMASTER_RECORD=0
$ export ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_mysql_backend_config.ini
0.08s$ perlbrew use 5.32

ERROR: The installation "5.32" is unknown.

$ perl --version

This is perl 5, version 22, subversion 1 (v5.22.1) built for x86_64-linux-gnu-thread-multi
(with 73 registered patches, see perl -V for more detail)
```

To discover this discrepancy you have to inspect the Travis output, because Travis does not report this as an error.

Make sure that this reported as an error by Travis.

[.travis.yml]: https://github.com/zonemaster/zonemaster-backend/blob/develop/.travis.yml

pnax:
I made some testing on what should the string be to use Perl 5.32 in Travis, see #741.

I tried with :
* `5.32.1` : this fails
* `5.32.0` : this works
* `5.32` : I know this should fail

The list of the Travis jobs :
https://travis-ci.org/github/zonemaster/zonemaster-backend/builds/766931924

mattias-p:
[...]

PR #741, opened 2021-04-13
Test Travis and Perl 5.32

pnax:
Just a try to see which Perl version should be defined to support 5.32 in Travis.

Do not merge.

Results :

* `- perl: "5.32.1"` : nope, Travis uses Perl 5.22 and this fails hard (I haven't investigate any further)
* `- perl: "5.32.0"` : this works, `This is perl 5, version 32, subversion 0 (v5.32.0) built for x86_64-linux`
* `- perl: "5.32"` : nope, Travis uses Perl 5.22, `This is perl 5, version 22, subversion 1 (v5.22.1) built for x86_64-linux-gnu-thread-multi`

So I think the string should be updated to `5.32.0` to make Travis use Perl 5.32

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,27 +2,29 @@ language: perl
 
 jobs:
   include:
-  # Cover MySQL and PostgreSQL with the latest supported Perl version
+  - perl: "5.32.1"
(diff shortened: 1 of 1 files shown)


pnax:
The fix has been done in https://github.com/zonemaster/zonemaster-backend/pull/745/commits/30b75efa4828d748d2c11a0bcfdfc6ec2e125e2c

matsduf:
Resolves #740.

PR #745, opened 2021-04-15
Parse config

mattias-p:
## Purpose

This PR adds Zonemaster::Backend::Config->parse(). This method makes unit testing more predictable by letting unit tests embed their configuration(s) instead of having Zonemaster::Backend::Config figure out which file to open and read and risking to use another configuration than the one the developer intended.

An illustration of how such unit tests can be written is provided in t/config.t.

## Context

This PR is made as a first step towards #685, but it also provides a way to update other unit tests to make them more predictable.

## Changes

Zonemaster::Backend::Config:
* Z::B::Config->parse() is broken out from Z::B::Config->load_config().
* Existing validations are moved out from accessor methods and into parse().
* Some drive-by clean-up of the Z::B::Config code.
* A seconds-to-minutes conversion is moved away from Z::B::Config and into business logic code.

Dependencies:
* Add test dependency on Test::NoWarnings. Used in the new t/config.t. It makes unit tests fail if warnings are emitted.
* Install libio-stringy-perl on Debian. It's an undeclared dependency of libconfig-inifiles-perl.

Travis config:
* The version number of to Perl 5.32 is fixed. (Credits to @pnax for investigating.)

## How to test this PR

Run the unit tests. They're automatically run as part of CI but in principle it's a good idea to verify they also work locally.

The new t/config.t file contains tests for all important aspects of Z::B::Config->parse().
The old t/test01.t verifies that the old Z::B::Config->load_config() still works the same way.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -1,9 +1,9 @@
 package Zonemaster::Backend::Config;
-our $VERSION = '1.1.0';
-
 use strict;
 use warnings;
 use 5.14.2;
+
+our $VERSION = '1.1.0';
 
 use Config;
 use Config::IniFiles;
@@ -29,29 +29,30 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
 =cut
 
 sub load_config {
-    my ( $class, $params ) = @_;
-    my $self = {};
+    my ( $class ) = @_;
 
+    my $obj = bless( {}, $class );
+
+    # Load file
     $log->notice( "Loading config: $path" );
 
-    $self->{cfg} = Config::IniFiles->new( -file => $path );
-    die "UNABLE TO LOAD $path ERRORS:[".join('; ', @Config::IniFiles::errors)."] \n" unless ( $self->{cfg} );
-    bless( $self, $class );
+    $obj->{cfg} = Config::IniFiles->new( -file => $path )
+      or die "UNABLE TO LOAD $path ERRORS:[" . join( '; ', @Config::IniFiles::errors ) . "]\n";
 
-    if ( defined $self->{cfg}->val( 'DB', 'database_host' ) ) {
+    if ( defined $obj->{cfg}->val( 'DB', 'database_host' ) ) {
         $log->warning( "Use of deprecated config property DB.database_host. Use MYSQL.host or POSTGRESQL.host instead." );
     }
-    if ( defined $self->{cfg}->val( 'DB', 'user' ) ) {
+    if ( defined $obj->{cfg}->val( 'DB', 'user' ) ) {
         $log->warning( "Use of deprecated config property DB.user. Use MYSQL.user or POSTGRESQL.user instead." );
     }
-    if ( defined $self->{cfg}->val( 'DB', 'password' ) ) {
+    if ( defined $obj->{cfg}->val( 'DB', 'password' ) ) {
         $log->warning( "Use of deprecated config property DB.password. Use MYSQL.password or POSTGRESQL.password instead." );
     }
-    if ( defined $self->{cfg}->val( 'DB', 'database_name' ) ) {
+    if ( defined $obj->{cfg}->val( 'DB', 'database_name' ) ) {
         $log->warning( "Use of deprecated config property DB.database_name. Use MYSQL.database, POSTGRESQL.database or SQLITE.database_file instead." );
     }
 
-    return $self;
+    return $obj;
 }
 
 sub check_db {
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -5,9 +5,10 @@ use 5.14.2;
 
 our $VERSION = '1.1.0';
 
-use Config;
 use Config::IniFiles;
+use Config;
 use File::ShareDir qw[dist_file];
+use File::Slurp qw( read_file );
 use Log::Any qw( $log );
 use Readonly;
 
@@ -31,13 +32,42 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
 sub load_config {
     my ( $class ) = @_;
 
+    $log->notice( "Loading config: $path" );
+    my $text = read_file $path;
+
+    my $obj = eval { $class->parse( $text ) };
+    if ( $@ ) {
+        die "File $path: $@";
+    }
+
+    return $obj;
+}
+
+=head2 parse
+
+Parse a new Zonemaster::Backend::Config from the contents of a
+L<configuration|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md>
+file.
+
+    my $config = Zonemaster::Backend::Config->parse(
+        q{
+            [DB]
+            engine = SQLite
+
+            [SQLITE]
+            database_file = /var/db/zonemaster.sqlite
+        }
+    );
+
+=cut
+
+sub parse {
+    my ( $class, $text ) = @_;
+
     my $obj = bless( {}, $class );
 
-    # Load file
-    $log->notice( "Loading config: $path" );
-
-    $obj->{cfg} = Config::IniFiles->new( -file => $path )
-      or die "UNABLE TO LOAD $path ERRORS:[" . join( '; ', @Config::IniFiles::errors ) . "]\n";
+    $obj->{cfg} = Config::IniFiles->new( -file => \$text )
+      or die "Failed to parse config: " . join( '; ', @Config::IniFiles::errors ) . "\n";
 
     if ( defined $obj->{cfg}->val( 'DB', 'database_host' ) ) {
         $log->warning( "Use of deprecated config property DB.database_host. Use MYSQL.host or POSTGRESQL.host instead." );
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -412,9 +412,9 @@ L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuratio
 
 =head3 RETURNS
 
-A scalar value of the number of minutes old the previous test with the same
+A scalar value of the number of seconds old the previous test with the same
 parameters can be when it is reused instead of starting a new test (default
-10). Note that it returns in minutes, whereas the input is in seconds.
+600).
 
 =cut
 
@@ -423,7 +423,6 @@ sub age_reuse_previous_test {
     my $default = 600; # in seconds
     my $val = $self->{cfg}->val( 'ZONEMASTER', 'age_reuse_previous_test', $default );
     $val = ($val > 0) ? $val : 0;
-    $val = int ( ($val / 60) + 0.5); # in minutes
     return $val;
 }
 
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -59,6 +59,8 @@ file.
         }
     );
 
+Throws an exception if the given configuration file contains errors.
+
 =cut
 
 sub parse {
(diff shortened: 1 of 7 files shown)


matsduf (commented):
Will parse() discover any undefined keys in the config file?

mattias-p:
I believe the answer is no, except what's already being done. This PR is mainly groundwork and should not change anything with regard to the interpretation of the configuration file. Later PRs will follow up to tighten the validation among other things.

matsduf:
> I believe the answer is no, except what's already being done. This PR is mainly groundwork and should not change anything with regard to the interpretation of the configuration file. Later PRs will follow up to tighten the validation among other things.

Just so that I know, it means that if the config file contains anything extra, it will just be silently ignored? -- That is how it is today.

mattias-p:
> Just so that I know, it means that if the config file contains anything extra, it will just be silently ignored? -- That is how it is today.

That is correct. My intention is to keep that as it is in this PR.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -29,29 +30,152 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
...
+        q{
+            [DB]
+            engine = SQLite
+
+            [SQLITE]
+            database_file = /var/db/zonemaster.sqlite
+        }
+    );
+
+Throws an exception if the given configuration file contains errors.
```
matsduf, line 62: I think it makes sense to state, "Undefined sections and keys will be silently ignored", or something similar.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -29,29 +30,152 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
...
+    $obj->{_public_profiles} = {
+        default => '',
+    };
+    for my $name ( $ini->Parameters( 'PUBLIC PROFILES' ) ) {
+        $obj->{_public_profiles}{lc $name} = $ini->val( 'PUBLIC PROFILES', $name );
+    }
+    $obj->{_private_profiles} = {};
+    for my $name ( $ini->Parameters( 'PRIVATE PROFILES' ) ) {
+        $obj->{_private_profiles}{lc $name} = $ini->val( 'PRIVATE PROFILES', $name );
+    }
```
matsduf, line 123: Profile names must match the specification in [API doc] and the profile name must point at a profile file in the file system. I suggest the following tests for profile names:
* Then name must meet the specification.
* The file must exists and being readable by running process.
* The file must be JSON data.

Checking that the file is a valid profile file is too complexed.

[API doc]: https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#profile-name

lib/Zonemaster/Backend/Config.pm
```diff
@@ -29,29 +30,152 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
...
         $log->warning( "Use of deprecated config property DB.database_name. Use MYSQL.database, POSTGRESQL.database or SQLITE.database_file instead." );
+
+        $obj->{_MYSQL_database} = $value
+          if !defined $obj->MYSQL_database;
+
+        $obj->{_POSTGRESQL_database} = $value
+          if !defined $obj->POSTGRESQL_database;
+
+        $obj->{_SQLITE_database_file} = $value
+          if !defined $obj->SQLITE_database_file;
```
matsduf, line 163: If the file does not exist, is not reable or not valid, will that be discovered immediately after the parsing? The equivalent is applicable for the other databases.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -29,29 +30,152 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
...
+    $obj->{_POSTGRESQL_database}                                 = $ini->val( 'POSTGRESQL', 'database',                                 undef );
+    $obj->{_SQLITE_database_file}                                = $ini->val( 'SQLITE',     'database_file',                            undef );
+    $obj->{_ZONEMASTER_max_zonemaster_execution_time}            = $ini->val( 'ZONEMASTER', 'max_zonemaster_execution_time',            '600' );
+    $obj->{_ZONEMASTER_maximal_number_of_retries}                = $ini->val( 'ZONEMASTER', 'maximal_number_of_retries',                '0' );
+    $obj->{_ZONEMASTER_number_of_processes_for_frontend_testing} = $ini->val( 'ZONEMASTER', 'number_of_processes_for_frontend_testing', undef );
+    $obj->{_ZONEMASTER_number_of_processes_for_batch_testing}    = $ini->val( 'ZONEMASTER', 'number_of_processes_for_batch_testing',    undef );
+    $obj->{_ZONEMASTER_lock_on_queue}                            = $ini->val( 'ZONEMASTER', 'lock_on_queue',                            undef );
+    $obj->{_ZONEMASTER_age_reuse_previous_test}                  = $ini->val( 'ZONEMASTER', 'age_reuse_previous_test',                  '600' );
+
```
matsduf, line 102: Some of these should be a non-negative integer, unless undefined. Could this be checked here?

matsduf (changes requested):

matsduf:
This is a very good step forward for bankend.

lib/Zonemaster/Backend/Config.pm
mattias-p: My intention is to limit the scope of this PR by not adding any new validations. In later pull requests I will address all the validations.

While we're on the subject though, the validation I have in mind for filesystem paths is to let  Z::B::Config validate that the value is an absolute path and be satisfied with that. After successfully parsing the configuration but before the daemons fork themselves, the daemons should connect to the database and load all the profile files. This will give us the desired behavior.

lib/Zonemaster/Backend/Config.pm
mattias-p: Yes, the daemons are verifying that they can connect to the database, or else they refuse to start.

lib/Zonemaster/Backend/Config.pm
mattias-p: Yes, I'll add validations for that in a later PR.

lib/Zonemaster/Backend/Config.pm
matsduf: Any comment here?

mattias-p: Yeah, I guess parse() isn't much of a goal keeper at this point. Though it does detect a few errors.

If you insist I'll change the wording now and then change it back in the next PR. (Though the validation layer won't be complete until the PR after that.)

matsduf: Yes include that, and I'll approve.

mattias-p:
I made a couple of updates.
* A note about unrecognized sections and properties being silently ignored was added per request from @matsduf.
* Underscores are no longer used in example host names in t/config.t.
* The deprecated properties under the DB section are now only applied to the database engine selected by DB.engine. Otherwise loading an old configuration file would guarantee the assignment of at least one invalid value.

lib/Zonemaster/Backend/Config.pm
mattias-p: I added a note on this.

lib/Zonemaster/Backend/Config.pm
matsduf: Ok.

lib/Zonemaster/Backend/Config.pm
matsduf: Ok.

matsduf (approved):

pnax (approved):
This looks great !

Just a question, I had a look to the Travis outputs and with [the job using MySQL](https://travis-ci.org/github/zonemaster/zonemaster-backend/jobs/767298669) and [the one using PostrgreSQL](https://travis-ci.org/github/zonemaster/zonemaster-backend/jobs/767298670), I see this following line :

> Use of uninitialized value $file in concatenation (.) or string at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/DB/SQLite.pm line 37.

It looks like SQLite is called at some point and I can't get why. I tried to reproduce without success :/

lib/Zonemaster/Backend/Config.pm
```diff
@@ -29,29 +30,153 @@ Readonly my @SIG_NAME => split ' ', $Config{sig_name};
...
         $log->warning( "Use of deprecated config property DB.database_name. Use MYSQL.database, POSTGRESQL.database or SQLITE.database_file instead." );
+
+        $obj->{_MYSQL_database} = $value
+          if $obj->BackendDBType eq 'MySQL' && !defined $obj->MYSQL_database;
+
+        $obj->{_POSTGRESQL_database} = $value
+          if $obj->BackendDBType eq 'PostgreSQL' && !defined $obj->POSTGRESQL_database;
+
+        $obj->{_SQLITE_database_file} = $value
+          if $obj->BackendDBType eq 'SQLite' && !defined $obj->SQLITE_database_file;
```
matsduf, line 164: Is this correct or should it be
```
&& !defined $obj->{_SQLITE_database_file};
```

mattias-p:
@pnax Good work spotting this warning! I made a little investigation and I think the result is quite something.

First I noticed that the same warning is emitted in 746. I added a call to [cluck](https://perldoc.perl.org/Carp) in that PR to see what's going on because I didn't want to mess with your approvals in this one.

Here's the stack trace:
```
SQLITE_database_file is unset! at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/DB/SQLite.pm line 37.

	Zonemaster::Backend::DB::SQLite::BUILD(Zonemaster::Backend::DB::SQLite=HASH(0x65496b8), HASH(0x6557720)) called at constructor Zonemaster::Backend::DB::SQLite::new (defined at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/DB/SQLite.pm line 416) line 50

	Zonemaster::Backend::DB::SQLite::new("Zonemaster::Backend::DB::SQLite", HASH(0x6557678)) called at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/RPCAPI.pm line 68

	eval {...} called at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/RPCAPI.pm line 64

	Zonemaster::Backend::RPCAPI::_init_db(Zonemaster::Backend::RPCAPI=HASH(0x6549358), "SQLite") called at /home/travis/build/zonemaster/zonemaster-backend/build_dir/lib/Zonemaster/Backend/RPCAPI.pm line 56

	Zonemaster::Backend::RPCAPI::new("Zonemaster::Backend::RPCAPI", HASH(0x650dbc0)) called at t/test_validate_syntax.t line 25
```

It turns out that t/test_validate_syntax.t has a hardcoded reference to SQLite. This goes against the grain with our overriding ZONEMASTER_BACKEND_CONFIG_FILE with a MySQL/PostgreSQL based configuration. Before #730 this happened to work because DB.database_name was set to either "/tmp/zonemaster" for SQLite and "zonemaster" for the others. Both are perfectly valid file system paths – and this also explains why we were mysteriously sometimes getting a SQLite file named zonemaster in the zonemaster-backend repository root!

After #730 this also happens to work but in a different way. At this point we've separated the MYSQL.database and SQLITE.database_file. This makes `t/test_validate_syntax.t` tell `Z::B::DB::SQLite` to construct an instance while `Z::B::Config->SQLITE_database_file` is `undef`. So when `Z::B::DB::SQLite->new` builds the data source name for DBI this warning about an uninitialized value is emitted, and the data source name ends up being `"dbi:SQLite:dbname="`.

But why the test pass then? It turns out that in DBD::SQLite [the empty dbname has a special meaning](https://metacpan.org/pod/DBD::SQLite#Database-Name-Is-A-File-Name) – a temp file is created and used for the database. And the temp file is automatically cleaned up for us when the connection is closed. This accidental behavior is actually more suitable than what we're doing in the SQLite Travis configuration!

What is the moral of this story? I guess that in Perl it's customary to solve the problem of syntax errors by assigning meaning to every conceivable input. In effect the principle of minimal surprise is inverted to become the principle of maximum happy surprise! I feel like Alice in wonderland.

What do we do about this? I suggest we 1) make t/test_validate_config.t take its DB.engine from config, and 2) make Travis always back its SQLite database files with auto-cleaned up temp files. I'll make a separate PR for that.

lib/Zonemaster/Backend/Config.pm
mattias-p: One is equivalent to the other. I think it's nicer maintenance-wise to call the method because that way we're limiting the number of places where the private field name is used.

pnax:
Thank you for this detailed explanation on the origin of this warning message.

> I tried to reproduce without success :/

Well, I was looking at the wrong file, thanks for spotting the hard codded value in t/test_validate_syntax.t. Now I can reproduce the error perfectly well :)

> 1) make t/test_validate_config.t take its DB.engine from config,

Agreed

> 2) make Travis always back its SQLite database files with auto-cleaned up temp files.

Good idea

PR #749, opened 2021-04-22
Uniform config property getter methods

mattias-p:
## Context

This PR lays some groundwork for #685.
It also fixes a warning that was noted by @pnax in https://github.com/zonemaster/zonemaster-backend/pull/745#pullrequestreview-638608629.

## Scope

* No behavioral changes with regard to validation.
* No documentation changes to property getters.
* No code changes with regard to complex properties.
* New property getters are introduced, but the old ones are kept in place. Call sites aren't updated to use the new ones either. At first I wanted to include this cleanup in the PR, but the surface area was large enough that I'm leaving that for a follow-up PR.

## Changes

Format specification:
* Some details are included that were previously only documented in the POD of the getter methods.
* Simple properties undergo minor editing to make them more structurally similar to each other.

Implementation:
* Uniform getter methods are added for all simple properties.
* Defaults are added to all simple properties where it makes sense.
* All uniform getter method implementations are gathered up into a table-like section. IMHO the code is more readable this way as opposed to having each method placed next to its own POD. Pod::Coverage is still able to help us make sure all the methods are documented. Nothing interesting is going on in the method implementations anyway.
* The `SUBROUTINES` POD heading is replaced with `CONSTRUCTORS` and `METHODS`.

Extra:
* t/test_validate_syntax.t:
  1. The warning from https://github.com/zonemaster/zonemaster-backend/pull/745#pullrequestreview-638608629 is fixed
  2. The test now fails if it emits any warnings

## How to test this PR

1. Verify that t/confg.t contains unit tests for all default values in the format specification.
2. Verify that t/test_validate_syntax.t uses Test::NoWarnings.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -252,34 +252,49 @@ The ZONEMASTER section has several keys :
 
 ### max_zonemaster_execution_time
 
-Time in seconds before reporting an unfinished test as failed. Default
-value: `600`.
+An integer.
+Time in seconds before reporting an unfinished test as failed.
+Default value: `600`.
 
 ### maximal_number_of_retries
 
+An integer.
 Number of time a test is allowed to be run again if unfinished after
-`max_zonemaster_execution_time`. Default value: `0`.
+`max_zonemaster_execution_time`.
+Default value: `0`.
+
+This option is experimental and all edge cases are not fully tested.
+Do not use it (keep the default value "0"), or use it with care.
 
 ### number_of_processes_for_frontend_testing
+
+A positive integer.
 
 used -> todo
 
 ### number_of_processes_for_batch_testing
 
+An integer.
+
 used -> todo
 
 ### lock_on_queue
 
-Integer working as a label to associate a test to a specific Test Agent.
+An integer.
+A label to associate a test to a specific Test Agent.
+Default value: `0`.
 
 ### age_reuse_previous_test
 
-Positiv integer (in seconds) for how old a previous test of the same
-zone name and parameters must be before we start a new test. Internally
-the value is converted to whole minutes. If the conversion results in
-zero minutes, then the default value (600 seconds) is used.
+Positive integer.
+How old (in seconds) a previous test of the same zone name and parameters must
+be before we start a new test.
+Default value: `600`.
 
---------
+Internally the value is converted to whole minutes.
+If the conversion results in zero minutes, then the default value is used.
+
+
 
 [DB.database_host]:                   #database_host
 [DB.database_name]:                   #database_name
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -59,6 +59,7 @@ over this.
 ### polling_interval
 
 Time in seconds between database lookups by Test Agent.
+Default value: `0.5`.
 
 
 ## MYSQL section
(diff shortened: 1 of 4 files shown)


docs/Configuration.md
```diff
@@ -252,34 +253,53 @@ The ZONEMASTER section has several keys :
...
 ### maximal_number_of_retries
 
+An integer.
 Number of time a test is allowed to be run again if unfinished after
-`max_zonemaster_execution_time`. Default value: `0`.
+`max_zonemaster_execution_time`.
+Default value: `0`.
+
+This option is experimental and all edge cases are not fully tested.
+Do not use it (keep the default value "0"), or use it with care.
```
matsduf, line 268: This comment is not directly for this PR, but just a note. We should not keep things as "experimental" forever.

mattias-p: I created #750 to stabilize or remove it.

matsduf (commented):
This looks fine, but the profile names should also be validated so that they match permitted name format.

mattias-p:
> profile names should also be validated so that they match permitted name format.

Yes, definitely. I have a plan for those but they're a couple PRs away. At first I want to get the simple properties under control. I.e. those that can be read using a getter method whose name corresponds to its section and property names, and whose validation can be done in a reasonable way using only scalar values. This includes everything but the profiles and LANGUAGE.locale.

matsduf:
> This includes everything but the profiles and LANGUAGE.locale.

Validation of locale is already there, isn't it?

matsduf (approved):

lib/Zonemaster/Backend/Config.pm
```diff
@@ -59,7 +65,16 @@ file.
         }
     );
 
-Throws an exception if the given configuration file contains errors.
+The configuration is interpreted according to the
+L<configuration format specification|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md>.
+
+Returns a new Z::B::Config instance with its properties set to values according
```
pnax, line 71: I think it would be better to avoid abbreviation in documentation and explicitly write `Zonemaster::Backend::Config`.

pnax (commented):
This looks good, just a remark on an abbreviation usage

lib/Zonemaster/Backend/Config.pm
mattias-p: I agree completely. I'll make an update.

matsduf: Mee to. I think we should not use such a breviation anywhere, not even in issue or PR descriptions.

pnax (approved):

matsduf (approved):

pnax:
@mattias-p I think that the following code has not been updated with this PR

https://github.com/zonemaster/zonemaster-backend/blob/50f3acc2f9427cf36b4cf2189fb6e54fb4cf2c32/lib/Zonemaster/Backend/Config.pm#L186
https://github.com/zonemaster/zonemaster-backend/blob/50f3acc2f9427cf36b4cf2189fb6e54fb4cf2c32/lib/Zonemaster/Backend/Config.pm#L192

Shouldn't it be
```
if !defined $obj->ZONEMASTER_number_of_processes_for_frontend_testing; 
if !defined $obj->ZONEMASTER_number_of_processes_for_batch_testing; 
```
Maybe you can add it to #752 ?

mattias-p:
Yeah, and I forgot the unit tests for these cases too. I'll add those to #752 as well.

PR #752, opened 2021-04-27
Use uniform accessors

mattias-p:
## Context

This PR is a follow-up to #749.

## Scope

Included:
* Cleaning up after #749.
* Documentation is added for the new getters from #749.
* Documentation for the database connection related properties is updated for consistency.
* Relevant pieces of cleaned up documentation is integrated into the backend config format specification.
* SQL-queries adjacent to uses of the new property getters are refactored to use the ?-syntax instead of string concatenation. To facilitate a few other refactorings are also included.

Excluded:
* Functional changes.

## Changes

Security:
* In a couple of places another layer of security is added by using SQL-injection resistant ? syntax. This also makes the code easier to review with regard to security.

Documentation:
* The backend config format specification gets a tiny improvement by adopting a wording from the getter POD.
* The POD for the uniform getters is made very light for easier maintenance and higher reliability by not duplicating information provided in the format specification. The idea is to make the format specification a single source of truth. Implementation details are still kept in the getter POD, e.g. the form in which the property values are returned.

Clean-up:
* The free-form getters that received new replacements in #749 are removed. All calls to them are replaced with calls to the new getters.

## How to test this PR

This PR contains no functional changes. It should be enough that the automatic tests pass.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -297,9 +297,10 @@ Default value: `0`.
 
 ### age_reuse_previous_test
 
-Positive integer.
-
+A positive integer.
 The shelf life of a test in seconds after its creation.
+Default value: `600`.
+
 If a new test is requested for the same zone name and parameters within the
 shelf life of a previous test result, that test result is reused.
 Otherwise a new test request is enqueued.
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -298,9 +298,11 @@ Default value: `0`.
 ### age_reuse_previous_test
 
 Positive integer.
-How old (in seconds) a previous test of the same zone name and parameters can be
-when it is reused instead of starting a new test.
-Default value: `600`.
+
+The shelf life of a test in seconds after its creation.
+If a new test is requested for the same zone name and parameters within the
+shelf life of a previous test result, that test result is reused.
+Otherwise a new test request is enqueued.
 
 Internally the value is converted to whole minutes.
 If the conversion results in zero minutes, then the default value is used.
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -298,8 +298,8 @@ Default value: `0`.
 ### age_reuse_previous_test
 
 Positive integer.
-How old (in seconds) a previous test of the same zone name and parameters must
-be before we start a new test.
+How old (in seconds) a previous test of the same zone name and parameters can be
+when it is reused instead of starting a new test.
 Default value: `600`.
 
 Internally the value is converted to whole minutes.
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -463,25 +463,11 @@ Returns one of C<"SQLite">, C<"PostgreSQL"> or C<"MySQL">.
 Get the value of L<DB.polling_interval|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#polling_interval>.
 
 
-=head2 MaxZonemasterExecutionTime
+=head2 ZONEMASTER_max_zonemaster_execution_time
 
-=head3 INPUT
+Get the value of L<ZONEMASTER.max_zonemaster_execution_time|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
 
-'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file. See
-L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
-
-=head3 RETURNS
-
-Integer (number of seconds).
-
-=cut
-
-sub MaxZonemasterExecutionTime {
-    my ($self) = @_;
-
-    return $self->ZONEMASTER_max_zonemaster_execution_time;
-}
-
+Returns an integer (number of seconds).
 
 
 =head2 ZONEMASTER_number_of_processes_for_frontend_testing
@@ -500,44 +486,20 @@ L<ZONEMASTER.number_of_processes_for_batch_testing|https://github.com/zonemaster
 Returns an integer.
 
 
-=head2 lock_on_queue
+=head2 ZONEMASTER_lock_on_queue
 
-=head3 INPUT
+Get the value of
+L<ZONEMASTER.lock_on_queue|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#lock_on_queue>.
 
-'lock_on_queue' from [ZONEMASTER] section in ini file. See
-L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#lock_on_queue>.
-
-=head3 RETURNS
-
-Integer.
-
-=cut
-
-sub lock_on_queue {
-    my ($self) = @_;
-
-    return $self->ZONEMASTER_lock_on_queue;
-}
+Returns an integer.
 
 
-=head2 maximal_number_of_retries
+=head2 ZONEMASTER_maximal_number_of_retries
 
-=head3 INPUT
+Get the value of
+L<ZONEMASTER.maximal_number_of_retries|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#maximal_number_of_retries>.
 
-'maximal_number_of_retries' from [ZONEMASTER] section in ini file. See
-L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#maximal_number_of_retries>.
-
-=head3 RETURNS
-
-A scalar value of the number of retries.
-
-=cut
-
-sub maximal_number_of_retries {
-    my ($self) = @_;
-
-    return $self->ZONEMASTER_maximal_number_of_retries;
-}
+Returns an integer.
 
 
 =head2 ZONEMASTER_age_reuse_previous_test
@@ -643,7 +605,7 @@ The values of the following attributes affect the construction of the returned o
 
 =over
 
-=item MaxZonemasterExecutionTime
+=item ZONEMASTER_max_zonemaster_execution_time
 
 =item ZONEMASTER_number_of_processes_for_batch_testing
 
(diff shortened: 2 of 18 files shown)


docs/Configuration.md
```diff
@@ -292,8 +292,8 @@ Default value: `0`.
 ### age_reuse_previous_test
 
 Positive integer.
-How old (in seconds) a previous test of the same zone name and parameters must
-be before we start a new test.
+How old (in seconds) a previous test of the same zone name and parameters can be
+when it is reused instead of starting a new test.
```
pnax, line 296: Not sure if `can` conveys the right meaning. From my understanding if the time limit is reached, then a new test is run, otherwise the old test's results are displayed. And the usage of _can_ here make it sounds like there is a possibility to run the test even if the time limit is not reached.

I understand that there might be something unclear with the current sentence. 
If you feel we need to rewrite this sentence, I tried to formulate it another way (but I'm still not satisfied)
-> Time (in seconds) to wait before running a new test instead of reusing the results of a previous test of the same zone name and parameter.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -370,134 +348,67 @@ sub ListLanguageTags {
...
-}
-
-
-=head2 MaxZonemasterExecutionTime
-
-=head3 INPUT
 
-'max_zonemaster_execution_time' from [ZONEMASTER] section in ini file. See
-L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
+=head2 DB_engine
```
pnax, line 352: Can you move all these POD (DB_engine ... ZONEMASTER_age_reuse_previous_test) above line 266 with the other ones ? I think it will make more sense to have them all together (however I should say that I don't know much on POD, so if there is a reason to keep them here, I can live with that :) )

lib/Zonemaster/Backend/Config.pm
```diff
@@ -370,134 +348,67 @@ sub ListLanguageTags {
...
 
-=head2 NumberOfProcessesForFrontendTesting
+=head2 ZONEMASTER_max_zonemaster_execution_time
 
-=head3 INPUT
+Get the value of L<ZONEMASTER.max_zonemaster_execution_time|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#max_zonemaster_execution_time>.
 
-'number_of_processes_for_frontend_testing' from [ZONEMASTER] section in ini file. See
-L<https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#number_of_processes_for_frontend_testing>.
+Returns an integer (number of seconds).
```
pnax, line 368: Why specifying what kind of value is returned here and not for all the other one ?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -336,27 +336,38 @@ sub add_batch_job {
...
-            WHERE test_start_time < DATE_SUB(NOW(), INTERVAL ".$self->config->MaxZonemasterExecutionTime()." SECOND) 
-            AND nb_retries <= ".$self->config->maximal_number_of_retries()." 
+            FROM test_results
+            WHERE test_start_time < DATE_SUB(NOW(), INTERVAL ? SECOND)
+            AND nb_retries <= ?
             AND progress > 0
             AND progress < 100
-            AND queue = ".$self->config->lock_on_queue();
+            AND queue = ?" );
+        $sth->execute(    #
```
pnax, line 351: Is the trailing empty comment mandatory ?
(this question applies to all following `execute()` calls)

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -371,28 +372,39 @@ sub add_batch_job {
...
+
+    if ( $self->config->ZONEMASTER_lock_on_queue ) {
+        my $sth = $self->dbh->prepare( "
+            SELECT hash_id, results, nb_retries
+            FROM test_results
+            WHERE test_start_time < DATETIME('now', ?)
+            AND nb_retries <= ?
+            AND progress > 0
+            AND progress < 100
+            AND queue=?" );
```
pnax, line 386: Can you add spaces around `=` ?

pnax (commented):
This looks great ! Thanks
I have few remarks and questions to make it even greater ;)

docs/Configuration.md
mattias-p: When I tried to look for a real life analogy to what we're trying to express and the closes I came was "shelf life". I'll come up with a new wording based on that.

lib/Zonemaster/Backend/Config.pm
mattias-p: POD doesn't care about order. There are coverage tools for verifying that every named subroutine has a corresponding POD heading but otherwise you may do as you please.

For this PR I've mainly tried to do something at least half-way reasonable regarding the placement of the POD while also keeping down the size of the diff. But I'll take another look.

lib/Zonemaster/Backend/DB/MySQL.pm
mattias-p: I put it there to coax perltidy into doing something reasonable. If I leave it out perltidy just pulls all the arguments up on the same line making it harder to read :)

lib/Zonemaster/Backend/DB/SQLite.pm
mattias-p: Sure!

lib/Zonemaster/Backend/Config.pm
mattias-p: I'll remove it for consistency.

mattias-p:
> This looks great ! Thanks
> I have few remarks and questions to make it even greater ;)

Of course! Great is great but greater is greater :)

lib/Zonemaster/Backend/Config.pm
pnax: This was just to have the POD close to the code (if you prefer reading the code file directly). Thanks for the update

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: I see. Thanks for the explanation. And I do agree that it is better to have them on several lines.

docs/Configuration.md
```diff
@@ -292,9 +292,11 @@ Default value: `0`.
...
 
 Positive integer.
-How old (in seconds) a previous test of the same zone name and parameters must
-be before we start a new test.
-Default value: `600`.
+
+The shelf life of a test in seconds after its creation.
+If a new test is requested for the same zone name and parameters within the
+shelf life of a previous test result, that test result is reused.
+Otherwise a new test request is enqueued.
```
pnax, line 299: I think you removed the default value line by mistake.
(I like the rewording btw)

pnax (commented):
Thanks for all the updates. Sadly a small mistake slipped in ^^'

docs/Configuration.md
mattias-p: Good spotting! I added it back.
I also also made a small edit to make age_reuse_previous_test to make its formatting more like the other simple properties.

> (I like the rewording btw)

I'm glad to hear that :)

pnax (approved):

lib/Zonemaster/Backend/Config.pm
mattias-p: I completely agree. The only reason I didn't make it so in the first place was a practical one. I had a large set of changes that I was trying to rebase into a series of PRs, so having smaller diffs made a big difference at the time. But I'm glad I got to set it straight.

mattias-p:
@pnax Thanks for reviewing.

@matsduf Could you review this too?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -216,72 +224,109 @@ sub check_db {
...
 }
 
-sub BackendDBType {
-    my ($self) = @_;
-    return $self->DB_engine;
-}
+
+=head2 DB_engine
+
+Get the value of L<DB.engine|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#polling_interval>.
```
matsduf, line 230: It should not be "polling_interval" here.

lib/Zonemaster/Backend/DB/MySQL.pm
matsduf: What a trick! I will remember that.

matsduf (changes requested):

matsduf:
Just one small change.

pnax (changes requested):
While going through the code, I realize that the way some deprecated values are handled is not right. I am referring to `number_of_processes_for_frontend_testing` and `number_of_processes_for_batch_testing`. The code on line 183 and 192 of Config.pm will never be used because we set a default value to 20. So the value is always defined at this stage and will never be overridden by the config value.

If we care about stability for deprecated values, I think this need to be reworked.

mattias-p:
Thanks for your reviews!

I must say I didn't see the default value problem coming. Getting the order right in this kind of code is surprisingly tricky. Unfortunately I can't fix it the way I want to without conflicting with #753, and I'd rather merge that one first.

After #753 is merged I'll rebase this one and fix the problems you guys found. I'll also add the unit tests I for `number_of_processes_for_{frontend,batch}_testing` that I forgot to add earlier.

pnax:
Yes I agree that this looks tricky. I discovered this while trying to look how to properly deprecate the `number_of_processes_for_*` keys. And I said to myself, well let's see if @mattias-p can come up with a solution ;-p

mattias-p:
Yeah, having test coverage of all the cases is key, though. Once you get the tests to pass you also want to make it look easy. It's so much more satisfying when the tricky stuff shows up in more glamorous places. Configuration file validation just doesn't get enough hype these days.

But I rebased the commits and fixed the problems. I ended up assigning the private fields multiple times going from least to most prioritized source instead going the other way around and using guards. I think it turned out pretty good. I just hope I didn't forget more tests.

lib/Zonemaster/Backend/Config.pm
mattias-p: Good catch! I fixed it.

matsduf (approved):
This is a large and complex PR. It looks fine, but will have to test it thoroughly.

pnax (approved):
Well done ! You came up with a very nice solution :)

PR #753, opened 2021-04-27
Validate structure

mattias-p:
## Context

This PR is a concrete step towards #685.

## Scope

This PR concerns itself with what config properties are present and absent. It does not concern itself with the values of properties. Neither does it concern itself with the properties profile sections since those are more dynamic in nature.

## Changes

This PR updates Zonemaster::Backend::Config->parse() to validate that all present properties are recognized and that all required properties are present.

This PR also updates Zonemaster::Backend::Config->parse() to defer all deprecation warnings so that they are only emitted if all strict validations pass. This should make future maintenance easier since it gives us more room to rearrange the validation code without affecting the behavior with regard to deprecation warnings.

## How to test this PR

New unit tests are added for the relevant cases, so additional manual testing should not be necessary.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -76,7 +76,16 @@ Emits a log warning with a deprecation message for each deprecated property that
 is present.
 
 Throws an exception if the given configuration contains errors.
-Unrecognized sections and properties are silently ignored.
+
+In a valid config file:
+
+=over 4
+
+=item
+
+all sections and properties are recognized.
+
+=back
 
 =cut
 
@@ -88,9 +97,26 @@ sub parse {
     my $ini = Config::IniFiles->new( -file => \$text )
       or die "Failed to parse config: " . join( '; ', @Config::IniFiles::errors ) . "\n";
 
+    my $get_and_clear = sub {    # Read and clear a property from a Config::IniFiles object.
+        my ( $section, $param ) = @_;
+        my $value = $ini->val( $section, $param );
+        $ini->delval( $section, $param );
+        return $value;
+    };
+
+    # Validate section names
+    {
+        my %sections = map { $_ => 1 } ( 'DB', 'MYSQL', 'POSTGRESQL', 'SQLITE', 'LANGUAGE', 'PUBLIC PROFILES', 'PRIVATE PROFILES', 'ZONEMASTER' );
+        for my $section ( $ini->Sections ) {
+            if ( !exists $sections{$section} ) {
+                die "config: unrecognized section: $section\n";
+            }
+        }
+    }
+
     # Validate, normalize, and apply default values
     {
-        my $engine = $ini->val( 'DB', 'engine' );
+        my $engine = $get_and_clear->( 'DB', 'engine' );
         eval {
             $engine = $obj->check_db($engine);
         };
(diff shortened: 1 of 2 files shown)


pnax (dismissed):
This looks good to me

mattias-p:
@pnax Thanks for reviewing.

@matsduf Could you review this too?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -100,25 +130,25 @@ sub parse {
...
+    $obj->{_SQLITE_database_file}                                = $get_and_clear->( 'SQLITE',     'database_file' )                            // undef;
+    $obj->{_ZONEMASTER_max_zonemaster_execution_time}            = $get_and_clear->( 'ZONEMASTER', 'max_zonemaster_execution_time' )            // '600';
+    $obj->{_ZONEMASTER_maximal_number_of_retries}                = $get_and_clear->( 'ZONEMASTER', 'maximal_number_of_retries' )                // '0';
+    $obj->{_ZONEMASTER_number_of_processes_for_frontend_testing} = $get_and_clear->( 'ZONEMASTER', 'number_of_processes_for_frontend_testing' ) // '20';
+    $obj->{_ZONEMASTER_number_of_processes_for_batch_testing}    = $get_and_clear->( 'ZONEMASTER', 'number_of_processes_for_batch_testing' )    // '20';
+    $obj->{_ZONEMASTER_lock_on_queue}                            = $get_and_clear->( 'ZONEMASTER', 'lock_on_queue' )                            // '0';
+    $obj->{_ZONEMASTER_age_reuse_previous_test}                  = $get_and_clear->( 'ZONEMASTER', 'age_reuse_previous_test' )                  // '600';
```
matsduf, line 148: When there are default values in Config.pm the default configuration file should not have the same value.
```
[ZONEMASTER]
max_zonemaster_execution_time            = 300
#age_reuse_previous_test=600
```
The max_zonemaster_execution_time is different from the Config.pm. I suggest that the configuration file has the same values, but the key is commented, or that they are just removed from the configuration file.

matsduf (changes requested):
Great work, but one comment.

lib/Zonemaster/Backend/Config.pm
mattias-p: The only place I can find the value 300 is in the Travis backend config files. Do you think those must use the default value too?

matsduf: 1. If the default values are set in Config.pm, the keys in the default configuration file should be commented as `age_reuse_previous_test` is.
2. In https://github.com/zonemaster/zonemaster-backend/blob/develop/share/backend_config.ini I find the the following:
```
[ZONEMASTER]
max_zonemaster_execution_time            = 300
#age_reuse_previous_test=600
```

share/backend_config.ini
```diff
@@ -21,10 +21,10 @@ database = zonemaster
 database_file = /var/lib/zonemaster/db.sqlite
 
 [ZONEMASTER]
-max_zonemaster_execution_time            = 300
+#max_zonemaster_execution_time           = 600
 number_of_processes_for_frontend_testing = 20
 number_of_processes_for_batch_testing    = 20
-#age_reuse_previous_test=600
+#age_reuse_previous_test                 = 600
```
matsduf, line 27: The `number_of_processes_*` keys have also default values in config.pm. I suggest them to be commented too:
```
# number_of_processes_for_frontend_testing = 20
# number_of_processes_for_batch_testing    = 20
```

matsduf (approved):

Issue #496, opened 2019-04-10
Add database_port in backend_config.ini


pamasse:
Make it possible to modify the default database connection port

pnax:
Resolved by #755

PR #755, opened 2021-04-29
Configuration of the database port

pnax:
## Context

Adds a new configuration key to specify the database port for MySQL and PostgreSQL as suggested in #496.
~This PR needs #753 to be merged first, and then rebased on top of develop.~

## Changes

This adds 2 new configuration entries, one in the MYSQL section and the other in the POSTGRESQL section. Each entry makes it possible to specify the port number on which the database server is listening. If the `host` key is not present, then the behavior is the same as before, each DBI uses its default port value.

With MySQL, if you want to talk with the database server on the local machine with another port number, you need to define the loopback IP address in the MYSQL.host (otherwise using `localhost` will result in using a UNIX socket as explained in the [DBD::mysql documentation](https://metacpan.org/pod/DBD::mysql#connect))

## How to test this PR

1. modify the configuration of your database to listen on a specific port, and restart your database server
2. if you are testing against MySQL with a local database server, update the `MYSQL.host` value to `127.0.0.1`
3. if you are testing with a remote database server, make sure you have allowed remote access to the database
4. add or update the `MYSQL.port` or `POSTGRESQL.port` config key/value accordingly to your setup
5. run a test with `zmb`

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -70,7 +70,7 @@ Default value: `0.5`.
 
 ## MYSQL section
 
-Available keys : `host`, `user`, `password`, `database`.
+Available keys : `host`, `port`, `user`, `password`, `database`.
 
 ### host
 
@@ -79,6 +79,13 @@ An [LDH domain name] or IP address.
 The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
+
+### port
+
+The port the MySQL server is listening on. If the server is on the local
+machine, then the [MYSQL.host] value must contain the loopback IP
+address (127.0.0.1).
+Default value: `3306`.
 
 ### user
 
@@ -111,7 +118,7 @@ If this property is unspecified, the value of [DB.database_name] is used instead
 
 ## POSTGRESQL section
 
-Available keys : `host`, `user`, `password`, `database`.
+Available keys : `host`, `port`, `user`, `password`, `database`.
 
 ### host
 
@@ -120,6 +127,11 @@ An [LDH domain name] or IP address.
 The host name of the machine on which the PostgreSQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
+
+### port
+
+The port the PostgreSQL server is listening on.
+Default value: `5432`.
 
 ### user
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -85,7 +85,7 @@ If this property is unspecified, the value of [DB.database_host] is used instead
 The port the MySQL server is listening on.
 Default value: `3306`.
 
-If [MYSQL.host] is set to `localhost` (but not `127.0.0.1` nor `::1`),
+If [MYSQL.host] is set to `localhost` (but neither `127.0.0.1` nor `::1`),
 then the value of the [MYSQL.port] property is discarded as the driver
 connects using a UNIX socket (see the [DBD::mysql documentation]).
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -80,12 +80,18 @@ The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
+If this property is set to `localhost`, then the [MYSQL.port] property
+is discarded as the driver connects using a UNIX socket (see the
+[DBD::mysql documentation]).
+
 ### port
 
-The port the MySQL server is listening on. If the server is on the local
-machine, then the [MYSQL.host] value must contain the loopback IP
-address (127.0.0.1).
+The port the MySQL server is listening on.
 Default value: `3306`.
+
+If [MYSQL.host] is set to `localhost`, then this property is discarded as
+the driver connects using a UNIX socket (see the [DBD::mysql
+documentation]).
 
 ### user
 
@@ -359,6 +365,7 @@ Otherwise a new test request is enqueued.
 [DB.database_name]:                   #database_name
 [DB.password]:                        #password
 [DB.user]:                            #user
+[DBD::mysql documentation]:           https://metacpan.org/pod/DBD::mysql#host
 [Default JSON profile file]:          https://github.com/zonemaster/zonemaster-engine/blob/master/share/profile.json
 [File format]:                        https://metacpan.org/pod/Config::IniFiles#FILE-FORMAT
 [ISO 3166-1 alpha-2]:                 https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -80,18 +80,14 @@ The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
-If this property is set to `localhost`, then the [MYSQL.port] property
-is discarded as the driver connects using a UNIX socket (see the
-[DBD::mysql documentation]).
-
 ### port
 
 The port the MySQL server is listening on.
 Default value: `3306`.
 
-If [MYSQL.host] is set to `localhost`, then this property is discarded as
-the driver connects using a UNIX socket (see the [DBD::mysql
-documentation]).
+If [MYSQL.host] is set to `localhost` (but not `127.0.0.1` nor `::1`),
+then the value of the [MYSQL.port] property is discarded as the driver
+connects using a UNIX socket (see the [DBD::mysql documentation]).
 
 ### user
 
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -38,8 +38,8 @@ sub dbh {
     else {
         my $database = $self->config->MYSQL_database;
         my $host     = $self->config->MYSQL_host;
-        my $user     = $self->config->MYSQL_user();
-        my $password = $self->config->MYSQL_password();
+        my $user     = $self->config->MYSQL_user;
+        my $password = $self->config->MYSQL_password;
 
         $log->notice( "Connecting to MySQL: database=$database host=$host user=$user" ) if $log->is_notice;
 
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -38,8 +38,11 @@ sub dbh {
     else {
         my $database = $self->config->MYSQL_database;
         my $host     = $self->config->MYSQL_host;
+        my $port     = $self->config->MYSQL_port;
         my $user     = $self->config->MYSQL_user;
         my $password = $self->config->MYSQL_password;
+
+        my $data_source_name = "DBI:mysql:database=$database;host=$host;port=$port";
 
         $log->notice( "Connecting to MySQL: database=$database host=$host user=$user" ) if $log->is_notice;
 
@@ -48,7 +51,7 @@ sub dbh {
         }
 
         $dbh = DBI->connect(
-            "DBI:mysql:database=$database;host=$host",
+            $data_source_name,
             $user,
             $password,
             {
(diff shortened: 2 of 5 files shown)


docs/Configuration.md
```diff
@@ -64,14 +64,20 @@ Default value: `0.5`.
...
 
 The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
+### port
+
+The port the MySQL server is listening on. If the server is on the local
+machine, then the [MYSQL.host] value must contain the loopback IP
+address (127.0.0.1).
```
matsduf, line 79:
```
address, 127.0.0.1 or ::1, as applicable.
```

mattias-p:
I think this looks really good! I have a couple comments regarding default values though.

In principle I like it better when the default value can be explicitly set. In this case though, there's no way to explicitly specify `undef` in the config file. I guess it's less important in config files than in bona fide API. And in this case it's well known what the default port is for each of the databases. But in principle, the idea is that it's easier to always specify all the properties when using things programmatically, and you just switch out the values depending on what you want to achieve. In the alternative, i.e. if there are properties with default values that can't be specified explicitly you also need to care for which properties to exclude, which is often annoying.

In any case there ought to be a unit test to verify that MYSQL_port and POSTGRESQL_port keep returning reasonable values when their config properties are unspecified.

Also the if-statements in the DB adapters wouldn't be needed if the default values were set to 3306 and 5432.

pnax:
I moved it to 2020.2 to let us time to discuss and properly implement this new property.

I didn't set the default port value because each database driver says it handles it already :

> Should the MySQL server be running on a non-standard port number ...
https://metacpan.org/pod/DBD::mysql#port

> If a parameter is not given, the connect() method will first look for specific environment variables, and then fall back to hard-coded defaults:
> port         PGPORT                  5432
https://metacpan.org/pod/DBD::Pg#connect

I didn't want to hard code something already handle by the driver.

But from your comment I do understand that in our case we can hard code those default values and this will also make things easier.

> In any case there ought to be a unit test to verify that MYSQL_port and POSTGRESQL_port keep returning reasonable values when their config properties are unspecified.
> Also the if-statements in the DB adapters wouldn't be needed if the default values were set to 3306 and 5432.

If we have hard coded default, I totally agree with these 2 sentences.

So the question is: do we explicitly set the default database port in the code ? To be honest, I really don't know what is best practice in our case.

matsduf:
> So the question is: do we explicitly set the default database port in the code ? To be honest, I really don't know what is best practice in our case.

I do not think we should. 

Since there is a default in client code, we should accept that as default. I am not sure that we should specify host either. If I read the documentation correctly, setting it to "localhost" means to Mysql using Unix socket, and then port is not relevant. If we leave host out, the same thing is true. When it comes to Postgresql, it was not clear if specifying "localhost" means going over to TCP, else if unset Unix socket will be used.

pnax:
>  If I read the documentation correctly, setting it to "localhost" means to Mysql using Unix socket, and then port is not relevant

This is what I understand as well. I also tried and indeed if the host is set to `localhost` then a Unix socket is used (even if a port number is specified).

> When it comes to Postgresql, it was not clear if specifying "localhost" means going over to TCP, else if unset Unix socket will be used.

Doing some tests (using commit 5c0c1a2), it looks like it depends.

| posgresql | backend_config.ini | using TCP * |
| --- | --- | :---: |
| `port = 5432` | `POSTGRESQL.host = localhost` | no |
| `port = 5432` | `POSTGRESQL.host = localhost`<br/>`POSTGRESQL.port = 5432` | no |
| `port = 5435` | `POSTGRESQL.host = localhost`<br/>`POSTGRESQL.port = 5435` | yes |

_* I can see traffic with tcpdump_

> I am not sure that we should specify host either.

This makes sense, either we use the default values from the driver or we set our own default values.

mattias-p:
We could use a sentinel value to specify that the client library default port should be used. You can't have anything listen on port 0, right? So we could use `0` should be up for grabs but it's probably more obvious to use something different like `default`. That way we can provide a nicer interface than both DBD::mysql and DBD::Pg, while still respecting the single definition of the database default port numbers. We'll have to deal with the conditionals in the DSN construction code though.

But how do we document which port is connected to when the port property is omitted unless we set our own default?

pnax:
I've rebased on top of develop and I choose to hard code the default port values. One reason I made this choice comes from #757 where we are testing properties based on the MySQL and PostgreSQL allowed values. Another is for code clarity, as it is simpler than having to test all cases before the DSN construction.

pnax:
I marked this PR as ready for review, however, depending on which PR is merged first between this one and #757, more work might be needed here.

docs/Configuration.md
```diff
@@ -64,14 +64,21 @@ Default value: `0.5`.
...
+Default value: `3306`.
```
mattias-p, line 80: If MYSQL.host = localhost, then MYSQL.port is not being used because the driver connects using a UNIX socket. I think this should be clear from the documentation.

Maybe we should also do something in the implementation in this case? Like making parse() fail or emit a warning. I'm leaning towards having parse() fail, but I'm not sure.

I guess the special `localhost` value should be documented in MYSQL.host as well.

pnax: I've added some documentation, I'll add some code to catch this behavior. I think I'll go with a warning as this is not an error per se.

t/config.t
```diff
@@ -88,6 +88,38 @@ subtest 'Everything but NoWarnings' => sub {
...
+
+            [MYSQL]
+            host     = ::1
+            user     = mysql_user
+            password = mysql_password
+            database = mysql_database
+        };
+        my $config = Zonemaster::Backend::Config->parse( $text );
+        is $config->MYSQL_host, '::1', 'set: MYSQL.host';
+        is $config->MYSQL_port, 3306,  'default: MYSQL.port';
```
mattias-p, line 104: Did you consider extending the first two `lives_and` clauses instead of creating new ones?

pnax: Not at all. I made this choice because this is specific to different database engine, but I'm fine with checking all the default values in one place.

docs/Configuration.md
```diff
@@ -64,14 +64,27 @@ Default value: `0.5`.
...
+If this property is set to `localhost`, then the [MYSQL.port] property
+is discarded as the driver connects using a UNIX socket (see the
+[DBD::mysql documentation]).
+
+### port
+
+The port the MySQL server is listening on.
+Default value: `3306`.
+
+If [MYSQL.host] is set to `localhost`, then this property is discarded as
```
matsduf, line 84:
```
If [MYSQL.host] is set to `localhost` (but not `127.0.0.1` or  `::1`), then (...)
```

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -37,12 +37,15 @@ sub dbh {
     else {
         my $database = $self->config->MYSQL_database;
         my $host     = $self->config->MYSQL_host;
-        my $user     = $self->config->MYSQL_user();
-        my $password = $self->config->MYSQL_password();
+        my $port     = $self->config->MYSQL_port;
+        my $user     = $self->config->MYSQL_user;
+        my $password = $self->config->MYSQL_password;
+
+        my $data_source_name = "DBI:mysql:database=$database;host=$host;port=$port";
```
matsduf, line 44: Is there any risk of warning from DBI if port is set when host is 'localhost'?

t/config.t
mattias-p: I think it's simple and scalable to have all the defaults grouped together. I had a couple of design ideas when I wrote config.t but I guess I failed to communicate them.

When I had another look at config.t I realized a couple of other mistakes I made. I'll take another stab to improve it a little after you've merged this PR.

mattias-p (approved):

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: Do you mean, does the DBI emits warning in this case ? Not that I am aware of.

matsduf: We should test for that so that we do not get unwanted warnings.

mattias-p: If DBD::mysql starts emitting warnings those will be caught in config.t. It's already using Test::NoWarnings.

pnax:
I rebased on top of develop and added the validation of the input.

I realize that I wrote in the documentation that the MYSQL.port property is disregarded if MYSQL.host is set to «localhost». However, this does not mean that you can put garbage in MYSQL.port. The property will always be validated first. I don't think we need to document this case, but if you feel it's better, feel free to say so.

I also choose to validate using the `untaint_strictly_positive_int` method, it allows up to 5 digits (which is fine for a port number). However, if we think we'll reduce that value in the future, this might break things. Do you think I should use a different method ?

docs/Configuration.md
matsduf: Any comment here?

docs/Configuration.md
```diff
@@ -80,6 +80,19 @@ The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
+If this property is set to `localhost` (but not `127.0.0.1` nor `::1`),
+then the [MYSQL.port] property is discarded as the driver connects using
```
matsduf, line 84:
```
then the value of the [MYSQL.port] property is discarded as the driver connects using
```

matsduf: I think the text should be moved under "### port" instead.

mattias-p:
> I also choose to validate using the untaint_strictly_positive_int method, it allows up to 5 digits (which is fine for a port number). However, if we think we'll reduce that value in the future, this might break things. Do you think I should use a different method?

If you specify a port number of 65536 or higher it won't work anyway (unless MYSQL.host == localhost). Instead of a validation error you'll get a failed connection. If we go with untaint_strictly_positive_int now and change our minds later to lower the upper bound to 65536, that won't be a problem.

mattias-p:
> I realize that I wrote in the documentation that the MYSQL.port property is disregarded if MYSQL.host is set to «localhost». However, this does not mean that you can put garbage in MYSQL.port. The property will always be validated first. I don't think we need to document this case, but if you feel it's better, feel free to say so.

I think it's fine. The file needs to be syntactically valid before we regard any of the actual values. And the properties themselves are part of that syntax check.

docs/Configuration.md
pnax: Ok. I'll change to keep this message only in the MYSQL.port section.

docs/Configuration.md
pnax: Oops, I forgot to update it indeed.

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: Well, I cannot see any message in the daemons' logs. Should I look somewhere else to see if warning are raised by the driver in this case ?

docs/Configuration.md
```diff
@@ -80,6 +80,15 @@ The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
+### port
+
+The port the MySQL server is listening on.
+Default value: `3306`.
+
+If [MYSQL.host] is set to `localhost` (but not `127.0.0.1` nor `::1`),
```
matsduf, line 88: Maybe
```
(but neither `127.0.0.1` nor `::1`)
```

matsduf (approved):

lib/Zonemaster/Backend/DB/MySQL.pm
matsduf: Then it should be fine.

mattias-p (dismissed):

docs/Configuration.md
pnax: modified, please re-review.

matsduf (approved):

pnax:
## Testing v2021.2

LGTM

PR #756, opened 2021-05-03
Stop rounding test shelflife to whole minutes

mattias-p:
## Context

This PR is a follow-up to https://github.com/zonemaster/zonemaster-backend/pull/692#discussion_r551941117.

## Changes

The ZONEMASTER.age_reuse_previous_test property is specified in seconds, but it was rounded to whole minutes.

This PR removes rounding logic and relieves us from having to describe this normalization in the configuration documentation.

## How to test this PR

1. Set `ZONEMASTER.age_reuse_previous_test` to 30 in your backend config file.
2. (Re)start your RPCAPI daemon to make sure the configuration is in effect.
3. Run:
   ```sh
   zmtest null.example >/dev/null
   sleep 15
   zmtest null.example >/dev/null
   sleep 30
   zmtest null.example >/dev/null
   ```
4. Before the fix: All three runs result int the same testid.
   After the fix: The first two runs result in the same testid, but the third one results in a different one.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -305,9 +305,6 @@ If a new test is requested for the same zone name and parameters within the
 shelf life of a previous test result, that test result is reused.
 Otherwise a new test request is enqueued.
 
-Internally the value is converted to whole minutes.
-If the conversion results in zero minutes, then the default value is used.
-
 
 
 [DB.database_host]:                   #database_host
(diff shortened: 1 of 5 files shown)


mattias-p:
The test procedure described above can be automated after #758 is implemented.

matsduf (approved):

pnax (approved):

mattias-p:

PR #757, opened 2021-05-03
Validate simple config properties

mattias-p:
## Purpose

This PR is a step towards full validation of the backend configuration file.

## Context

This PR builds on top of #759. (Initially those changes were introduced here, but they were split out to shrink the scope of this PR.)

This PR is an incremental step towards fixing #685.

### Unrelated changes

This PR updates the validation of digests in fake DS info in RPCAPI requests. Previously the only allowed digest lengths were 40 and 64 hex digits. But we need to support SHA-384 digests as well, so this PR opens up for 96 digits as well.

### Configuration

This PR changes how you enter IPv6 addresses in the MYSQL.host property. Previously you'd enter `host = [::1]`, but now you enter `host = ::1`.

### Documentation

This PR clarifies the config format specification with more detailed validation rules regarding allowed values and maximum lengths. Please take a close look at them.

IMHO the most eye catching thing is that only ASCII domain names are allowed for the database host properties. I made it so mainly to avoid complicated validation and normalization rules. It's possible to convert every IDN label to an ASCII label so this i not a technical restriction. If we don't support IDN labels in the database host properties from the get-go we might get feature requests to support them later. It is possible to implement that feature later on within the framework introduced here and it can be done without breaking changes.

### Zonemaster::Backend::Validator

Validator primitives are added for all the simple properties in the Config module.

Some of the primitives are implemented using intersection between regexes. I haven't seen this technique used anywhere else. The aspects that appeal to me is that it's easy to implement, in some cases you can have two simpler regexes instead of two more complicated ones. What I don't like about it is that I don't know if it's possible to use the same technique with JSON::Validator, which limits the reusability of those expressions. If there is a DRY way to define validation primitives for both JSON and INI, I'd like to hear it.

### Zonemaster::Backend::Config

The generated property setter methods are extended so with validation.

### Unit tests

Unit tests are added to make sure that invalid values are caught for all of the simple properties.

## How to test this PR

New unit tests are added for the relevant cases, so additional manual testing should not be necessary.


lib/Zonemaster/Backend/Validator.pm
```diff
@@ -6,7 +6,84 @@ use strict;
...
+);
+
+Readonly my $IPV4_RE => qr/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\$/;
+Readonly my $IPV6_RE => qr/^([0-9A-Fa-f]{1,4}:[0-9A-Fa-f:]{1,}(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?)\$|([0-9A-Fa-f]{1,4}::[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})\$/;
+
+Readonly my $API_KEY_RE                 => qr/^[a-zA-Z0-9-_]{1,512}$/;
+Readonly my $CLIENT_ID_RE               => qr/^[a-zA-Z0-9-+~_.: ]{1,50}$/;
+Readonly my $CLIENT_VERSION_RE          => qr/^[a-zA-Z0-9-+~_.: ]{1,50}$/;
+Readonly my $DIGEST_RE                  => qr/^[A-Fa-f0-9]{40}\$|^[A-Fa-f0-9]{64}\$/;
+Readonly my $DOMAIN_NAME_RE1            => qr{^[A-Z0-9-_/]{1,253}[.]?$}i;
```
matsduf, line 51: Shouldn't it be
```
qr{^[.A-Z0-9-_/]{1,253}[.]?$}i
```

Isn't
```
qr{^[.a-zA-Z0-9-_/]{1,253}[.]?$}
```
cheaper?

mattias-p:
I sure did forget the dot, I'll look into that.

But both regexes should, theoretically, compile to the same state machine. Just to make sure I ran both of them (with ^ and $ removed) across the .se zone file a hundred times each. On average the first one was 0.6% faster. I have a hunch that's well within the margin of error.

docs/Configuration.md
```diff
@@ -97,24 +111,33 @@ Available keys : `host`, `user`, `password`, `database`.
...
+A string of US ASCII letter, digit and/or punctuation characters. Max length 100
+characters.
+
 The password of the configured user.
 
 If this property is unspecified, the value of [DB.password] is used instead.
 
 ### database
 
+A [PostgreSQL identifier].
```
matsduf, line 139: I suggest ASCII only.

mattias-p:
As @matsduf pointed out this PR is rather large. I'll look for a way to split it up.

docs/Configuration.md
```diff
@@ -253,13 +278,15 @@ The ZONEMASTER section has several keys :
 
 ### max_zonemaster_execution_time
 
-An integer.
+A positive integer. Max length 9 digits.
```
matsduf, line 281: I think we should limit it further to make testing easier. I do not think we need to support anything longer than 3600 s.

docs/Configuration.md
```diff
@@ -253,13 +278,15 @@ The ZONEMASTER section has several keys :
...
+
 Time in seconds before reporting an unfinished test as failed.
 Default value: `600`.
 
 ### maximal_number_of_retries
 
-An integer.
+A non-negative integer. Max length 9 digits.
```
matsduf, line 288: Until we get a request for larger number, I think 1 digit is enough.

docs/Configuration.md
```diff
@@ -269,7 +296,8 @@ Do not use it (keep the default value "0"), or use it with care.
 
 ### number_of_processes_for_frontend_testing
 
-A positive integer.
+A positive integer. Max length 9 digits.
```
matsduf, line 299: This is extreme. Can run that many processes? I think limiting to 4 digits gives enought room.

docs/Configuration.md
```diff
@@ -280,7 +308,8 @@ frontend, but is used in combination of
 
 ### number_of_processes_for_batch_testing
 
-An integer.
+A non-negative integer. Max length 9 digits.
```
matsduf, line 311: As for the other processes.

docs/Configuration.md
```diff
@@ -291,13 +320,15 @@ batch pool of tests, but is used in combination of
 
 ### lock_on_queue
 
-An integer.
+A non-negative integer. Max length 9 digits.
+
 A label to associate a test to a specific Test Agent.
 Default value: `0`.
```
matsduf, line 326: This is just a name, isn't it? Or will that create that many queues? The description is incomplete. And the maximum value is probably unnecssarily hign.

docs/Configuration.md
```diff
@@ -291,13 +320,15 @@ batch pool of tests, but is used in combination of
...
 
 ### age_reuse_previous_test
 
-A positive integer.
+A positive integer. Max length 9 digits.
+
 The shelf life of a test in seconds after its creation.
 Default value: `600`.
```
matsduf, line 333: That is too high. 5 digits will give room for extreme values. We get less to test.

matsduf (changes requested):
This is just a review of Configuration.md.

docs/Configuration.md
```diff
@@ -68,24 +70,35 @@ Available keys : `host`, `user`, `password`, `database`.
...
+
 The host name of the machine on which the MySQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
 ### user
 
+A [MariaDB unquoted identifier]. Max length [80 characters][MariaDB identifier
+max lengths].
+
```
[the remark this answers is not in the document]
mattias-p: That seems reasonable. I'll make that update.

docs/Configuration.md
```diff
@@ -68,24 +70,35 @@ Available keys : `host`, `user`, `password`, `database`.
...
 The name of the user with sufficient permission to access the database.
 
 If this property is unspecified, the value of [DB.user] is used instead.
 
 ### password
 
+A string of US ASCII letter, digit and/or punctuation characters. Max length
+1024 characters.
```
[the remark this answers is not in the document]
mattias-p: You're right. `$` isn't an ASCII punctuation character: https://en.wikipedia.org/wiki/ASCII#Character_set. I'll add "symbol characters" to the list of allowed character classes.

[OWASP](https://owasp.org/about/) has a [cheat sheet on password strength](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls). Let's look at it.
* It says that minimum length should be enforced. While I do agree with this statement I don't think it's our role to enforce that. We should allow the empty password.
* Maximum length should not be too low and recommend at least 64 characters. That sounds like a reasonable limit to me.
* No limitations should be made on allowed characters. They explicitly list whitespace and Unicode.

I also found [this](https://dev.to/mitchpommers/password-max-length-limits-are-dumb-but-we-need-them-1dpp) where people seem to agree that 100 characters is a good maximum.

Speaking for myself I feel disinclined to support Unicode characters. It feels like a can of worms. We should probably accept space though, so we don't invalidate passwords like "correct horse battery staple". When it comes to US ASCII symbols and punctuation it doesn't seem to me like we win much by being restrictive.

docs/Configuration.md
```diff
@@ -58,6 +58,9 @@ over this.
 
 ### polling_interval
 
+A positive decimal number. Max 9 and 6 digits in the integer and fraction
+components respectively.
```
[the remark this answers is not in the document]
mattias-p: I'd prefer 0.001 over 0.0001 for lower limit. Millisecond is a commonly used unit. Or 0.000001 since a microsecond is the granularity that Time::HiRes speaks of.

I don't see how we get less to test by limiting the range. We'll just be checking the extremevalue and something just outside of that no matter if its 10 or 10 million.

docs/Configuration.md
```diff
@@ -97,24 +111,33 @@ Available keys : `host`, `user`, `password`, `database`.
...
 
+An ASCII domain name.
+
 The host name of the machine on which the PostgreSQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
 ### user
 
+A [PostgreSQL identifier].
```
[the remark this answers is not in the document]
mattias-p: Agreed.

docs/Configuration.md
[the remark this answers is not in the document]
matsduf: The problem that I see is not the RE in the perl code, but how to specify how to enter the password into the configuration file. If you allow for space and have a password starting with space you probably have to put the password within some quoting such as "" or '' or ´´or something. And how will the library reading the configuration file handle "\"? The lost of excluding space, "\" and "quotes" is probably smaller the gain of having less complex documentation.

The default password today is not safer than empty password.

docs/Configuration.md
mattias-p: I'll take that :)

docs/Configuration.md
mattias-p: I thought of that as well. Having fewer ranges to choose from makes it easier on the implementation side though.

docs/Configuration.md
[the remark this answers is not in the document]
matsduf: Testing if it works with 10.000.000 seconds can take 116 days. Testing 10 seconds takes 10 seconds.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -11,6 +11,13 @@ use File::ShareDir qw[dist_file];
 use File::Slurp qw( read_file );
 use Log::Any qw( $log );
 use Readonly;
+use Zonemaster::Backend::Validator qw( :untaint );
+
+Readonly my %DB_MODULE_NAMES => (
+    sqlite     => 'SQLite',
+    mysql      => 'MySQL',
+    postgresql => 'PostgreSQL',
```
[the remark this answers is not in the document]
mattias-p: Actually yes! I tried to get rid of it but we do use it to derive the name of the database adapter module.

docs/Configuration.md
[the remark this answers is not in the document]
matsduf: I am fine with 0.001 as the lower limit. I we get a request of a smaller value we can consider it then. I do not think we will get that. Anyone that wants to test that is probably capable of patching the code.

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
matsduf: A good reason! I suggest just a small comment, maybe right here in the code, why we have that.

matsduf:
> @matsduf Thanks for your comments! I like the idea of having less to test. But we don't do that by limiting the size of our ranges but by having fewer ranges to choose from. I'll make an update that limits all of the integer ranges to five digits. I think your suggestion to limit lock_on_queue to just a single digit is super cute, but I think I'll still make all the integer ranges max out at five digits.

I agree that having fewer ranges makes things less complicated. Five digits gives 28 hours. I suggest than limiting to four digits which will give three hours. That makes a huge difference when (if) testing the limits, and nobody will have such high numbers anyway.

docs/Configuration.md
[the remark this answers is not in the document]
mattias-p: Config::Inifiles supports quoting using the backslashes and "here document"-syntax. Leading whitespace in the value is trimmed. Even quoted leading whitespace is trimmed, at least that's how I read the documentation.

I'll add a link to the file format documentation and call out the whitespace rule and each of the quoting syntaxes.

mattias-p:
I've rebased the commits onto the current develop branch and addressed most review comments. Notably I have not addressed the numeric ranges.

docs/Configuration.md
```diff
@@ -68,24 +70,35 @@ Available keys : `host`, `user`, `password`, `database`.
...
+1024 characters.
+
 The password of the configured user.
 
 If this property is unspecified, the value of [DB.password] is used instead.
 
 ### database
 
+A [MariaDB unquoted identifier]. Max length [64 characters][MariaDB identifier
+max lengths].
```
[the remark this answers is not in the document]
mattias-p: Fixed.

docs/Configuration.md
```diff
@@ -97,24 +111,33 @@ Available keys : `host`, `user`, `password`, `database`.
...
+A [PostgreSQL identifier].
+
 The name of the user with sufficient permission to access the database.
 
 If this property is unspecified, the value of [DB.user] is used instead.
 
 ### password
 
+A string of US ASCII letter, digit and/or punctuation characters. Max length 100
+characters.
```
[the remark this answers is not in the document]
mattias-p: Fixed.

docs/Configuration.md
mattias-p: Fixed.

docs/Configuration.md
```diff
@@ -97,24 +115,34 @@ Available keys : `host`, `user`, `password`, `database`.
...
 
+An ASCII domain name.
+
 The host name of the machine on which the PostgreSQL server is running.
 
 If this property is unspecified, the value of [DB.database_host] is used instead.
 
 ### user
 
+A US ASCII-only [PostgreSQL identifier].
```
matsduf, line 126: Max length?

docs/Configuration.md
```diff
@@ -97,24 +115,34 @@ Available keys : `host`, `user`, `password`, `database`.
...
+The first character must be neither space nor `<`.
+Max length 100 characters.
+
 The password of the configured user.
 
 If this property is unspecified, the value of [DB.password] is used instead.
 
 ### database
 
+A US ASCII-only [PostgreSQL identifier].
```
matsduf, line 144: Max length?

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,83 @@ use warnings;
...
+
+# Up to 9 and 6 digits in the integer and fraction components respectively
+Readonly my $MILLIS_RE => qr/^(?:0|[1-9][0-9]{0,4})(?:[.][0-9]{1,3})?$/;
+
+# Printable ASCII but first character must not be space or '<'
+Readonly my $PASSWORD_RE => qr/^(?:[\x21-\x3b\x3d-\x7e][\x20-\x7e]{0,99})?$/;
+
+# Allow for 15 significant digits
+Readonly my $POSITIVE_INT_RE => qr/^[1-9][0-9]{0,4}$/;
+
```
matsduf, line 74: "Allow for up to 5 digits"

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,83 @@ use warnings;
...
+# See: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+Readonly my $POSTGRESQL_IDENT_RE    => qr/^[a-z_][a-z0-9_\$]{0,62}$/i;
 Readonly my $PROFILE_NAME_RE        => qr/^[a-z0-9]$|^[a-z0-9][a-z0-9_-]{0,30}[a-z0-9]$/i;
 Readonly my $RELAXED_DOMAIN_NAME_RE => qr/^[.]$|^.{2,254}$/;
 Readonly my $TEST_ID_RE             => qr/^[0-9a-f]{16}$/;
-Readonly my $USERNAME_RE            => qr/^[a-z0-9-.@]{1,50}$/i;
+
+# Allow for 15 significant digits
+Readonly my $UNSIGNED_INT_RE => qr/^(?:0|[1-9][0-9]{0,4})$/;
```
matsduf, line 85: 5 digits.

docs/Configuration.md
mattias-p: Fixed.

docs/Configuration.md
mattias-p: Fixed.

lib/Zonemaster/Backend/Validator.pm
mattias-p: Fixed.

lib/Zonemaster/Backend/Validator.pm
mattias-p: Fixed.

mattias-p:
Thanks for reviewing, @matsduf! I made an update addressing all your comments (and a couple of other things too).

matsduf:
@mattias-p, is this for v2021.1 or v2021.2?

mattias-p:
Let's keep it for next release.

matsduf (changes requested):

mattias-p:
I rebased onto current develop and resolved the conflicts. I also fixed a typo in one of the commits.

mattias-p:
@matsduf @pnax, Please review again.

matsduf:
I have already added some comments for clarifications of the documentation. Can you work those in?

mattias-p:
I though I addressed those, but apparently I forgot. But I did it now.

docs/Configuration.md
```diff
@@ -97,24 +115,34 @@ Available keys : `host`, `user`, `password`, `database`.
 
 ### host
 
+An [LDH domain name] extended with underscore (`_`) and slash (`/`).
```
matsduf, line 118: No, that is not a good idea. For zone names it is one thing to extend, but here we have something that should be a hostname that must have either an A or a AAAA record or both. That name should be an LDH name.

mattias-p: Okay, I'll add another data type.

matsduf (approved):

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,83 @@ use warnings;
...
+          untaint_postgresql_ident
+          untaint_unsigned_int
           )
     ],
 );
 
+# Does not check value ranges within the groups
 Readonly my $IPV4_RE => qr/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/;
+
+# Does not the length and number of the hex groups, nor the value ranges in the IPv4 groups
```
pnax, line 47: missing word -> `Does not check`

t/config.t
```diff
@@ -201,6 +201,335 @@ subtest 'Everything but NoWarnings' => sub {
...
+            [DB]
+            engine = SQLite
+            polling_interval = hourly
+
+            [SQLITE]
+            databse_file = /var/db/zonemaster.sqlite
+        };
+        Zonemaster::Backend::Config->parse( $text );
+    }
+    qr{DB\.polling_interval.*hourly}, 'die: Invalid DB.polling_inteval value';
```
pnax, line 215: typo `DB.polling_inteval` -> `DB.polling_interval`

pnax (commented):
I had a quick look on this PR. I spotted 2 typos so far.

lib/Zonemaster/Backend/Validator.pm
mattias-p: Fixed.

t/config.t
mattias-p: You do have a good eye. Also fixed.

mattias-p:
@matsduf @pnax Please review again.

matsduf (changes requested):

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,83 @@ use warnings;
...
-Readonly my $CLIENT_VERSION_RE      => qr/^[a-z0-9-+~_.: ]{1,50}$/i;
-Readonly my $DIGEST_RE              => qr/^[a-f0-9]{40}$|^[a-f0-9]{64}$/i;
-Readonly my $ENGINE_TYPE_RE         => qr/^(?:mysql|postgresql|sqlite)$/i;
-Readonly my $IPADDR_RE              => qr/^$|$IPV4_RE|$IPV6_RE/;
-Readonly my $JSONRPC_METHOD_RE      => qr/^[a-z0-9_-]*$/i;
-Readonly my $LANGUAGE_RE            => qr/^[a-z]{2}(_[A-Z]{2})?$/;
+Readonly my $API_KEY_RE                 => qr/^[a-z0-9-_]{1,512}$/i;
+Readonly my $CLIENT_ID_RE               => qr/^[a-z0-9-+~_.: ]{1,50}$/i;
+Readonly my $CLIENT_VERSION_RE          => qr/^[a-z0-9-+~_.: ]{1,50}$/i;
+Readonly my $DIGEST_RE                  => qr/^[a-f0-9]{40}$|^[a-f0-9]{64}$/i;
```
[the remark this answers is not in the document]
mattias-p: Thanks for looking it up! I'll add the 96 digit case as well.

mattias-p:
I force pushed to fix a spelling error in the last commit.

@matsduf @pnax Please review again.

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,83 @@ use warnings;
...
+Readonly my $MILLIS_RE => qr/^(?:0|[1-9][0-9]{0,4})(?:[.][0-9]{1,3})?$/;
+
+# Printable ASCII but first character must not be space or '<'
+Readonly my $PASSWORD_RE => qr/^(?:[\x21-\x3b\x3d-\x7e][\x20-\x7e]{0,99})?$/;
+
+# Allow for 5 significant digits
+Readonly my $POSITIVE_INT_RE => qr/^[1-9][0-9]{0,4}$/;
+
+# At least one non-zero digit
+Readonly my $POSITIVE_NUM_RE => qr/[^0.]/;
```
pnax, line 76: Not sure if this does what the comment says...

And looking at the code, this is only used in [`untaint_positive_millis`](https://github.com/zonemaster/zonemaster-backend/blob/bcfed7d49ab211589bcb6d3183d85e2352db575b/lib/Zonemaster/Backend/Validator.pm#L218..L221):

```
sub untaint_positive_millis {
    my ( $value ) = @_;
    return _untaint_pat( $value,  $MILLIS_RE, $POSITIVE_NUM_RE );
}
```
And here is the definition of `$MILLIS_RE` :
```
Readonly my $MILLIS_RE => qr/^(?:0|[1-9][0-9]{0,4})(?:[.][0-9]{1,3})?$/;
```

This looks like the `MILLIS_RE` regex is enough to match a positive decimal number between `0` and `99999.999`

(TBH I haven't tried with the unit tests in `t/validator.t` yet)

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -136,14 +195,64 @@ sub untaint_engine_type {
...
+}
+
+sub untaint_positive_int {
+    my ( $value ) = @_;
+    return _untaint_pat( $value,  $POSITIVE_INT_RE );
+}
+
+sub untaint_positive_millis {
+    my ( $value ) = @_;
+    return _untaint_pat( $value,  $MILLIS_RE, $POSITIVE_NUM_RE );
```
pnax, line 220: see my comment above on `POSITIVE_NUM_RE`, I'm wondering if this is needed here.

pnax (commented):
I still need to go over both test files

docs/Configuration.md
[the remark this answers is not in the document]
pnax: And the code has been updated to `5 and 3 digits`.

docs/Configuration.md
```diff
@@ -68,24 +74,36 @@ Available keys : `host`, `user`, `password`, `database`.
 
 ### host
 
+An [LDH domain name].
```
pnax, line 77: Shouldn't we also allow IP address ?

pnax (commented):
I realized that we restrict the database host to LDH, maybe we should allow IP addresses as well.

lib/Zonemaster/Backend/Validator.pm
pnax: Looking at the unit tests in `t/validator.t` I understand that we refuse null values. So we do need something more to avoid the `0`. Would it be possible to make it clearer in the method's name (for instance `untaint_strictly_positive_millis`) ?
And I think the `POSITIVE_NUM_RE` can be changed to `[1-9]` so that it reflects what the comment says.

docs/Configuration.md
pnax: After looking at the unit tests, and again at the regular expression, it looks like IP address are not rejected here. However, we might need to be clearer on the fact that we accept them, because the provided link is about restriction on domain names and doesn't say anything on IP address.

matsduf: IPv6 address is rejected, isn't it?

mattias-p: I'll update to "An LDH domain name or and IP address".

mattias-p:
@pnax It seems we have a disagreement on the properties of the number zero. I pose that zero is strictly non-positive. There's a thing called the trichotomy of real numbers where you split the real numbers into three disjoint sets - positive numbers, negative numbers and zero. In my experience this is a very useful way to think about things and I don't think we should mess with that.

lib/Zonemaster/Backend/Validator.pm
mattias-p: I replaced the POSITIVE_NUM_RE with a NON_ZERO_NUM that uses `[1-9]`, per your suggestion. I think the resulting code is much better. I used that in untaint_positive_int() as well together with UNSIGNED_INT_RE, and this made it possible to delete POSITIVE_INT_RE.

Regarding the membership of zero in the set of positive numbers, see my comment below.

mattias-p:
Maybe it's confusing that I've used both "unsigned" and "non-negative" for the same thing. I'll replace all "unsigned" with "non-negative".

lib/Zonemaster/Backend/Validator.pm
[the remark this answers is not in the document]
matsduf: In one way it is good that the digest is validated, but the problem is that the error message from RPCAPI is not shown to the user in the GUI, and even if it did it would be hard to understand it for most users. I removed the last character in a valid hash and sent it to zmb:

```
"error":{
  "message":"Invalid method parameter(s).",
  "code":"-32602",
  "data":"/ds_info/0/digest: String does not match (?^ui:^[a-f0-9]{40}$|^[a-f0-9]{64}$)."
  }
```
It the digest is corrupted you do not get any help. I think that data that users can paste in GUI should be handled by RPCAPI in such a way that a meaningful error message can be returned. Of course, the GUI must also be prepared to display it for the user.

Another possibility of error in digest is that some digit is replaced. In that case the test is executed and there is some meaningful error messages for the user.

pnax:
Well in school, I learned that zero is both a negative number and a positive number. I guess it depends on what we learned, so I feel more comfortable to explicitly say that we disallow the number zero in the input.

matsduf:
> Well in school, I learned that zero is both a negative number and a positive number. I guess it depends on what we learned, so I feel more comfortable to explicitly say that we disallow the number zero in the input.

In Wikipedia is says "0 is neither positive nor negative,[61] or both positive and negative." (https://en.wikipedia.org/wiki/0#Mathematics)

I think that supports the view that the text should be explicit about zero.

mattias-p:
@pnax Going through your comments on inconsistencies in the validation of integers I found that the `max_zonemaster_execution_time` and `number_of_processes_for_batch_testing` properties were using the wrong validation rules, so I fixed those.

mattias-p:
> I learned that zero is both a negative number and a positive number

and

> 0 is neither positive nor negative,[61] or both positive and negative.

I'm shocked! It's weird to me how often number theorists and algebraists have different definitions of basic terms. Myself being brought up on algebra.

I'll change all instances of "positive" to "strictly positive".

docs/API.md
```diff
@@ -287,12 +287,10 @@ A default installation will accept the following `language tags`:
 * `sv` or `sv_SE` for Swedish language.
 
 
-### Unsigned integer
+### Non-negative integer
 
 Basic data type: number (integer)
  
-An unsigned integer is either positive or zero.
```
pnax, line 294: Why removing this line ? It could be nice to keep it, so it's clear what we mean by `non-negative integer`.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -425,7 +425,7 @@ Returns an integer.
 Get the value of
 L<ZONEMASTER.number_of_processes_for_frontend_testing|https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md#number_of_processes_for_frontend_testing>.
 
-Returns a positive integer.
+Returns a strictly positive integer.
```
pnax, line 428: Some POD says `Returns an integer.`, should we update them as well ?

pnax (commented):
> It's weird to me how often number theorists and algebraists have different definitions of basic terms

Well I guess choices have been made throughout algebra's history depending on where you live in the world (and they didn't have Internet at that time to share their views and (dis)agreements)

docs/Configuration.md
pnax: A note on how IPv6 should be formatted for MySQL :

> When connecting to a MySQL Server with IPv6, a bracketed IPv6 address should be used.

https://metacpan.org/pod/DBD::mysql#connect

(sorry to make this PR heavier with my remarks :/ )
(and reading the PR title again, this is about validating simple properties, if you feel this isn't simple anymore, feel free to say so)

lib/Zonemaster/Backend/Validator.pm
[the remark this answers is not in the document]
matsduf: @mattias-p, any comments?

Also see #789

lib/Zonemaster/Backend/Config.pm
mattias-p: The idea I'm trying to implement is to let the configuration documentation specify the range of valid values in the configuration file, and to let the POD specify the data type of the return values. This way we can communicate the important details while keeping the redundancy across files to a minimum.

Since Perl is so loosely typed it's somewhat arbitrary how precisely the data type should be specified. I took another look at this. For scalar values the [perldata/Variable names](https://perldoc.perl.org/perldata#Variable-names) seems to provide a meaningful granularity without going into too much detail. I.e. distinguishing between numbers, strings and undefined.

I'll make an update to this effect.

docs/API.md
mattias-p: I was changing all "unsigned" to "non-negative" and all "positive" to "strictly positive". With those updates the statement became "A non-negative integer is either zero or strictly positive.", which felt to me like a non-statement. So I removed it. Do you think that that would be helpful to anyone? If you do I'll add it back in.

lib/Zonemaster/Backend/Config.pm
mattias-p: Oh, and I forgot. Some properties perform a transformation of the values parsed from the config file. These transformations also need to be documented. Especially for the complex configuration properties I believe these transformations are best described in relation to the the data type returned internally by the getter methods. So I think the documentation for these transformations should also be located in the POD for the getters.

docs/Configuration.md
mattias-p: At some point I was thinking about splitting this PR into controversial and non-controversial configuration properties so we could have the non-controversial ones merged and get a fresh PR for the controversial ones. But now it feels like we're getting close to the finish line.

Also from the start I actually expected this PR to be a rough ride because it contains so many details. And one of the worries I've been having since is that it would start forbidding reasonable edge cases. This IPv6 thing is a prime example of that. In hindsight I guess it was unwise cram all of these properties into a single PR.

I'll make the MySQL adapter bracket IPv6 addresses and see if I can get MariaDB to listen to IPv6.

mattias-p:
@pnax @matsduf Could you guys go back through you review comments here and "Resolve conversions" that you think are resolved? I believe I've addressed all of them except the MySQL IPv6 issue.

docs/Configuration.md
mattias-p: I just realized this is a breaking change. If you wanted IPv6 before you'd specify `host = [2001:db8::1]`, but after this PR you'll have update your config to `host = 2001:db8::1`.

mattias-p: I made the update and afterwards I was able to connect to MariaDB over IPv6. However I had to cheat - the IPv6 regex doesn't match `::1` so I had to hack around that.

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -7,36 +7,85 @@ use warnings;
...
+          untaint_strictly_positive_millis
           )
     ],
 );
 
+# Does not check value ranges within the groups
 Readonly my $IPV4_RE => qr/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/;
+
+# Does not check the length and number of the hex groups, nor the value ranges in the IPv4 groups
 Readonly my $IPV6_RE => qr/^([0-9a-f]{1,4}:[0-9a-f:]{1,}(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?)$|([0-9a-f]{1,4}::[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})$/i;
```
mattias-p, line 57: This regex does not accept `::1` which you're likely to put in the {MYSQL,POSTGRESQL}.host properties if you want to use IPv6 locally.

mattias-p: Fixed.

docs/API.md
mattias-p: I added the updated clarification.

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
matsduf: Have you considered a comment at `_normalize_engine_type` why the camel case forms are set here so that we do not have to think about it again? The one line that you states above would be enough.

lib/Zonemaster/Backend/Validator.pm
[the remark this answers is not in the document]
matsduf: > @pnax @matsduf Could you guys go back through your review comments above and "Resolve conversions" that you think are resolved? I believe I've addressed all of them except the MySQL IPv6 issue.

I have done so, and there are two places where I would like to see a reply from you.

mattias-p:
I believe I've addressed all comments and I made another self-review and tweaked some details.

@pnax @matsduf Please re-review.

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
mattias-p: I think you'll find out soon enough if you try to change it. You'll get complaints about "Can't locate Zonemaster/Backend/DB/mysql.pm". I think that's quite clear.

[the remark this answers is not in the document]
matsduf: Why are you against adding a comment, documentation, in the code?

lib/Zonemaster/Backend/Validator.pm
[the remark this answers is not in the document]
mattias-p: I agree that regexes don't make for great validation errors. But I don't think this PR is the place to fix it. Maybe you could create an issue for it?

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
mattias-p: Because in this particular case I think it adds more noise than value. More documentation isn't always better.

lib/Zonemaster/Backend/Validator.pm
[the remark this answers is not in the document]
matsduf: See #789

lib/Zonemaster/Backend/Config.pm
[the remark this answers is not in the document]
matsduf: I disagree that it would add to noise. It would clarify the choice. You are not the only one reading the code. With a comment I ask for this thread would never have been started.

docs/API.md
pnax: I think it's always better to clarify our choices, thanks for the update.

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -41,6 +42,11 @@ sub dbh {
         my $password = $self->config->MYSQL_password();
 
         $log->notice( "Connecting to MySQL: database=$database host=$host user=$user" ) if $log->is_notice;
+
+        if ( untaint_ipv6_address( $host ) ) {
+            $host = "[$host]";
+        }
```
pnax, line 48: Would it make sense to swap the `log` and the `if` statements so the value used for the `host` is caught in the log ?

t/validator.t
```diff
@@ -66,6 +55,31 @@ subtest 'Everything but NoWarnings' => sub {
...
+        is scalar untaint_ip_address( '2001:db8::' ),                             '2001:db8::',                             'accept: 2001:db8::';
+        is scalar untaint_ip_address( '2001:db8::/32' ),                          undef,                                    'reject: 2001:db8::/32';
+        is scalar untaint_ip_address( '2001:db8:ffff:ffff:ffff:ffff:ffff:ffff' ), '2001:db8:ffff:ffff:ffff:ffff:ffff:ffff', 'accept: 2001:db8:ffff:ffff:ffff:ffff:ffff:ffff';
+        is scalar untaint_ip_address( '2001:db8:ffff:ffff:ffff:ffff:ffff' ),      undef,                                    'reject: 2001:db8:ffff:ffff:ffff:ffff:ffff';
+        is scalar untaint_ip_address( '2001:db8::255.255.255.254' ),              '2001:db8::255.255.255.254',              'accept: 2001:db8::255.255.255.254';
+        is scalar untaint_ip_address( '2001:db8::255.255.255' ),                  undef,                                    'reject: 2001:db8::255.255.255';
+        is scalar untaint_ip_address( '::1' ),                                    '::1',                                    'accept: ::1';
```
pnax, line 68: Could you test with `:::1` (triple colons), I think this should fail, and it would be nice to have it tested, so we have a test case where the IPv6 regex succeed but the `Zonemaster::Engine::Net::IP::ip_is_ipv6()` fails.

lib/Zonemaster/Backend/DB/MySQL.pm
mattias-p: To me it makes more sense to report the plain format rather than how it's represented in the DBD::mysql DSN. That way IPv6 addresses are reported the same way when using MariadDB and PostgreSQL.

Another option would be to report the DSN and username in the same format for all database adapters.

t/validator.t
mattias-p: Done.

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: Ok that makes sense indeed.

pnax (commented):
This looks pretty good to me, I'll run some tests on my side before approving this PR.

pnax (approved):

mattias-p:
@matsduf Can you approve?

matsduf (approved):
This is a large PR with many changes. In the description is says "New unit tests are added for the relevant cases, so additional manual testing should not be necessary", but then the unit tests should be tested.

This PR must be tested before or at release testing.

docs/API.md
```diff
@@ -865,8 +865,8 @@ Example response:
 
 An object with the following properties:
 
-* `"offset"`: An *unsigned integer*, optional. (default: 0). Position of the first returned element from the database returned list.  
-* `"limit"`: An *unsigned integer*, optional. (default: 200). Number of element returned from the *offset* element.
+* `"offset"`: A *non-negative integer*, optional. (default: 0). Position of the first returned element from the database returned list.
+* `"limit"`: A *non-negative integer*, optional. (default: 200). Number of element returned from the *offset* element.
```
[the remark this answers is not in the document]
matsduf: @mattias-p, any comment on this?

PR #763, opened 2021-05-20
Update MANIFEST files

mattias-p:
## Purpose

This PR does some cleaning up in the MANIFEST files.

## Changes

Added to the dist tarball:
* newly added unit tests.

Removed from the dist tarball:
* the Travis yaml file, and
* newly deleted files.

New files that aren't added to the dist tarball:
* newly added Github related stuff,
* newly added internal documentation.

Other changes:
* the regex `maint` is replaced with `^maint/`, and
* the regex `.last_cover_stats` is replaced with `^\.last_cover_stats$`.

## How to test this PR

Make sure distcheck doesn't report any problems:
```sh
git clean -dfx
perl Makefile.PL
make distcheck
```


matsduf (commented):
The files in MANIFEST are straight forward because those are files that we have. The patterns in MANIFEST.SKIP are not always straight forward. I think they should be documented with comments in the file. I do not say that this PR must solve all this, but at least the additions to MANIFEST.SKIP should be documented.

matsduf:
Issue #765 asks for check of MANIFEST when PR is created.

matsduf:
> IMHO the new ones in this PR are straight forward enough. 

Presently it is probably so, but when someone looks into it the future it might not be as clear.

> For `maint` and `.last_cover_stats` I looked around on the internet to try to figure out what they mean. I guess I could have removed them instead. I don't believe we're currently using them.

And that could be worth a comment.

matsduf (approved):

matsduf:
@pnax, can you review?

pnax (approved):

PR #764, opened 2021-05-20
Fix broken installation

### LINKED ISSUE #762: Installation fails
I tried to install Backend using cpanm and I got this:

```
Internal error 002: Unexpected error in the _init_db API call: [Failed to initialize the [SQLite] database backend module: [DB
I connect('dbname=/var/lib/zonemaster/db.sqlite','',...) failed: unable to open database file at /root/.cpanm/work/1621443136.
4007/Zonemaster-Backend-6.1.0/blib/lib/Zonemaster/Backend/DB/SQLite.pm line 37. ]]
Internal error 002
# Looks like your test exited with 2 just after 2.
```

mattias-p:
## Purpose

This PR fixes a problem when installing the dist tarball where test01.t would require that /var/lib/zonemaster/ exists.

## Context

Fixes #762.

## Changes

Basically this PR does two things.
1. It makes test01.t independent of the default config file, and
2. It adds a test to config.t to make sure the default config file is valid.

Instead of creating its SQLite database file under /var/lib/zonemaster it creates a temporary directory for the database file. (The temporary directory is automatically cleaned up.)

## How to test this PR

Make sure test suite passes when executed by cpanm:
```sh
make dist
sudo rm -rf /var/lib/zonemaster
cpanm --test-only Zonemaster-Backend-*.tar.gz
```

### DIFF
--- a/t/config.t
+++ b/t/config.t
@@ -6,6 +6,10 @@ use Test::More tests => 2;
 use Test::NoWarnings;
 use Test::Exception;
 use Log::Any::Test;    # Must come before use Log::Any
+
+use File::Basename qw( dirname );
+use File::Slurp qw( read_file );
+use File::Spec::Functions qw( catfile );
 use Log::Any qw( $log );
 
 subtest 'Everything but NoWarnings' => sub {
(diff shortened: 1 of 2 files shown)


matsduf (approved):

matsduf:
@pnax, can you review?

pnax (approved):
Looks good.

PR #767, opened 2021-05-21
Tell OS about failed daemon startup (Linux)

mattias-p:
## Purpose

This PR makes systemd aware of failures during daemon startup. In particular it makes `systemctl {start,status} {zm-rpcapi,zm-testagent}` actually report failures.

## Context

Without this the config validation feature is pretty lame.

## Changes

The LSB service scripts are updated to exit with an error status if any of their commands fail.

I have NOT tested how FreeBSD behaves in this respect. I suggest that is handled separately from this PR.

## How to test this PR

Make sure the daemons are stopped and that the config is invalid:
```
sudo systemctl stop zm-rpcapi
sudo systemctl stop zm-testagent
sudo sed -i '/\bengine\b/ s/=.*/= Excel/' /etc/zonemaster/backend_config.ini
```

Verify that you get an error message when you starting `zm-rpcapi` and that systemctl knows that it failed:
```
sudo systemctl start zm-rpcapi
sudo systemctl status zm-rpcapi
```

Verify that you get an error message when you starting `zm-testagent` and that systemctl knows that it failed:
```
sudo systemctl start zm-testagent
sudo systemctl status zm-testagent
```

### DIFF
--- a/share/zm-rpcapi.lsb
+++ b/share/zm-rpcapi.lsb
@@ -27,7 +27,7 @@ STARMAN=`PATH="$PATH:/usr/local/bin" /usr/bin/which starman`
 . /lib/lsb/init-functions
 
 start () {
-    $STARMAN --user=$USER --group=$GROUP --error-log=$LOGFILE --pid=$PIDFILE --listen=$LISTENIP:$LISTENPORT --preload-app --daemonize $BINDIR/zonemaster_backend_rpcapi.psgi
+    $STARMAN --user=$USER --group=$GROUP --error-log=$LOGFILE --pid=$PIDFILE --listen=$LISTENIP:$LISTENPORT --preload-app --daemonize $BINDIR/zonemaster_backend_rpcapi.psgi || exit 1
 }
 
 stop () {
--- a/share/zm-testagent.lsb
+++ b/share/zm-testagent.lsb
@@ -28,7 +28,7 @@ if [ -n "$ZM_BACKEND_TESTAGENT_LOGLEVEL" ] ; then
 fi
 
 start () {
-    $BINDIR/zonemaster_backend_testagent $testagent_args start
+    $BINDIR/zonemaster_backend_testagent $testagent_args start || exit 1
 }
 
 stop () {


matsduf:
This update is for Linux OS. I will check if we have the same issue in FreeBSD.

mattias-p:
@matsduf and @pnax, can you review?

matsduf (dismissed):
This looks fine as  such. I have not verified that it works on Linux.

FreeBSD has the same problem, but only for RPCAPI daemon. And "set -e" makes no difference.

pnax:
I encounter an error, I cannot revert the described test, the `zm-testagent` daemon does not restart.
1. start the daemons
2. modify the `/etc/zonemaster/backend_config.ini` file with `DB.engine = garbage`
3. restart the daemons `sudo systemctl restart zm-testagent zm-rpcapi`
4. I can see the error with `systemctl status zm=testagent zm-rpcapi`
5. undo the change to `backend_config.ini`
6. restart the deamons
7. the daemon `zm-testagent` fails to restart

Digging a little bit, I redo until step 4 and I looked at the output of `ps aux | grep test`. It seems a test-agent process is still running...

```
$ ps aux | grep test                                                                                                                                                                                                                                                                                                                              
zonemas+    2127  0.0  9.6 122200 96828 ?        S    20:20   0:00 perl /usr/local/bin/zonemaster_backend_testagent --logfile=/var/log/zonemaster/zm-testagent.log --outfile=/var/log/zonemaster/zm-testagent.out --pidfile=/var/run/zonemaster/zm-testagent.pid --user=zonemaster --group=zonemaster start
```

There is no RPCAPI running though `ps aux | grep -i rpc` gives nothing.


Is it normal to still have a process even though the initialization has failed ? Do you think this process prevents to restart the service ?

mattias-p:
@pnax The problem you discovered is unrelated to this PR. The problem is that it is Daemon::Control->run() that performs all the service commands (e.g. start and stop) so when we validate the config file early and terminate if it's invalid, we're actually preventing the stop command from running. I have a PoC to fix this, but I need to clean it up a bit.

Also, according to the [LSB documentation], we shouldn't be using `set -e` in the init scripts. At least not when using /lib/lsb/init-functions. But we already use that in zm-rpcapi.lsb, and we should probably be using it in zm-testagent.lsb as well. A less problematic (though incomplete) solution to the exit status propagation seem to be using `|| exit 1` for the start commands instead.

[LSB documentation]: https://refspecs.linuxfoundation.org/LSB_3.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptfunc.html

mattias-p:
I replaced `set -e` with `|| exit 1` for better compatibility with LSB, and I also rebased to current develop.

matsduf (approved):

pnax (approved):
I've tested and this works.
Thanks for your explanation about the failing testagent.

PR #768, opened 2021-05-21
Implement zmb get_test_params

mattias-p:
## Purpose

Add another RPC API method to zmb.

## Context

N/A

## Changes

Add `zmb get_test_params`.

Rename the `--testid` parameter to `--test-id` for the `test_progress` and `get_test_results` commands. The old parameter name still works but emits a deprecation warning.

## How to test this PR

Start a test and verify that the output from `get_test_params` looks reasonable:
```sh
testid="$(zmb start_domain_test --domain example | jq -r .result)"
zmb get_test_params --test-id "$testid" | jq .
```

Verify that zmtest still works:
```sh
zmtest example.com
```

### DIFF
--- a/script/zmb
+++ b/script/zmb
@@ -249,7 +249,6 @@ sub cmd_start_domain_test {
 
  Options:
    --test-id TEST_ID
-   --testid TEST_ID   # deprecated, use --test-id instead
 
 =cut
 
@@ -257,21 +256,11 @@ sub cmd_test_progress {
     my @opts = @_;
 
     my $opt_lang;
-    my $opt_testid;
     my $opt_test_id;
     GetOptionsFromArray(
         \@opts,
         'test-id|t=s' => \$opt_test_id,
-        'testid|t=s'  => \$opt_testid,
     ) or pod2usage( 2 );
-
-    if ( defined $opt_testid ) {
-        warn( "The --testid parameter deprecated. Use --test-id instead.\n" );
-    }
-    if ( defined $opt_testid && defined $opt_test_id ) {
-        pod2usage( "The --testid parameter is forbidden when --test-id is present." );
-    }
-    $opt_test_id //= $opt_testid;
 
     return to_jsonrpc(
         id     => 1,
@@ -317,7 +306,6 @@ sub cmd_get_test_params {
 
  Options:
    --test-id TEST_ID
-   --testid TEST_ID   # deprecated, use --test-id instead
    --lang LANGUAGE
 
 =cut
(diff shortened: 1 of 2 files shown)


matsduf (dismissed):
It looks fine. It works as expected. I think we should include all API methods into `zmb` (but not in this PR).

matsduf:
@pnax, can you review this?

script/zmb
```diff
@@ -270,6 +270,34 @@ sub cmd_test_progress {
 }
 
 
+=head2 get_test_params
+
+ zmb [GLOBAL OPTIONS] get_test_params [OPTIONS]
+
+ Options:
+   --test-id TEST_ID
```
pnax, line 278: It looks like other methods use `--testid` (see `test_progress` and `get_test_results`). Could you change this one to be consistent with the other ? (Or the other way around).

pnax (changes requested):

script/zmb
matsduf: Outside this change I find two `--testid` and one `--test-id`. I agree that making them consistent would be good. I might also affect `zmtest` for which `zmb` is a backend.

mattias-p: I agree that this sucks, but I actually made it so in order to be consistent with the [RPCAPI](https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#api-method-get_test_params).

matsduf: I think we can start by making it consistent in zmb because it is more likely that you do things manually there. Or support both.

pnax: Could you then update all the other one to `--test-id` in this PR, so they all evolve together ?

mattias-p: I just checked and it's inconsistent all over the place. Maybe we should start migrating to `testid` for test ids across the board. I'll deprecate all `test-id`, `test_id` and `id` that I can find.

mattias-p: I created #790 to better facilitate such a migration. The migration itself is out of scope for this PR.

mattias-p:
I renamed the --testid parameters so they're all named --test-id now.

@pnax @matsduf Please re-review.

mattias-p:
I'll make another update that does away with the deprecations in zmb. It clearly states that it's interface is unstable so we may well make use of that.

matsduf (approved):

matsduf:
@pnax, are your comments on this PR adressed by the last updates?

pnax (approved):
yes, this looks good to me.

pnax:
## v2021.2 testing

LGTM

Issue #761, opened 2021-05-18
Unit test for age_reuse_previous_test


mattias-p:
The test procedure described in #756 should be automated. Doing this should be fairly straight forward after #758 is implemented.

mattias-p:
Fixed by #769.

PR #769, opened 2021-05-25
Control the flow of time from unit test

### LINKED ISSUE #758: Possibility to override the system time in unit tests
Today all testing that involves waiting for time to pass has to be done manually. I guess it would be possible to have a set of slow tests that use `sleep` to make time pass, but that's no fun. A better model is to allow unit test code to manipulate how the application code perceives the current time. That way we could have better coverage of our automatic tests without compromising the ergonomics of testing.

Today we have calls to NOW() in SQL, which means that we'd have to manipulate the time in the MariaDB and PostgreSQL daemons. I have a hunch that won't be easy.

Another approach is to never call NOW() in SQL, but instead get the current time in Perl and send those timestamps into SQL using query parameters. In Perl code we could call `time` or something and the unit tests could [override it](https://perldoc.perl.org/CORE#OVERRIDING-CORE-FUNCTIONS) with something like `sub { CORE::time + $LEAP }` where `$LEAP` can be advanced as needed in the test code.

mattias-p:
## Purpose

This PR introduces rpcapi.t which is intended to host unit tests for the entire RPC-API. It also adds a feature to control the flow of time from unit tests. Finally it makes use of this feature to add a unit test for the behavior of ZONEMASTER.age_reuse_previous_test.

## Context

Fixes #758 and #761.

## Changes

* The formatting commit is just there to reduce the diffs of the remaining commits.
* All questions to SQL about the current time (and some adjacent time calculations) are removed and replaced questions to Perl. It's much easier to control Perl's perception of time, and we'll be able to leverage this to be able to procedurally control time from unit tests. Fixes #758.
* Thanks to the removal of some database specific syntax we're able to deduplicate some (now identical) code in the database specific classes and move it to the common base class.
* Lastly a fix is included for #761. It takes a slightly lighter approach than what's suggested in the issue by not involving any RPC-JSON but rather hitting the RPCAPI::start_domain_test() method directly. It also deviates from the issue by letting time() stand perfectly still except for when advance_time() is called.

N.B. The database schemas still use references to NOW() in default value definitions for certain fields, but all insert queries are updated to make sure that those default values are never used. Clean up of those default value definitions is deferred to a later PR - it's more complicated because it requires an update to the database schema.

### Manual testing
Until #827 is fixed we don't have a good way to automate testing of batch jobs.
I'm opting to not expand this PR to cover #827.
Instead the changes in `DB::add_batch_job` (called from RPCAPI::add_batch_job and add-batch-job.pl) need to be tested manually.

```sh
$ zmb add_api_user --username testuser --api-key secret
$ zmb add_batch_job --username testuser --api-key secret --domain zonemaster.net
$ zmb get_batch_job_result --batch-id 1234
```

### New automatic tests
The changes in these parts are covered by the new rpcapi.t:
* DB::create_new_test
* DB::test_progress
* DB::process_unfinished_tests
* DB::process_dead_test

### Existing automatic tests
The changes in these parts are covered by test01.t:
* DB::test_progress
* DB::test_results

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -18,6 +18,7 @@ use Zonemaster::Backend::Errors;
 requires qw(
   add_batch_job
   create_schema
+  drop_tables
   from_config
   get_test_history
   process_unfinished_tests_give_up
@@ -83,22 +84,6 @@ sub dbh {
     }
 
     return $self->dbhandle;
-}
-
-=head2 drop_tables
-
-Drop all the tables if they exist.
-
-=cut
-
-sub drop_tables {
-    my ( $self ) = @_;
-
-    $self->dbh->do( "DROP TABLE IF EXISTS test_results" );
-    $self->dbh->do( "DROP TABLE IF EXISTS users" );
-    $self->dbh->do( "DROP TABLE IF EXISTS batch_jobs" );
-
-    return;
 }
 
 sub user_exists {
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -10,6 +10,7 @@ use Digest::MD5 qw(md5_hex);
 use Encode;
 use JSON::PP;
 use Log::Any qw( $log );
+use POSIX qw( strftime );
 
 use Zonemaster::Engine::Profile;
 use Zonemaster::Backend::Errors;
@@ -128,15 +129,17 @@ sub create_new_test {
 
     my $priority    = $test_params->{priority};
     my $queue_label = $test_params->{queue};
+    my $now         = time();
+    my $threshold   = $now - $seconds_between_tests_with_same_params;
 
-    my $recent_hash_id = $self->recent_test_hash_id( $seconds_between_tests_with_same_params, $fingerprint );
+    my $recent_hash_id = $self->recent_test_hash_id( $fingerprint, $threshold );
 
     if ( $recent_hash_id ) {
         # A recent entry exists, so return its id
         $hash_id = $recent_hash_id;
     }
     else {
-        $hash_id = substr(md5_hex(time().rand()), 0, 16);
+        $hash_id = substr(md5_hex($now.rand()), 0, 16);
         $dbh->do(
             "INSERT INTO test_results (hash_id, batch_id, priority, queue, fingerprint, params, domain, undelegated) VALUES (?,?,?,?,?,?,?,?)",
             undef,
@@ -473,6 +476,11 @@ sub undelegated {
     return 1 if defined( $$projected_params{ds_info}[0] );
     return 1 if defined( $$projected_params{nameservers}[0] );
     return 0;
+}
+
+sub format_time {
+    my ( $class, $time ) = @_;
+    return strftime "%Y-%m-%d %H:%M:%S", gmtime( $time );
 }
 
 no Moose::Role;
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -85,6 +85,22 @@ sub dbh {
     return $self->dbhandle;
 }
 
+=head2 drop_tables
+
+Drop all the tables if they exist.
+
+=cut
+
+sub drop_tables {
+    my ( $self ) = @_;
+
+    $self->dbh->do( "DROP TABLE IF EXISTS test_results" );
+    $self->dbh->do( "DROP TABLE IF EXISTS users" );
+    $self->dbh->do( "DROP TABLE IF EXISTS batch_jobs" );
+
+    return;
+}
+
 sub user_exists {
     my ( $self, $user ) = @_;
 
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -17,7 +17,7 @@ use Zonemaster::Backend::Errors;
 
 requires qw(
   add_batch_job
-  create_db
+  create_schema
   from_config
   get_test_history
   process_unfinished_tests_give_up
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -166,7 +166,8 @@ sub recent_test_hash_id {
             SELECT hash_id
             FROM test_results
             WHERE fingerprint = ?
-              AND test_start_time > ?
+              AND ( test_start_time IS NULL
+                 OR test_start_time >= ? )
         ],
         undef,
         $fingerprint,
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -21,8 +21,6 @@ requires qw(
   from_config
   get_test_history
   process_unfinished_tests_give_up
-  recent_test_hash_id
-  select_unfinished_tests
   test_progress
   test_results
   get_relative_start_time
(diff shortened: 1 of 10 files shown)


mattias-p:
I've rebased this to current develop. I plan to flesh out the unit tests some more and describe a manual test procedure for whatever remains. But it would be nice to have some eyeballs on the code.

blacksponge:
I have started reading through the changes, already looking good! It's nice to see new tests added for the rpcapi and deduplication of code for DBs. I will try to do some testing later on today.

mattias-p:
BTW, I recommend reviewing the changes commit by commit. At least up to and including "Deduplicate some code". The large commits up to that point are just changing whitespace and moving entire subroutines.

t/rpcapi.t
```diff
@@ -0,0 +1,85 @@
...
+        $rpcapi->{db}->create_db()
+          or BAIL_OUT( "$db_backend database could not be created" );
+    }
+
+    subtest 'start_domain_test' => sub {
+        my $testid1 = $rpcapi->start_domain_test( { domain => "zone1.rpcapi.example" } );
+        advance_time( 11 );
+        my $testid2 = $rpcapi->start_domain_test( { domain => "zone1.rpcapi.example" } );
+
+        $rpcapi->{db}->test_progress( $testid1, 1 );    # mark test as started
```
pnax, line 73: Have you considered using `get_test_request`?

pnax (commented):
I've looked over the commits and it looks very good! I'll run few tests next week.

t/rpcapi.t
mattias-p: I'd forgotten about that! Thanks for pointing it out, I'll update it.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -316,47 +329,37 @@ sub add_batch_job {
...
     my ( $self, $result, $hash_id ) = @_;
 
-    $self->dbh->do("UPDATE test_results SET progress = 100, test_end_time = NOW(), results = ? WHERE hash_id=?", undef, encode_json($result), $hash_id);
+    $self->dbh->do(
+        q[
+            UPDATE test_results
+            SET progress = 100,
+                test_end_time = ?,
+                results = ?
+            WHERE hash_id = ?",
```
blacksponge, line 341: What is this extra double quote for ?

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -266,7 +285,19 @@ sub add_batch_job {
         eval {$dbh->do( "DROP INDEX IF EXISTS test_results__progress " );};
         eval {$dbh->do( "DROP INDEX IF EXISTS test_results__domain_undelegated " );};
 
-        my $sth = $dbh->prepare( 'INSERT INTO test_results (hash_id, domain, batch_id, priority, queue, fingerprint, params, undelegated) VALUES (?, ?, ?, ?, ?, ?, ?, ?) ' );
+        my $sth = $dbh->prepare( '
+            INSERT INTO test_results (
+                hash_id,
+                domain,
+                batch_id,
+                $creation_time,
```
blacksponge, line 293: Shouldn't that be `creation_time` instead of `$creation_time` ?

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -266,7 +285,19 @@ sub add_batch_job {
...
+                priority,
+                queue,
+                fingerprint,
+                params,
+                undelegated
+            ) VALUES (?,?,?,?,?,?,?,?)'
```
blacksponge, line 299: Missing a `?` ?

t/rpcapi.t
mattias-p: I had another look at this and realized `test_progress` has an important advantage over `get_test_request` - it lets you specify a test id. This makes the test robust against unrelated tests in the database waiting to be allocated.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
mattias-p: Fixed

lib/Zonemaster/Backend/DB/SQLite.pm
mattias-p: Fixed

lib/Zonemaster/Backend/DB/SQLite.pm
mattias-p: Fixed

mattias-p:
I tried to restart Travis for the failed test but apparently we're out of credits.

matsduf:
> I tried to restart Travis for the failed test but apparently we're out of credits.

I will take a look at it and request more credits. Usually we get those quickly.

matsduf:
> > I tried to restart Travis for the failed test but apparently we're out of credits.
> 
> I will take a look at it and request more credits. Usually we get those quickly.

Fixed.

mattias-p:
I'm not sure why Travis failed. The error is intermittent. I've updated the unit test to print a little more details when it does fail in this way.

mattias-p:
I've addressed all comments, making updates for most of them, and rebased onto the current develop branch.

However there is still one thing I'm unhappy about. The test file I've added is named rpcapi.t, but only one of three tests is really oriented around an RPCAPI procedure.

mattias-p:
I reframed the unit test to be about the lifecycle of tests. Currently it only covers the parts of the lifecycle that are affected by timeouts.

@blacksponge @matsduf @pnax @tgreenx Please review again.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -154,6 +155,28 @@ sub create_new_test {
...
+    my ( $self, $fingerprint, $threshold ) = @_;
+
+    my $dbh = $self->dbh;
+    my ( $recent_hash_id ) = $dbh->selectrow_array(
+        q[
+            SELECT hash_id
+            FROM test_results
+            WHERE fingerprint = ?
+              AND ( test_start_time IS NULL
+                 OR test_start_time >= ? )
```
pnax, line 170: Nice ! I think this should close the discussion we had in #840.

pnax (approved):
This is a very nice improvement!

mattias-p:
I rebased and resolved the conflict in test01.t.

pnax (commented):
The new `t/lifecycle.t` test file does not drop/create the tables. #947 changed the way we create the tables. Tables creation and removal need to be done within each test plan.

mattias-p:
* I updated lifecycle.t to drop tables. I also refactored test01.t because DRY.
* And I refactored the code to create the schema as well for cleanliness reasons.

t/lifecycle.t
```diff
@@ -0,0 +1,116 @@
+use strict;
+use warnings;
+use 5.14.2;
+
+use Test::More tests => 2;
+use Test::Exception;
+use Test::NoWarnings;
```
pnax, line 7: It seems that PostgreSQL emits warnings when trying to DROP a table that does not exist. This makes the tests in `t/lifecycle.t` fail.

mattias-p: Good catch!

mattias-p:
* I updated drop_tables() to temporarily set the minimum message level.
* I also replaced a `diag` with a `note` to get prettier output from `prove`.

blacksponge (approved):
Looks good to me!

lib/Zonemaster/Backend/DB.pm
```diff
@@ -316,6 +340,37 @@ sub process_unfinished_tests {
...
+            FROM test_results
+            WHERE test_start_time < ?
+            AND progress > 0
+            AND progress < 100
+            AND queue = ?" );
+        $sth->execute(    #
+            $self->format_time( time() - $test_run_timeout ),
+            $queue_label,
+        );
+        return $sth;
```
pnax, line 358: Have you considered having only one `return` statement in this function?

mattias-p: Yeah, I really think this pattern:
```perl
if ( ... ) {
    return ...;
} else {
    return ...;
}
```
is better than this pattern:
```perl
my $z;
if ( ... ) {
    $z = ...;
} else {
    $z = ...;
}
return $z;
```

mattias-p: I guess it could be useful to consider a third pattern here as well:
```perl
...
if ( ... ) {
    return ...;
}
...
return ...;
```
To me this pattern is problematic because it's easy misunderstand the structure of the control flow if you don't spot the return statement in the middle of things. But I don't think the problem is the number of `return` statements.

It's different when **all** branches of an `if / elsif / else` construct return (or throw). In this case the `if / elsif / else` construct is the last thing that happens in the procedure, so there's no code after it where you can get confused about the control flow structure.

PR #771, opened 2021-05-26
Makes SQLite installation shell independent on FreeBSD

matsduf:
Installation must work with both csh and sh.


matsduf:
I did not think of env(), but your suggestion, with a "=" seems to work, i.e.
```
env ZONEMASTER_BACKEND_CONFIG_FILE=/usr/local/etc/zonemaster/backend_config.ini su -m zonemaster -c "perl create_db_sqlite.pl"
```

I will update.

matsduf:
@mattias-p, please re-review.

mattias-p (approved):

PR #772, opened 2021-05-26
Make it easier to copy from installation document on Github

matsduf:
Github makes it easy to copy the entire content in so called fenced code block (e.g. with backticks). This PR splits one such code block because you will never want to paste all lines to a server that you install.


mattias-p (approved):

matsduf:
@pnax, can you review?

pnax (approved):

Issue #789, opened 2021-06-10
RPCAPI must give meaningful error messages


matsduf:
PRs #460 and #494 implement validation of input data to RPCAPI. Some input data is fully controlled by the client (GUI), wheras other data is what the user has provided.

In one way it is good that the digest is validated, but the problem is that the error message from RPCAPI is not shown to the user in the GUI, and even if it did it would be hard to understand it for most users. I removed the last character in a valid hash and sent it to zmb:

```
"error":{
  "message":"Invalid method parameter(s).",
  "code":"-32602",
  "data":"/ds_info/0/digest: String does not match (?^ui:^[a-f0-9]{40}$|^[a-f0-9]{64}$)."
  }
```
It the digest is corrupted you do not get any help. I think that data that users can paste in GUI should be handled by RPCAPI in such a way that a meaningful error message can be returned. Of course, the GUI must also be prepared to display it for the user.

Another possibility of error in digest is that some digit is replaced. In that case the test is executed and there is some meaningful error messages for the user.

For the following set of data, the RPCAPI must return an error message that the GUI can present to the user and be helpful so that the user can correct it and have the domain being tested. The errors here are those that are returned when the cast will not even start.

* domain name
* name server name
* IP address
* DS key tag value
* DS digest

Edit: I updated the PR references above. (@mattias-p 2021-06-14)

mattias-p:
Is it correct that this issue only concerns itself with the following methods?
* start_domain_test
* get_data_from_parent_zone
* get_ns_ips

Could you give examples of error messages you would like to see?

matsduf:
> Is it correct that this issue only concerns itself with the following methods?
> 
>     * start_domain_test
>     * get_data_from_parent_zone
>     * get_ns_ips

Yes. Only those methods that directly depend on user data.

> Could you give examples of error messages you would like to see?

Domain name field:
* Illegal character in domain name
* Domain name too long
* Domain name label too long

Name server name sub field:
* Illegal character in domain name
* Domain name too long
* Domain name label too long

Name server IP address sub field:
* Not a valid IP address

DS key tag sub field:
* Wrong format

DS digest data sub field:
* Incorrect data size
* Illegal character in data
* Data size does not match digest type

This is based on the current GUI where e.g. hash type is a drop down list, which means that you cannot  select an illegal digest type. Error messages could be short if we add documentation for these things, e.g. in the FAQ.

[...]

Issue #703, opened 2021-02-03
RPCAPI input validation error reporting


mattias-p:
Today when you send a request to the RPCAPI with an invalid parameter it responds with an error response. The error message points out the invalid parameter. However if more than one validation error is present, only one of them is noted in the error message.

The error response should list all invalid parameters in the input.

This issue was created in response to zonemaster/zonemaster#925.

blacksponge:
resolved by #820

Issue #637, opened 2020-09-15
Remove duplicated internal calls to validate_syntax


mattias-p:
The RPC API has a `validate_syntax` method. It is called internally for input validation purposes by the implementations of other RPC API methods. Since #399 we have a more formal and reviewable validation layer, absolving the method implementations of much of the input validation responsibility. By validating the input twice (once in the input validation layer and once from within the method implementation) using different validation implementations we gain very little but the risk of bugs is increased to at least the double.
A review should be made to identify where the internal calls to `validate_syntax` simply duplicate work that's already done by the input validation layer. In all those cases the call to `validate_syntax` should be removed.

This issue was prompted by https://github.com/zonemaster/zonemaster-backend/pull/634#issuecomment-692149272.

Future work:
* Replace any remaining internal calls to `validate_syntax` with specific the validations that are not covered by the input validation layer.
* Consider updating the input validation layer to include any resulting specific input validations.
* Consider the viability of `validate_syntax` as a method in the RPC API.

pnax:
The `validate_syntax` method has first evolved in several methods in #820 and the duplication and anonymous methods have been completely removed with #957.

PR #795, opened 2021-06-16
Improve validation error returned

blacksponge:
## Purpose

This PR changes how error related to param validation are return to the client to allow better analysis client side. It also prevent the internal error due to failed validation.

## Context

Addresses #703, is a requirement for https://github.com/zonemaster/zonemaster-gui/issues/177 and https://github.com/zonemaster/zonemaster-gui/issues/178 .

## Changes

* First the json validation error are split in an array of object to allow better parsing client side (thus enabling better error message on the GUI)
* Second the errors coming from `validate_syntax` are also collected in an array instead of just returning the first one. The two sets of error are concatenated.

While I was doing this I noticed that some validation test in `validate_syntax` are duplicates of the json validation (see the `NOTE: ...` inserted in the code). Is that normal?
## How to test this PR

`curl -sH "Content-Type: application/json" --data '{"method":"start_domain_test","params":{"nameservers":[{"ns":"/!\\", "ip":":::1"}], "ds_info":[{"alg":"plop"}], "client_id":"toto", "domain":" iis.se"},"id":1,"jsonrpc":"2.0"}' localhost:5000/api | jq`  
```json
{
  "error": {
    "data": [
      {
        "message": "Missing property.",
      },
      {
        "path": "/ds_info/0/digest",
        "message": "Missing property."
      },
      {
        "path": "/ds_info/0/digtype",
        "message": "Missing property."
      },
      {
        "path": "/ds_info/0/keytag",
        "message": "Missing property."
      },
      {
        "message": "The domain name character(s) not supported",
        "path": "/domain"
      },
      {
        "path": "/nameservers/0/ns",
        "message": "The domain name character(s) not supported"
      },
      {
        "message": "Invalid IP address: [:::1]",
        "path": "/nameservers/0/ns"
      },
      {
        "message": "Invalid digest format: []",
        "path": "/ds_info/0/digest"
      }
    ],
    "code": "-32602"
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

Example of duplicate validation (field `notnameservers`)

`curl -sH "Content-Type: application/json" --data '{"method":"start_domain_test","params":{"notnameservers":[]},"id":1,"jsonrpc":"2.0"}' localhost:5000/api | jq`
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "error": {
    "data": [
      {
        "path": "/",
        "message": "Properties not allowed: notnameservers."
      },
      {
        "path": "/",
        "message": "Unknown option [notnameservers] in parameters"
      },
      {
        "path": "/domain",
        "message": "Domain name required"
      }
    ],
    "code": "-32602",
  }
}
```

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -28,6 +28,7 @@ use Zonemaster::Backend::Validator;
 
 my $zm_validator = Zonemaster::Backend::Validator->new;
 my %json_schemas;
+my %extra_validators;
 my $recursor = Zonemaster::Engine::Recursor->new;
 
 sub joi {
@@ -248,16 +249,20 @@ sub validate_syntax {
     my ( $self, $syntax_input ) = @_;
 
     my $result = eval {
+        my @errors;
+
         my @allowed_params_keys = (
             'domain',   'ipv4',      'ipv6', 'ds_info', 'nameservers', 'profile',
             'client_id', 'client_version', 'config', 'priority', 'queue'
         );
 
+        # NOTE: Already covered in JSON validation? (strict mode)
         foreach my $k ( keys %$syntax_input ) {
-            return { status => 'nok', message => encode_entities( "Unknown option [$k] in parameters" ) }
+            push @errors, { path => '/', message => encode_entities( "Unknown option [$k] in parameters" ) }
             unless ( grep { $_ eq $k } @allowed_params_keys );
         }
 
+        # NOTE: Already covered in JSON validation? (strict mode)
         if ( ( defined $syntax_input->{nameservers} && @{ $syntax_input->{nameservers} } ) ) {
             foreach my $ns_ip ( @{ $syntax_input->{nameservers} } ) {
                 foreach my $k ( keys %$ns_ip ) {
@@ -269,21 +274,24 @@ sub validate_syntax {
         if ( ( defined $syntax_input->{ds_info} && @{ $syntax_input->{ds_info} } ) ) {
             foreach my $ds_digest ( @{ $syntax_input->{ds_info} } ) {
                 foreach my $k ( keys %$ds_digest ) {
+                    # NOTE: Already covered in JSON validation? (strict mode)
                     delete( $ds_digest->{$k} ) unless ( $k eq 'algorithm' || $k eq 'digest' || $k eq 'digtype' || $k eq 'keytag' );
                 }
             }
         }
 
+        # NOTE: Already covered in JSON validation?
         if ( defined $syntax_input->{ipv4} ) {
-            return { status => 'nok', message => encode_entities( "Invalid IPv4 transport option format" ) }
+            push @errors, { path => '/ipv4', message => encode_entities( "Invalid IPv4 transport option format" ) }
             unless ( $syntax_input->{ipv4} eq JSON::PP::false
                 || $syntax_input->{ipv4} eq JSON::PP::true
                 || $syntax_input->{ipv4} eq '1'
                 || $syntax_input->{ipv4} eq '0' );
         }
 
+        # NOTE: Already covered in JSON validation?
         if ( defined $syntax_input->{ipv6} ) {
-            return { status => 'nok', message => encode_entities( "Invalid IPv6 transport option format" ) }
+            push @errors, { path => '/ipv6', message => encode_entities( "Invalid IPv6 transport option format" ) }
             unless ( $syntax_input->{ipv6} eq JSON::PP::false
                 || $syntax_input->{ipv6} eq JSON::PP::true
                 || $syntax_input->{ipv6} eq '1'
@@ -292,42 +300,44 @@ sub validate_syntax {
 
         if ( defined $syntax_input->{profile} ) {
             my @profiles = map lc, $self->{config}->ListPublicProfiles();
-            return { status => 'nok', message => encode_entities( "Invalid profile option format" ) }
+            push @errors, { status => '/profile', message => encode_entities( "Unknown profile name" ) }
             unless ( grep { $_ eq lc $syntax_input->{profile} } @profiles );
         }
 
         my ( undef, $dn_syntax ) = $self->_check_domain( $syntax_input->{domain}, 'Domain name' );
 
-        return $dn_syntax if ( $dn_syntax->{status} eq 'nok' );
+        push @errors, { path => "/domain", message => $dn_syntax->{message} } if ( $dn_syntax->{status} eq 'nok' );
 
         if ( defined $syntax_input->{nameservers} && @{ $syntax_input->{nameservers} } ) {
-            foreach my $ns_ip ( @{ $syntax_input->{nameservers} } ) {
+            while (my ($index, $ns_ip) = each @{ $syntax_input->{nameservers} }) {
                 my ( $ns, $ns_syntax ) = $self->_check_domain( $ns_ip->{ns}, "NS [$ns_ip->{ns}]" );
-                return $ns_syntax if ( $ns_syntax->{status} eq 'nok' );
-            }
+                push @errors, { path => "/nameservers/$index/ns", message => $ns_syntax->{message} } if ( $ns_syntax->{status} eq 'nok' );
 
-            foreach my $ns_ip ( @{ $syntax_input->{nameservers} } ) {
+                my $ip_error;
+
+                # NOTE: IP regex check already covered in JSON validation? (Regex differs a bit)
                 # Although counterintuitive both tests are necessary as Zonemaster::Engine::Net::IP accepts incomplete IP adresses (network adresses) as valid IP adresses
-                return { status => 'nok', message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
+                $ip_error = { path => "/nameservers/$index/ns", message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
                     unless( !$ns_ip->{ip} || $ns_ip->{ip} =~ /^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/ || $ns_ip->{ip} =~ /^([0-9A-Fa-f]{1,4}:[0-9A-Fa-f:]{1,}(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?)|([0-9A-Fa-f]{1,4}::[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})$/);
 
-                return { status => 'nok', message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
-                unless ( !$ns_ip->{ip}
+                $ip_error = { path => "/nameservers/$index/ns", message => encode_entities( "Invalid IP address: [$ns_ip->{ip}]" ) }
+                unless ( $ip_error || !$ns_ip->{ip}
                     || Zonemaster::Engine::Net::IP::ip_is_ipv4( $ns_ip->{ip} )
                     || Zonemaster::Engine::Net::IP::ip_is_ipv6( $ns_ip->{ip} ) );
+
+                push @errors, $ip_error if $ip_error
             }
 
-            foreach my $ds_digest ( @{ $syntax_input->{ds_info} } ) {
-                return {
-                    status  => 'nok',
+            # NOTE: Already covered in JSON validation? (emit warning when key do not exist)
+            while (my ($index, $ds_digest) = each @{ $syntax_input->{ds_info} }) {
+                push @errors, {
+                    path  => "/ds_info/$index/algorigthm",
                     message => encode_entities( "Invalid algorithm type: [$ds_digest->{algorithm}]" )
                 }
                 if ( $ds_digest->{algorithm} =~ /\D/ );
-            }
 
-            foreach my $ds_digest ( @{ $syntax_input->{ds_info} } ) {
-                return {
-                    status  => 'nok',
+                push @errors, {
+                    path => "/ds_info/$index/digest",
                     message => encode_entities( "Invalid digest format: [$ds_digest->{digest}]" )
                 }
                 if (
@@ -337,6 +347,14 @@ sub validate_syntax {
                         $ds_digest->{digest} =~ /[^A-Fa-f0-9]/
                 );
             }
+        }
+
+        if (@errors) {
+            return {
+                status => 'nok',
+                errors => \@errors,
+                message => 'Syntax validation error'
+            };
         }
     };
     if ($@) {
@@ -367,14 +385,13 @@ $json_schemas{start_domain_test} = joi->object->strict->props(
     priority => $zm_validator->priority,
     queue => $zm_validator->queue
 );
+$extra_validators{start_domain_test} = \&validate_syntax;
 sub start_domain_test {
     my ( $self, $params ) = @_;
 
     my $result = 0;
     eval {
         $params->{domain} =~ s/^\.// unless ( !$params->{domain} || $params->{domain} eq '.' );
-        my $syntax_result = $self->validate_syntax( $params );
-        die "$syntax_result->{message} \n" unless ( $syntax_result && $syntax_result->{status} eq 'ok' );
 
         die "No domain in parameters\n" unless ( $params->{domain} );
 
@@ -673,6 +690,15 @@ sub jsonrpc_validate {
             };
         }
         my @error = $method_schema->validate($jsonrpc_request->{params});
+
+        @error = map { { path => $_->path, message => $_->message } } @error;
+
+        if ( $extra_validators{$jsonrpc_request->{method}} ) {
+            my $sub_validator = $extra_validators{$jsonrpc_request->{method}};
+            my $result = $self->$sub_validator($jsonrpc_request->{params});
+            push @error, @{$result->{errors}} if $result->{status} eq 'nok';
+        }
+
         if ( @error ) {
             return {
                 jsonrpc => '2.0',
@@ -680,7 +706,7 @@ sub jsonrpc_validate {
                 error => {
                     code => '-32602',
                     message => 'Invalid method parameter(s).',
-                    data => "@error"
+                    data => \@error
                 }
             };
         }


matsduf:
Please see #789 too.

mattias-p:
> While I was doing this I noticed that some validation test in validate_syntax are duplicates of the json validation (see the NOTE: ... inserted in the code). Is that normal?

It's an old and mostly obsolete validation layer that no one has found the time to rip out. There's actually an issue for that: #637.

blacksponge:
> Please see #789 too. Do you adress that too?

It was not my main focus but I can see if what I can do.

>  Have you thought of the need of translation?

Yes, I have, still need to think about it though.

> It's an old and mostly obsolete validation layer that no one has found the time to rip out. There's actually an issue for that: #637.

Okay, thanks for confirming that. I'll have a look into it since I have already started.

matsduf:
I think we all agree that duplicate validation is bad, but I suggest that it is handled in a separate PR to keep down the size of this PR.

matsduf:
> > Please see #789 too. Do you address that too?
> 
> It was not my main focus but I can see if what I can do.

What in #789 is not in focus of this PR?

This PR suggests a new response to `start_domain_test`. I should also be documented in the API documentation.

It is important that translation of these error messages is handled correctly. Either translation can be done here in Backend or in the client (GUI).

If it is done in Backend, then `start_domain_test` query must be updated with language specification and it must be determined how the translation operation should be done. Currently there are no strings belonging to Backend being translated. Only strings that comes from Engine are being translated.

The other option is to let the client (GUI) translate, but that is probably an inferior solution since the messages come from Backend.

blacksponge:
@matsduf I've made an attempt to use custom messages for the JSON validation, I am not sure of the design though (PR description updated).

matsduf:
@blacksponge, I think the attempt looks good and promising. I am glad that you have take these steps. This will be an improvement of GUI.

> @matsduf I've made an attempt to use custom messages for the JSON validation, I am not sure of the design though (PR description updated).

1. Why don't you include #789 to be adressed by this PR?
2. I'd like to see a draft update of the API documentation along with the code update. Would you like to get help with that work?
3. I also like to see translation to be part of the description and goal of the RP. If we do not included from start it might be very complex to add it afterwards. A start is to list the alternatives that we see.

blacksponge:
> Why don't you include #789 to be adressed by this PR?

I have started addressing it in the last commit, 1416df0 propose a way to have custom error message when validating against the json schema.

I will work on documentation, in the mean time I have updated the description to list the current error messages that we have control of (mostly those that are displayed in the GUI). More check could be added afterwards for the remaining errors that you mentioned in #789.

docs/API.md
```diff
@@ -139,12 +139,12 @@ Basic data type: object
...
 
 Properties:
-* `"digest"`: A string, required. Either 40 or 64 hexadecimal characters (case insensitive).
+* `"digest"`: A string, required. Either 40, 64 or 96 hexadecimal characters (case insensitive).
 * `"algorithm"`: An non negative integer, required.
 * `"digtype"`: An non negative integer, required.
 * `"keytag"`: An non negative integer, required.
 
-Extra properties in *DS info* objects are ignored when present in RPC method arguments, and never returned as part of RPC method results.
+No extra properties are allowed.
```
matsduf, line 147: I think this statement should be moved to "the top", i.e. apply for all objects.

docs/API.md
```diff
@@ -153,8 +153,7 @@ Basic data type: string
 
 This parameter is a string that are an IPv4 or IPv6. It's validated with the following regexes:
  - IPv4 : `/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/`
- - IPv6 : `/^([0-9A-Fa-f]{1,4}:[0-9A-Fa-f:]{1,}(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?)|([0-9A-Fa-f]{1,4}::[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})$/`
-
+ - IPv6 : `/^[0-9a-f:]+(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?$/i`
```
matsduf, line 156: This is not fully correct. 300.300.300.300 matches the expression for IPv4, but it is not valid. I think it would be better to write:
```
IPv4: valid IPv4 address in [dot-decimal notation]
IPv6: valid IPv6 address in the [recommend text format for IPv6 addresses]


[Dot-decimal notation]: https://en.wikipedia.org/wiki/Dot-decimal_notation
[Recommend text format for IPv6 addresses]: https://datatracker.ietf.org/doc/html/rfc5952

File: docs/API.md [hunk 3]
```diff
@@ -187,12 +186,18 @@ When a method received an unknown *profile name* value for in parameter with thi
...
+    "data": [
+      {
+        "path": "/profile",
+        "message": "Unknown profile"
+      }
+    ],
+    "message": "Invalid method parameter(s)."
+  },
+  "id": 1,
+  "jsonrpc": "2.0"
```
matsduf, line 200: This should be included with the error message below, shouldn't it?

docs/API.md
```diff
@@ -619,8 +624,8 @@ An object with the following properties:
 * `"ipv4"`: A boolean, optional. (default `true`). Used to configure the test and enable IPv6 tests.
 * `"nameservers"`: A list of *name server* objects, optional. (default: `[]`). Used to perform un-delegated test.
 * `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated**. A string, optional. Ignored. Specify `"profile"` instead.
+* `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile. It must be a profile listed in the [available profiles].
+* `"config"`: **Deprecated**. Will cause an errror, specify `"profile"` instead.
```
matsduf, line 628: I think `config` is enough deprecated and could be removed.

docs/API.md
```diff
@@ -633,16 +638,50 @@ An object with the following properties:
...
+        "path": "/nameservers/0/ip",
+        "message": "Invalid IP address"
+      }
+    ],
+    "message": "Invalid method parameter(s)."
+  },
+  "id": 1,
+  "jsonrpc": "2.0"
+}
+```
```
matsduf, line 684: Could the `data` property be formally defined?

docs/API.md
```diff
@@ -139,12 +139,12 @@ Basic data type: object
 DS for [Delegation Signer](https://tools.ietf.org/html/rfc4034) references DNSKEY-records in the sub-delegated zone.
 
 Properties:
-* `"digest"`: A string, required. Either 40 or 64 hexadecimal characters (case insensitive).
+* `"digest"`: A string, required. Either 40, 64 or 96 hexadecimal characters (case insensitive).
```
matsduf, line 142: I am not sure that we really should have these values set here or in the validating code. This will prevent someone from using Zonemaster with an experimental digest type with another lenght.

mattias-p: Are experimental digest types supported by stock Zonemaster::Engine and Zonemaster::LDNS today? And if not, would it be difficult to add that feature?

docs/API.md
mattias-p: In the Protocol section we state that "Extra properties in the "params" object are allowed for some methods but ignored for others.". Do you know if we still have any methods that allow extra properties? I thought we got rid of that when we added the validation layer.

docs/API.md
mattias-p: Two things that are really nice about regular expressions is that they're portable and easy to compare. This is useful when you want to review the correctness of the implementation or when you want to perform client-side validation that's identical to the server-side validation.

I advocate a combined model based on the intersection between a regular expression and an optional free-form free-form validation algorithm description. Having a regular expression that overlays the free-form algorithm makes it easier to verify the validation logic as a whole because you only need to assess some aspects of the free-form algorightm logic.

In this case we'd specify two loose regular expressions, together with some free-form wording (like what @matsduf suggests above) detailing the additional requirements on top of the regular expressions.

docs/API.md
mattias-p: I'm not super comfy about detailing the behavior of methods in the data type definitions. E.g. what if you have two methods that take a profile argument, but in different locations of their arguments structures? I think it's better to let each method list its own error modes. Though in the case of validation errors it would be better with a generic description of how the validation errors are structured (with paths and messages etc) in the error response.

docs/API.md
mattias-p: For validation errors we should have a generic formal description somewhere at the top.

Regarding the old text, I think that came about because the API documentation doesn't contain all the information required to fully validate the profile names. The *profile name* data type specifies the general pattern in terms of a regular expression. But the check against the configuration wasn't performed by the validation layer at all, but rather by the method itself. (I haven't had time to look at what the changes in this PR yet, only the discussions.)

docs/API.md
blacksponge: Done.

docs/API.md
blacksponge: I updated this paragraph to mentioned both the free-form wording and regex.

docs/API.md
blacksponge: This example response has been removed

docs/API.md
blacksponge: > I think this statement should be moved to "the top", i.e. apply for all objects.

Done.

> Do you know if we still have any methods that allow extra properties?

None that I have seen.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -193,10 +200,10 @@ sub get_data_from_parent_zone {
 
 
 sub _check_domain {
-    my ( $self, $domain, $type ) = @_;
+    my ( $self, $domain ) = @_;
 
     if ( !defined( $domain ) ) {
-        return ( $domain, { status => 'nok', message => encode_entities( "$type required" ) } );
+        return ( $domain, { status => 'nok', message => N__ 'Domain name required' } );
```
matsduf, line 206: Does this `N__` translation function use MO files, compiled from PO files, in the same way as in Engine? In Engine (and CLI) it is possible to extract the strings to be translated. Does it work the same way here? Could you provide an example file with translations into French?

blacksponge: `N__` is a no op function to allow extraction of string for translation ([realted doc](https://metacpan.org/pod/Locale::TextDomain#N__(ARG1))). The real translation is done [line 737](https://github.com/zonemaster/zonemaster-backend/pull/795/files/a2c13b9199165c9885ebbe70ddff3e7e3523787f#diff-f159417cb7e91c84742d1cb44858b1eed39152de169ecc78fa6677fe879ac1b2R737). 

I will provide a base translation in French once I have a better setup regarding i18n (ie add makefile for extraction and generation, and fixing paths)

matsduf: Method `start_domain_test` does not include the language that the message is to be translated to. Do you suggest that to be added?

One option is to let Backend return codes like the messages tags, e.g. "DOMAIN_NON_ASCII_NO_IDN_SUPPORT", that should be fixed and not updated, and then use the normal translation mechanism in GUI (adding strings to en.json, fr.json, sv.json etc). This means that it is up to the client to produce the message. If we add new error messages (error codes) the client has to be updated.

The other option is to include language code in the calls and let Backend translate. But what should happen then if the user changes language? When the result is presented Backend re-requests for the result with a new language tag.

Today we have to solutions for translation files, one is the PO/MO files used by Engine and CLI, the other are the language specific json files used by GUI. It would be best of we could have a solution that is the same as one of the two. That could be json files in Backend if that is the best solution.

blacksponge: My current implementation uses PO/MO files on the backend. I personally prefer server-side translation because it allows for internationalization of multiple client (eg. a cli like zmtest / zmb could display the correct error message out of the box, without the need of duplicated translation files). It would probably use a parameter in the request, or a more ideal solution IMO is to use the accept-language header, which doesn't need to make any change in the request (but as we are already using a lang param for the result, we could stick to that solution to not have multiple way of doing similar stuff, I guess).

The only remaining point being if someone change their language after submitting a erroneous form. I haven't found a good solution yet. I think resending the form on the user behalf is not good UX, especially if they are in the process of correcting it.

On the other hand, having a unified translation file for client-side and server-side error is not a bad idea to have unified error messages without having to maintain duplicates translations. In that regard, I find your idea of having the backend serving json translation files interesting as it allows for centralized translation between clients without too much duplicates, beside the translation logic it self, and also allows for translation/validation updates without updating the GUI.

matsduf: There are pros and cons on putting the messages in Backend and GUI, but if we use PO/MO on Backend that should work fine, but we have to add the infrastructure there too to make it work smothly. The solution from Engine should be copied.

We should have the same solution to transfer selected language in all methods that need language. If we change to accept-language header it will mean breaking changes.

I do not think that not being able to change language is a show-stopper, so we should evaluate pros and cons.

docs/API.md
matsduf: I realize that the GUI only accepts pre-defined digest types, so if the focus of the API is the GUI and public systems (which it is) we can keep it as-is. If we get other requests then we could change it.

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: I have a semi-working implementation of translation with PO/MO files in the backend, heavily inspired by what has been done in the engine. Before going to far I want to test the client side translation with backend served json files, to see if we can benefit from that.

docs/API.md
matsduf: I do not think that the regular expressions add anything unless they are updated to correctly capture a valid IP address. The text talks about how it is implemented. That is not part of the API. The same API could be implemented without regular expressions.

docs/API.md
```diff
@@ -304,6 +291,14 @@ I.e. a string matching `/^[a-zA-Z0-9-.@]{1,50}$/`.
 
 Represents the name of an authenticated account (see *[Privilege levels]*)
 
+### Validation error data
+
+Basic data type: array
+
+The items of the array are objects with two keys, `"path"` and `"message"`:
+* `"path"`: a string. The JSON path a element in the request's param object.
```
matsduf, line 299: Is the way of giving paths standardized in JSON? Is the a reference? Could we have an example in the text?

blacksponge: It is a standard, probably not the only one but I had the name wrong, [RFC6901](https://datatracker.ietf.org/doc/html/rfc6901) standardized it as JSON Pointer. I will update the api doc accordingly.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -701,6 +694,21 @@ sub jsonrpc_validate {
                 }
             };
         }
+
+        my %locale = $self->{config}->Language_Locale_hash();
+
+        # NOTE: Default to first language in config
+        my $language = $jsonrpc_request->{params}->{language} // $self->{config}->{_LANGUAGE_locale}[0];
```
matsduf, line 701: It could be OK that the first `locale` i the default, but then it should be documented.

Another comment is that what comes over RPCAPI is normally a two-character language code, e.g. "en", whereas `$self->{config}->{_LANGUAGE_locale}[0]` will return a language-country string, e.g. "en_US".

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -701,6 +694,21 @@ sub jsonrpc_validate {
...
+
+        # NOTE: temporary workaround to not break everything
+        if (exists $jsonrpc_request->{params}->{language} && !exists $method_schema->{properties}->{language}) {
+            delete $jsonrpc_request->{params}->{language};
+        }
+
+        # TODO: Language validation
+        $ENV{LANGUAGE} = $locale{$language};
+        setlocale( LC_MESSAGES, $locale{$language} );
```
matsduf, line 710: I think `$language` here should be a full locale string, e.g. `en_US.UTF-8`.

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: > It could be OK that the first locale i the default, but then it should be documented.

Yes I have planned to update the doc on that matter.

> Another comment is that what comes over RPCAPI is normally a two-character language code, e.g. "en", whereas $self->{config}->{_LANGUAGE_locale}[0] will return a language-country string, e.g. "en_US".

The local hash returned by `Language_Locale_hash` support both two chars language code and language contry string.

```perl
{
          'fr_FR' => 'fr_FR.UTF-8',
          'sv' => 'sv_SE.UTF-8',
          'fi' => 'fi_FI.UTF-8',
          'sv_SE' => 'sv_SE.UTF-8',
          'fi_FI' => 'fi_FI.UTF-8',
          'en_US' => 'en_US.UTF-8',
          'fr' => 'fr_FR.UTF-8',
          'da' => 'da_DK.UTF-8',
          'nb' => 'nb_NO.UTF-8',
          'da_DK' => 'da_DK.UTF-8',
          'nb_NO' => 'nb_NO.UTF-8',
          'en' => 'en_US.UTF-8'
};
```

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: The language code to full locale conversion is performed by the `%locale` hash (see comment above).

docs/API.md
```diff
@@ -55,6 +55,8 @@ If the request object is invalid JSON, an error with code `-32700` is reported.
 
 If no method is specified or an invalid method is specified, an error with code `-32601` is reported.
 
+If no `params` object is specified, or the `param` object for the specified method is invalid, an error with code `-32602` is reported. For more information on the validation error data format see [Validation error data]. If an optionnal `language` field containing a [Language tag] is specified within the `params` object, error messages will be translated to the target language,
```
matsduf, line 58: Please keep the lines by a maximum of 80 characters.

Some API methods do no have any params object, what happens to them?

"optionnal" -> "optional"

The language field is mandatory in the "get_test_results" method. That collides with the statement that it is optional above.

If the "language" field is added to all or some methods,  I think it must be specified by the method. I think we could start by adding it to the "start_domain_test", "get_host_by_name" and "get_data_from_parent_zone".

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -646,11 +646,11 @@ sub jsonrpc_validate {
 
         my @error_response;
 
-        # NOTE: Make that configurable
+        # TODO: Make that configurable
         my $default_language =  'en_US';
         my ($locale, $locale_error) = $self->_get_locale( $jsonrpc_request->{params} );
 
-        # NOTE: temporary workaround to not break everything
+        # remove language param if not present in the real method parameters
```
matsduf, line 653: Why is it removed?

docs/API.md
blacksponge: Now, only the stated methods support the language field in param.

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: I first changed the comment to keep the implementation at a global lebel, now that the implementation has moved to only a few specific methods this is not useful anymore.

docs/API.md
```diff
@@ -55,6 +55,10 @@ If the request object is invalid JSON, an error with code `-32700` is reported.
 
 If no method is specified or an invalid method is specified, an error with code `-32601` is reported.
 
+If no `params` object is specified, or the `param` object for the specified
+method is invalid, an error with code `-32602` is reported. For more information
+on the validation error data format see [Validation error data].
```
matsduf, line 60: How does this match API methods that are defined to return JSON without any param object?

docs/API.md
```diff
@@ -1192,4 +1228,11 @@ The `"params"` object sent to `start_domain_test` or `add_batch_job` when the *t
...
 [LANGUAGE.locale]:              Configuration.md#locale
 [Language tag]:                 #language-tag
+<<<<<<< HEAD
+[Validation error data]:        #validation-error-data
+[Dot-decimal notation]:         https://en.wikipedia.org/wiki/Dot-decimal_notation
+[Recommend text format for IPv6 addresses]: https://datatracker.ietf.org/doc/html/rfc5952
+[JSON Pointer]:                 https://datatracker.ietf.org/doc/html/rfc6901
+=======
 [Privilege levels]:             #privilege-levels
+>>>>>>> origin/develop
```
matsduf, line 1238: Conflict marks?

docs/Configuration.md
```diff
@@ -204,9 +204,10 @@ language from being allowed. If there are more than one `locale tag`
 (with different country codes) for the same language, then
 all those must be removed to block that language.
 
-English is the Zonemaster default language, but it can be blocked
-from being allowed by RPC-API by not including it in the
-configuration.
+The first language in the list will be used as the default for the RPC API.
+If the translation is unavailable it will then default to English.
+Some other part of Zonemaster will default to English directly but it can be
+blocked from being allowed by RPC-API by not including it in the configuration.
```
matsduf, line 210: This written too general. The default only applies to the error messages for the API methods that have support for translated error messages.

"If the translation is unavailable it will then default to English." ==> "If translation is not avaiable, then the error messages will be sent untranslated, i.e. in English."

I do not understand:
"Some other part of Zonemaster will default to English directly but it can be
blocked from being allowed by RPC-API by not including it in the configuration."

I think that most of the information should be in API.md. In Configuration.md it only needs to be stated that the first `locale tag` will give the default language for error messages in some API methods that can be seen in the API.md.

docs/API.md
blacksponge: From what i have seen, all methods need a params object, if no parameters are needed then this object is empty, as it is the case with `version_info` for example:

```perl
$json_schemas{version_info} = joi->object->strict;
```

docs/API.md
blacksponge: oops

docs/Configuration.md
```diff
@@ -204,9 +204,10 @@ language from being allowed. If there are more than one `locale tag`
 (with different country codes) for the same language, then
 all those must be removed to block that language.
 
-English is the Zonemaster default language, but it can be blocked
-from being allowed by RPC-API by not including it in the
-configuration.
```
matsduf, line 209: This section should not be removed. It is updated in another PR in commit https://github.com/zonemaster/zonemaster-backend/pull/796/commits/b3a51cb1e683747f2a0e72ba75ba8711bdd41f1f

Now that translation is added for error messages, we still have the translation of test result, and that should not change.

blacksponge: I've re-added it.

docs/Configuration.md
blacksponge: This section has been updated.

docs/API.md
matsduf: That is not how it is defined in the API doucmentation, e.g. https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/API.md#api-method-version_info

blacksponge: My mistake, this schema is defined but never used as the params object is only validated if the schema has at least one property marked as required (I am not sure this is the right thing to do, a methods with only optional parameters would never have its request validated).

matsduf:
@blacksponge, can you resolve the conflict in lib/Zonemaster/Backend/Config.pm?

docs/API.md
matsduf: Can you remove the following lines since they do not make sense in an API document:
```
In addition a preliminary check is performed with the following regexes:
 - IPv4 : `/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/`
 - IPv6 : `/^[0-9a-f:]+(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?$/i`
```

docs/API.md
```diff
@@ -474,9 +475,10 @@ Example response:
 
 #### `"params"`
 
-An object with the property:
+An object with the properties:
 
-`"hostname"`: A *domain name*, required. The hostname whose IP addresses are to be resolved.
+* `"hostname"`: A *domain name*, required. The hostname whose IP addresses are to be resolved.
+* `"language"`: A [Language Tag], optional, used for validation error messages translation.
```
matsduf, line 483: Below there is a section "#### error" with no information. Could that be updated with the new error report that can be returned?

share/backend_config.ini
```diff
@@ -32,12 +32,11 @@ database_file = /var/lib/zonemaster/db.sqlite
 #maximal_number_of_retries=3
 
 [LANGUAGE]
-locale = da_DK en_US fi_FI fr_FR nb_NO sv_SE
+locale = en_US da_DK fi_FI fr_FR nb_NO sv_SE
```
matsduf, line 36: I suggest a comment here that the first locale tag has a special status.

matsduf (changes requested):

matsduf (changes requested):
```
$ make distcheck
No such file: share/locale/fr/LC_MESSAGES/Zonemaster-Engine.mo
Not in MANIFEST: inc/Module/Install/External.pm
Not in MANIFEST: share/GNUmakefile
Not in MANIFEST: share/locale/fr/LC_MESSAGES/Zonemaster-Backend.mo
Not in MANIFEST: share/po/fr.mo
Not in MANIFEST: share/po/fr.po
Not in MANIFEST: share/po/Zonemaster-Backend.pot
```

share/po/fr.po
```diff
@@ -0,0 +1,83 @@
+# SOME DESCRIPTIVE TITLE.
```
matsduf, line 1: In Engine and CLI the PO files are stored in `share/`, not in `share/po/`. We should have the same model in all respositories when possible.

The PO files could be moved to `share/po/`, but then it should be done in all three respositories. In this repository there are quite a few files already inte `share/`.

matsduf (changes requested):

matsduf (changes requested):
Installation of package fails:
```
Running Makefile.PL
Configuring Zonemaster-Backend-6.2.0 ... String found where operator expected at Makefile.PL line 55, near "requires_external_bin 'find'"
	(Do you need to predeclare requires_external_bin?)
String found where operator expected at Makefile.PL line 58, near "requires_external_bin 'gmake'"
	(Do you need to predeclare requires_external_bin?)
syntax error at Makefile.PL line 55, near "requires_external_bin 'find'"
syntax error at Makefile.PL line 58, near "requires_external_bin 'gmake'"
Execution of Makefile.PL aborted due to compilation errors.
N/A
! Configure failed for Zonemaster-Backend-6.2.0. See /home/matsd/.cpanm/work/1625217595.898/build.log for details.
```

blacksponge:
> Installation of package fails

I cannot reproduce your error, this part is taken from the engine, could it be a missing dependency? I'll try to dig into it. Nevertheless I have updated the make files for other reason.

blacksponge:
I have booted a freebsd 13 virtual machine with `perl5`, `git`, `p5-Module-Install`, `gmake` installed and I still can't reproduce your issue.

matsduf:
> > Installation of package fails
> 
> I cannot reproduce your error, this part is taken from the engine, could it be a missing dependency? I'll try to dig into it. Nevertheless I have updated the make files for other reason.

You have to include "inc/Module/Install/External.pm" in MANIFEST.

blacksponge:
> You have to include "inc/Module/Install/External.pm" in MANIFEST.

Oh that was the issue. I have updated the MANIFEST.

share/po/fr.po
blacksponge: I've moved them back to `share/`

docs/API.md
blacksponge: I have updated the error section of `get_data_from_parent_zone` and `get_host_by_name`.

share/backend_config.ini
blacksponge: Done.

matsduf:
Both share/GNUmakefile and share/Makefile should be in MANIFEST. See Engine.

matsduf:
Now it can be installed. I will do some testing. @blacksponge, please resolve the conflict so that we can see the resulting docuemntation.

matsduf:
I have installed Backend from this PR and GUI from zonemaster/zonemaster-gui#230, and many error checks seems to work and are reported as they should, and when I do an undelegated tests of iis.se with one NS (ns.nic.se) with IP address "91.226.36.450", "91.226.36", "91.226" or "91." it works as it should and a detailed error message is presented.

With IP address "91" there is no errror message and the test starts.  If I do the same on the release version I get "internal server error". Something in the verification has broken.

image

The test is not completed, but an error is returned:

image

blacksponge:
> With IP address "91" there is no errror message and the test starts. If I do the same on the release version I get "internal server error". Something in the verification has broken.

This was probably caught by the old regex in `validate_syntax`, the one used in the schema is a bit less complex. I'll investigate that.

matsduf:
> > With IP address "91" there is no errror message and the test starts. If I do the same on the release version I get "internal server error". Something in the verification has broken.
> 
> This was probably caught by the old regex in `validate_syntax`, the one used in the schema is a bit less complex. I'll investigate that.

It should be check by both regexp and Net::IP, shouldn't it?

blacksponge:
It is check by both, but before this PR it was tested by two regexes and Net::IP. I will change for the other regex.

blacksponge:
@matsduf the ip validation should be better now

matsduf (approved):
@blacksponge, please resolve the conflicts and I will re-approve.

matsduf (approved):

matsduf (dismissed):

docs/API.md
```diff
@@ -482,9 +479,10 @@ Example response:
...
```
mattias-p, line 485: Could you also link to where the default value is defined?

There are more of these that I haven't marked up.

docs/API.md
```diff
@@ -581,9 +596,26 @@ An object with the following properties:
 
 #### `"error"`
 
->
-> TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
->
+* If any parameter fails to be validated an error code of -32602 is returned. The `data` properties contains an array of all error, see [Validation error data].
```
mattias-p, line 599: `fails to be validated` -> `is invalid` maybe?

`data properties` -> `data property`

`all error` -> `all errors`

docs/API.md
```diff
@@ -656,16 +689,50 @@ An object with the following properties:
 
 #### `"result"`
 
-A *test id*. 
+A *test id*.
 
-If the test has been run with the same domain name within an interval of 10 mins (hard coded), 
+If the test has been run with the same domain name within an interval of 10 mins (hard coded),
```
mattias-p, line 694: `mins` -> `minutes`

mattias-p (commented):
I only completed reviewing a small part of this PR. It sure is on the large side.

docs/API.md
mattias-p: I realize now that you didn't introduce the "mins" wording, but if you make the update anyway while you're at it, I'd appreciate it.

matsduf (dismissed):

matsduf:
I suggest that we have this PR merged now. It is very big, and any adjustments needed can be done in new PRs. @mattias-p, can you approve this PR after the last updates?

pnax (commented):
This is quite huge indeed. So far I have only small remarks that can for sure be handled in other PRs.

When trying to test with ".fr", the error message is not related.

```
$ zmb start_domain_test --domain ".fr" | jq '.error.data[0].message'
"The domain name or label is too long"
```

share/fr.po
```diff
@@ -0,0 +1,83 @@
...
+msgstr "Adresse IP invalide"
+
+msgid "Keytag should be a positive integer"
+msgstr "L'indentifiant doit être un entier positif"
+
+msgid "Algorithm should be a positive integer"
+msgstr "L'algorithme doit être un entier positif"
+
+msgid "Digest type should be a positive integer"
+msgstr "Le type d'empreinte doit être un entier positif"
```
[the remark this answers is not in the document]
blacksponge: Yes I am aware of the non-consistency of that translation, I planned to correct that later on.

matsduf (dismissed):

matsduf:
@mattias-p, can you review now?

@blacksponge, please do not change this unless anything major things comes up. This PR is already too complex.

matsduf:
> This is quite huge indeed. So far I have only small remarks that can for sure be handled in other PRs.
> 
> When trying to test with ".fr", the error message is not related.
> 
> ```
> $ zmb start_domain_test --domain ".fr" | jq '.error.data[0].message'
> "The domain name or label is too long"
> ```

This PR cannot solve all validation problems. In https://github.com/zonemaster/zonemaster/pull/942 I try to define the checks of the domain before starting the test.

mattias-p:
When a change is too large to be reviewed in a reliable way I try to split it up into two or more reasonably sized PRs. I really prefer that approach over merging improperly reviewed code and patching the mistakes later. Of course it's more work in the short term, but I believe it's less work in the long term. Would you consider splitting this PR up into bite-sized chunks?

docs/API.md
mattias-p: Can you check this comment again? It looks to me like it was forgotten.

The same wording as this line is introduced in another place as well.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -446,23 +413,8 @@ sub get_test_results {
...
-    }
-    else {
-        if ( !exists $locales{substr $language, 0, 2}{$language} ) {
-            die "Undefined language string: '$language'\n";
-        }
-        $locale = $language;
-    }
-    $locale .= '.UTF-8';
+    # Already validated by json_validate
+    my ($locale, undef) = $self->_get_locale($params);
```
mattias-p, line 417: Have you considered letting the validation layer propagate the normalized values instead of only performing the validation and deferring the normalization until later? In my experience that's a more robust design - you avoid the risk of forgetting to normalize.

lib/Zonemaster/Backend/ErrorMessages.pm
```diff
@@ -0,0 +1,88 @@
+
+package Zonemaster::Backend::ErrorMessages;
+
+
+use Readonly;
+use Locale::TextDomain qw[Zonemaster-Backend];
+
+Readonly my @CUSTOM_MESSAGES_CONFIG => (
```
mattias-p, line 8: Could you document the type of this data structure? Like, what values are allowed and what's the meaning of different data structures.

lib/Zonemaster/Backend/ErrorMessages.pm
```diff
@@ -0,0 +1,88 @@
...
+    {
+        pattern => "/language",
+        config => {
+            string => {
+                pattern => N__ 'Invalid language tag format'
+            }
+        }
+    },
+    {
+        pattern => ".*",
```
mattias-p, line 77: Forgive me for jumping the gun here. I've haven't read through all changes and thought about all the consequences yet, so I don't have a clear view of what I'm looking at. But I'm not getting a good feeling about this general pattern matching feature. Especially with this catchall. It seems really easy to forget making sure that the parameters are appropriately covered. At the same time it seems less straight forward to actually check whether or not they are.

But maybe (hopefully) I'm wrong and this is actually easier to maintain than it looks to me?

mattias-p (commented):
This is only a partial review again. I have some suggestions, some questions and one worry.

docs/API.md
blacksponge: I did forget some occurrences, my mistake

blacksponge:
> Would you consider splitting this PR up into bite-sized chunks?

I can try to split the changes, I suggest 3 PRs:

1. change validation to return an array of error (the most basic change to be used by the frontend)
2. add custom messages support for json validation
3. add translation support to backend

what do you think?

lib/Zonemaster/Backend/ErrorMessages.pm
blacksponge: > But I'm not getting a good feeling about this general pattern matching feature.

I don't really know what else would be a good way to add custom messages support then. The pattern matching is used to provide custom messages on a per-path basis, the inner structure resembles what the JSON::Validator uses to handle error messages. I will provide documentation for the structure.

Otherwise I am open to suggestion if someone see an other solution to implement the feature. What I also thought of what to directly modify the validation module to add something that stores the error message in the JSON schema (by extending [the json schema](https://json-schema.org/understanding-json-schema/reference/schema.html#id3) and the  [schema class](https://metacpan.org/pod/JSON::Validator::Schema#Make-a-new-validation-class) and the validation class). That would work for specific error messages but not generic ones like `Missing properties` or `Wrong property type`, which may ideally require to embed some of the translation logic directly into the module. But that would definitely better, I totally agree that the pattern matching to customize the error message is more a hack or a work around that anything else.

> Especially with this catchall.

This is to handle generic common error messages that can be present for any element in the json object (hence matching all path).

> But maybe (hopefully) I'm wrong and this is actually easier to maintain than it looks to me.

I don't think it is very difficult to maintain, the main drawback would be that the error message are defined in a place away from the schema definition, which can make it hard to track.

lib/Zonemaster/Backend/ErrorMessages.pm
blacksponge: Sure, will do

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -151,20 +157,38 @@ sub get_host_by_name {
     return \@adresses;
 }
 
+sub get_data_from_parent_zone_validate_syntax {
+    my ( $self, $syntax_input ) = @_;
+    my @errors;
+
+    if ( defined $syntax_input->{domain} ) {
+        my ( undef, $dn_syntax ) = $self->_check_domain( $syntax_input->{domain} );
```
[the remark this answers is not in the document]
blacksponge: I am not sure that was not the case before and I don't want to change some behavior here. I am also not sure whether the validation function (jsonrpc_validate) can mutate the request as the routing is handled by the JSON::RPC module.

matsduf:
> > Would you consider splitting this PR up into bite-sized chunks?
> 
> I can try to split the changes, I suggest 3 PRs:
> 
>     1. change validation to return an array of error (the most basic change to be used by the frontend)
> 
>     2. add custom messages support for json validation
> 
>     3. add translation support to backend
> 
> 
> what do you think?

I think that sound like what @mattias-p is asking for (he is on vacation and will not respond for some time).

docs/API.md
```diff
@@ -60,7 +60,7 @@ for the specified method is invalid, an error with code `-32602` is reported.
 For more information on the validation error data format see
 [Validation error data].
 
-All error states that occur after the RPC method has been identified are reported as internal errors with code `-32603`.
+All errors states that occur after the RPC method has been identified are reported as internal errors with code `-32603`.
```
matsduf, line 63: I think that "error states" is more correct.

docs/API.md
```diff
@@ -501,7 +501,7 @@ value `0.0.0.0` if the lookup returned no A or AAAA records.
 
 #### `"error"`
 
-* If any parameter fails to be validated an error code of -32602 is returned. The `data` properties contains an array of all error, see [Validation error data].
+* If any parameter is invalid an error code of -32602 is returned. The `data` property contains an array of all errors, see [Validation error data].
```
matsduf, line 504: Improvement in readability!

docs/API.md
blacksponge: oops, I have replaced  too much.

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -151,20 +157,38 @@ sub get_host_by_name {
...
+    if (@errors) {
+        return {
+            status => 'nok',
+            errors => \@errors,
+            message => 'Syntax validation error'
+        };
+    }
+    else {
+        return { status => 'ok', message => 'Syntax ok' };
+    }
```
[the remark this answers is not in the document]
blacksponge: changed in #820

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -151,20 +157,38 @@ sub get_host_by_name {
     return \@adresses;
 }
 
+sub get_data_from_parent_zone_validate_syntax {
```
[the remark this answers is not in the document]
blacksponge: Actually I haven't, thanks for the tips, changed in #820

matsduf:
@blacksponge, please resolve conflicts.

matsduf:
@blacksponge, is this PR replaced by PR #853? In PR #853 is says that it will fix issue #789. This PR also claims to fix issue #789 besides issue #703. Please update the description of this issue, or close it and update any references.

blacksponge:
Replaced by #853, #891 and #820

PR #797, opened 2021-06-17
Make LANGUAGE section docs easier to navigate

mattias-p:
## Purpose

Today the `LANGUAGE section` documentation is quite long and if you're looking for something specific you may end up having to read all of it. This PR breaks it down into shorter sub-headings with clear titles.

## Context

I originally made these changes a good while ago, but I didn't get around to actually make a PR out of them until now.

## Changes

This PR moves around sections of text and inserts new headings, but it doesn't change any of the original words.

A list from the LANGUAGE.locale is integrated in to another one under `get_language_tags`.

Some text from LANGUAGE.locale and some from `get_language_tags` is integrated into the definition of the language tag data type definition.

## How to test this PR

This PR does not need testing as it only affects documentation.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -393,20 +393,6 @@ If there are two `locale tags` that would give the same short
 `language tag` then that is excluded. E.g. "en_US en_UK" will
 only give "en_US" and "en_UK" as `language tags`.
 
-The following `language tags` are generated using the default configuration:
-* da
-* da_DK
-* en
-* en_US
-* fi
-* fi_FI
-* fr
-* fr_FR
-* nb
-* nb_NO
-* sv
-* sv_SE
-
 Example request:
 ```json
 {
@@ -416,16 +402,22 @@ Example request:
 }
 ```
 
-Example response:
+Example response (as generated using the default configuration):
 ```json
 {
   "jsonrpc": "2.0",
   "id": 1,
   "result": [
+    "da",
+    "da_DK",
     "en",
     "en_US",
+    "fi",
+    "fi_FI",
     "fr",
     "fr_FR",
+    "nb",
+    "nb_NO",
     "sv",
     "sv_SE"
   ]
--- a/docs/API.md
+++ b/docs/API.md
@@ -156,6 +156,41 @@ This parameter is a string that are an IPv4 or IPv6. It's validated with the fol
  - IPv6 : `/^([0-9A-Fa-f]{1,4}:[0-9A-Fa-f:]{1,}(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?)|([0-9A-Fa-f]{1,4}::[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})$/`
 
 
+### Language tag
+
+Basic data type: string
+
+A string of A-Z, a-z and underscores matching the regular expression
+`/^[a-z]{2}(_[A-Z]{2})?$/`.
+
+The set of valid *language tags* is further constrained by the
+[LANGUAGE.locale] property.
+* If the *language tag* is a five character string, it needs to match a *locale
+  tag* in LANGUAGE.locale.
+* If the *language tag* is a two-character string, it needs to match the
+  first two characters of exactly one *locale tag* in LANGUAGE.locale.
+  (So that it is unambiguous which *locale tag* is matched.)
+
+E.g. if LANGUAGE.locale is "en_US en_UK sv_SE", all the valid *language tags*
+are "en_US", "en_UK", "sv_SE" and "sv".
+
+#### Design
+
+The two first characters of the *language tag* are intended to be an
+[ISO 639-1] two-character language code and the optional two last characters
+are intended to be an [ISO 3166-1 alpha-2] two-character country code.
+
+#### Out-of-the box support
+
+A default installation will accept the following *language tags*:
+* `da` or `da_DK` for Danish language.
+* `en` or `en_US` for English language.
+* `fi` or `fi_FI` for Finnish language.
+* `fr` or `fr_FR` for French language.
+* `nb` or `nb_NO` for Norwegian language.
+* `sv` or `sv_SE` for Swedish language.
+
+
 ### Name server
 
 Basic data type: object
@@ -164,6 +199,13 @@ Properties:
 
 * `"ns"`: A *domain name*, required.
 * `"ip"`: An *IP address* (IPv4 or IPv6), optional. (default: unset)
+
+
+### Non-negative integer
+
+Basic data type: number (integer)
+ 
+A non-negative integer is either zero or strictly positive.
 
 
 ### Priority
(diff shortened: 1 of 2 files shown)


docs/API.md
```diff
@@ -382,6 +382,17 @@ An array of *Profile names* in lower case. `"default"` is always included.
...
+The `language tags` are generated from the configured [locale tags].
+Each `locale tag` will generate two `language tags`, a short tag
+equal to the first two letters (usually the same as a language
+code) and a long tag which is equal to the full `locale tag`.
+If "en_US" is the `locale tag` then "en" and "en_US" are the
+`language tags`.
+
+If there are two `locale tags` that would give the same short
+`language tag` then that is excluded. E.g. "en_US en_UK" will
+only give "en_US" and "en_UK" as `language tags`.
```
matsduf, line 394: Lines 385 to 394, in your proposal, should rather be integrated into the language tag section that this section already refers to. You have a reference to "locale tag" but no link.

matsduf (changes requested):
It does improve the readability of the text.

docs/API.md
```diff
@@ -382,6 +382,17 @@ An array of *Profile names* in lower case. `"default"` is always included.
...
+The `language tags` are generated from the [LANGUAGE.locale] property.
+Each `locale tag` in LANGUAGE.locale generates two `language tags`,
+a short tag equal to the first two letters (usually the same as a language
+code) and a long tag which is equal to the full `locale tag`.
+If "en_US" is the `locale tag` then "en" and "en_US" are the
+`language tags`.
+
+If there are two `locale tags` that would give the same short
+`language tag` then that is excluded. E.g. "en_US en_UK" will
+only give "en_US" and "en_UK" as `language tags`.
```
pnax, line 394: Would it be possible to rephrase this part a little bit ?
Maybe something like :
```
If there are two `locale tags` that would give the same short
`language tag`, this tag will then not be reported (or will be excluded)
```

pnax (commented):
This is a good improvement, I like the new `Installation consideration` section.

docs/API.md
mattias-p: I fixed the link and tweaked the language a little bit.

I'm not sure I understand what your mean about integrating. But let me take this moment to explain why I moved these paragraphs here.

One way to to draw the line between the RPCAPI and the configuration is so that the RPCAPI is about "doing" and the configuration is about "being". From this perspective "generating" is a form of doing, and so it belongs in the RPCAPI. The configuration on the other hand just sits there quite inert concerning itself only with being valid and having a meaning. I believe this perspective is a good guiding principle that is fairly easy to understand and apply.

docs/API.md
matsduf:
```
If there are two `locale tags` that would give the same short
`language tag` then the short tag is excluded. E.g. "en_US en_UK" will
only give "en_US" and "en_UK" as `language tags` ("en" is excluded).
```

docs/API.md
matsduf: There are two sections in the API document on language tags. You have put the text under the `get_language_tags`method section. I think it is better to have all details under the "language tag" section and then referring to that.

docs/API.md
mattias-p: How updating the paragraph above as well? E.g. like this:

```
Each `locale tag` in LANGUAGE.locale generates one or two `language tags`.
A long `language tag` identical to the `locale tag` is always generated.
A short `language tag` consisting of the first two letters of the `locale tag`
(usually the same as a language code) may also be generated.
The short tag is generated when there is only one `locale tag` that would
generate this particular short tag (i.e. so that it is unambiguous).

E.g. if LANGUAGE.locale is "en_US en_UK sv_SE", the generated `language tags`
are "en_US", "en_UK", "sv_SE" and "sv".
```

docs/API.md
mattias-p: Oh, I see. Yeah. I'll make an update.

matsduf (approved):

mattias-p:
I added a couple of sub-headings to the *language tag* data type definition. Just like LANGUAGE.locale I thought it was a bit too long.

And I also moved around the *language tag* and *non-negative integer* data type definitions to make them comply with the lexicographic order.

docs/API.md
mattias-p: This section is no more. I've integrated it into the language tag data type definition, and this resulted in a completely different wording.

matsduf (dismissed):

docs/API.md
```diff
@@ -156,6 +156,41 @@ This parameter is a string that are an IPv4 or IPv6. It's validated with the fol
...
+
+The set of valid *language tags* is further constrained by the
+[LANGUAGE.locale] property.
+* If the *language tag* is a five character string, it needs to match a *locale
+  tag* in LANGUAGE.locale.
+* If the *language tag* is a two-character string, it needs to match the
+  first two characters of exactly one *locale tag* in LANGUAGE.locale.
+  (So that it is unambiguous which *locale tag* is matched.)
+
+E.g. if LANGUAGE.locale is "en_US en_UK sv_SE", all the valid *language tags*
```
pnax, line 174: Could you add a link on LANGUAGE.locale ?

docs/API.md
```diff
@@ -391,16 +397,22 @@ Example request:
 }
 ```
 
-Example response:
+Example response (as generated using the default configuration):
```
pnax, line 400: Have you considered the fact that this will imply that we need to update this part when we add a new language support ?

docs/API.md
mattias-p: Not linking here follows Wikipedia's policy on repeated links (since LANGUAGE.locale is already linked a few lines above):
https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#Duplicate_and_repeat_links

I like that guideline. It feels quite measured to me. Transferring their guideline to our context, I'd say "Language tag" passes for a glossary entry.

Thoughts on this?

docs/API.md
mattias-p: @matsduf Would you object if I remove this?

pnax: Not at all.

docs/API.md
pnax: The idea is interesting. However I am wondering if this applies to documentation, as you would like to make it easier for readers to jump around.

docs/Configuration.md
```diff
@@ -195,17 +194,7 @@ English is the Zonemaster default language, but it can be blocked
...
-`language tags` are generated from the `locale tags`. Each
-`locale tag` will generate two `language tags`, a short tag
-equal to the first two letters (usually the same as a language
-code) and a long tag which is equal to the full `locale tag`.
-If "en_US" is the `locale tag` then "en" and "en_US" are the
-`language tags`.
-
-If there are two `locale tags` that would give the same short
-`language tag` then that is excluded. E.g. "en_US en_UK" will
-only give "en_US" and "en_UK" as `language tags`.
```
pnax, line 208: It looks like this paragraph has been removed with commit 0b0d5af2. I think we should keep somewhere the fact that we remove the short _language tag_ when two (or more) _locale tags_ result in the same short tag (ie: "en_US" and "en_GB" will not give "en").

docs/API.md
pnax: I think we lose information by removing it completely (see https://github.com/zonemaster/zonemaster-backend/pull/797#discussion_r655337591).

docs/Configuration.md
mattias-p: Isn't that clear enough from this?

> * If the *language tag* is a two-character string, it needs to match the
>   first two characters of exactly one *locale tag* in LANGUAGE.locale.
>   (So that it is unambiguous which *locale tag* is matched.)
>
>  E.g. if LANGUAGE.locale is "en_US en_UK sv_SE", all the valid *language tags*
>  are "en_US", "en_UK", "sv_SE" and "sv".

pnax: From my point of view, both sentences describe the problem from a different perspective. One is starting from two _locale tags_ giving the same two-character _language tag_, the other one starts with one two-character _language tag_ which will match with the beginning of a _locale tag_. 
I think that to understand the output of `get_language_tags`, we should explain that we don't output the two-character _language tag_ when 2 (or more) _locale tags_ are resulting in the same short _language tag_.

mattias-p: Sure, I'll make a stab at it. I'll adapt @matsduf's suggestion from above to make it work in its new context.

matsduf (dismissed):

matsduf:
@pnax, can you re-review?

docs/API.md
```diff
@@ -386,7 +386,11 @@ An array of *Profile names* in lower case. `"default"` is always included.
 
 ## API method: `get_language_tags`
 
-Returns a list of all valid [*language tags*][language tag].
+Returns the set of valid [*language tags*][Language tag].
+
+> Note: If there are two [*locale tags*][LANGUAGE.locale] in [LANGUAGE.locale]
+> that would give the same [short language tag][Language tag] then the short tag
+> is excluded from the set of valid [*langauge tags*][Language tag].
```
pnax, line 393: small typo: langauge -> language

pnax (commented):
Thanks for the update. I spotted a typo.

matsduf (approved):

pnax (approved):

PR #798, opened 2021-06-17
Properly define properties to use for database fingerprint (deterministic hash)

### LINKED ISSUE #775: Improved criteria for test reuse
I just ran two tests simultaneously. One through the GUI and one through zmtest. Both tests were for the same domain but the testid wasn't reused.

Here are the parameters:
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "client_id": "Zonemaster GUI",
    "client_version": "3.2.2",
    "domain": "example.com",
    "ds_info": [],
    "ipv4": true,
    "ipv6": true,
    "nameservers": [],
    "priority": 10,
    "profile": "default",
    "queue": 0
  }
}
```
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "domain": "example.com",
    "ipv4": true,
    "ipv6": true,
    "priority": 10,
    "profile": "default",
    "queue": 0
  }
}
```
Their representations are different but semantically they are the same. When calculating the "deterministic hash" at least the `client_id` and `client_version` parameters should be disregarded. Also the parameters should be normalized so that you get the same "deterministic hash" when you use the default values no matter if they're given explicitly or implicitly.

pnax:
## Purpose

When we store a test in the database, we also store a fingerprint based on some criteria, that can be used to avoid running the same time in a short period of time.
This PR harmonizes the way we compute the fingerprint so that we can reuse tests that are similar, even if some properties are left out by the client.
This also normalize the value stored in the `params` database field.

## Context

As explained in #775, depending on the client used, some properties are used in the computation of the fingerprint or not. This PR solves this by always using the same properties (falling back to a default value if the property is not set).
Fixes #775

## Changes

* a first commit is only about removing trailing spaces and changing tabs into spaces (839f435)
* update variable name "deterministic hash" into "fingerprint"
* create new methods to normalize the parameters and to compute the fingerprint
* always include the same properties to compute the fingerprint, only the value will differ
* use default value if the property is not already set (`profile: default` and look into the profile to get the value for `ipv4 ` and `ipv6`)
* normalize the `params` object to a minimal set enlarged with extra values given by the client (such as `client_id` and `client_version`)

## Breaking change

Removal of the `config` property from the config file.
(TBH, this breaking change has not been introduced with this PR, however this PR removes the documentation associated with this property)

## How to test this PR

Unit tests are added with this PR. It is also possible to test using the scenario explained in #775 or by using the `get_test_params` API call.

Testing with the following test case allows to make sure the database is properly populated and that the behavior has not been altered (you may need to increase the `age_reuse_previous_test` value in the config while running the following tests).

1. Starting twice the same test should give the same test-id (the diff output should be empty):
    ```
    TESTID1=$(zmb start_domain_test --domain example.com | jq -r '.result')
    TESTID2=$(zmb start_domain_test --domain example.com | jq -r '.result')
    diff <(echo $TESTID1) <(echo $TESTID2)
    ```

2. Adding garbage option should give the same test-id:
    ```
    TESTID3=$(zmb start_domain_test --domain example.com --client-id "zmb" | jq -r '.result')
    diff <(echo $TESTID1) <(echo $TESTID3)
    ```

3. Adding nameservers without ip should give another test-id (the diff should give an error):
    ```
    TESTID4=$(zmb start_domain_test --domain example.com --nameserver a.example.com | jq -r '.result')
    diff <(echo $TESTID1) <(echo $TESTID4)
    ```

4. Adding ip to the nameserver should give another test-id (both diff should fail):
    ```
    TESTID5=$(zmb start_domain_test --domain example.com --nameserver a.example.com:192.0.2.1 | jq -r '.result')
    diff <(echo $TESTID1) <(echo $TESTID5)
    diff <(echo $TESTID4) <(echo $TESTID5)
    ```

5. Wait `age_reuse_previous_test` and run:
    ```
    TESTID6=$(zmb start_domain_test --domain example.com --client-id "my personal client" | jq -r '.result')
    TESTID7=$(zmb start_domain_test --domain example.com --client-id "another client" | jq -r '.result')
    diff <(echo $TESTID6) <(echo $TESTID7)
    ```

6. Finally output each test parameters and check that the values are correct
    ```
    zmb get_test_params --test-id $TESTID1
    zmb get_test_params --test-id $TESTID2
    zmb get_test_params --test-id $TESTID3  # should not contain any "client_id" property
    zmb get_test_params --test-id $TESTID4
    zmb get_test_params --test-id $TESTID5
    zmb get_test_params --test-id $TESTID6  # should have a "client_id" property set to "my personal client"
    zmb get_test_params --test-id $TESTID7  # should have a "client_id" property which is NOT "another client" but "my personal client"
    ```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -575,8 +575,8 @@ An object with the property:
 
 An object with the following properties:
 
-* `"ns_list"`: A list of *name server* objects representing the nameservers of the given *domain name*.
-* `"ds_list"`: A list of *DS info* objects representing delegated signer of the given *domain name*.
+* `"ns_list"`: A list of [*name server*][Name server] objects representing the nameservers of the given *domain name*.
+* `"ds_list"`: A list of [*DS info*][DS info] objects representing delegation signer (DS record data) of the given *domain name*.
 
 
 #### `"error"`
@@ -639,11 +639,11 @@ Example response:
 An object with the following properties:
 
 * `"domain"`: A *domain name*, required. The zone to test.
-* `"ipv6"`: A boolean, optional. (default `true`). Used to configure the test and enable IPv4 tests.
-* `"ipv4"`: A boolean, optional. (default `true`). Used to configure the test and enable IPv6 tests.
-* `"nameservers"`: A list of *name server* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile.
+* `"ipv6"`: A boolean, optional. (default: [`net.ipv4`][net.ipv4] profile value). Used to enable or disable testing over IPv4 transport protocol.
+* `"ipv4"`: A boolean, optional. (default: [`net.ipv6`][net.ipv6] profile value). Used to enable or disable testing over IPv6 transport protocol.
+* `"nameservers"`: A list of [*name server*][Name server] objects, optional. (default: `[]`). Used to perform un-delegated test.
+* `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`). Used to perform un-delegated test.
+* `"profile"`: A *profile name*, optional. (default: `"default"`). Run the tests using the given profile.
 * `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_id"`: A *client id*, optional. (default: unset). Used to monitor which client uses the API.
 * `"client_version"`: A *client version*, optional. (default: unset). Used to monitor which client use the API
@@ -1045,13 +1045,13 @@ An object with the following properties:
 The value of `"test_params"` is an object with the following properties:
 
 * `"client_id"`: A *client id*, optional. (default: unset)
-* `"profile"`: A *profile name*, optional (default `"default"`). Run the tests using the given profile.
+* `"profile"`: A *profile name*, optional (default: `"default"`). Run the tests using the given profile.
 * `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_version"`: A *client version*, optional. (default: unset)
-* `"nameservers"`: A list of *name server* objects, optional. (default: `[]`)
-* `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`)
-* `"ipv4"`: A boolean, optional. (default: `true`)
-* `"ipv6"`: A boolean, optional. (default: `true`)
+* `"nameservers"`: A list of [*name server*][Name server] objects, optional. (default: `[]`)
+* `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`)
+* `"ipv6"`: A boolean, optional. (default: [`net.ipv4`][net.ipv4] profile value).
+* `"ipv4"`: A boolean, optional. (default: [`net.ipv6`][net.ipv6] profile value).
 * `"priority"`: A *priority*, optional. (default: `5`)
 * `"queue"`: A *queue*, optional. (default: `0`)
 
@@ -1196,8 +1196,12 @@ The `"params"` object sent to `start_domain_test` or `add_batch_job` when the *t
 >
 
 [Available profiles]:           Configuration.md#profiles-section
+[DS info]:                      #ds-info
 [ISO 3166-1 alpha-2]:           https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2
 [ISO 639-1]:                    https://en.wikipedia.org/wiki/ISO_639-1
 [LANGUAGE.locale]:              Configuration.md#locale
 [Language tag]:                 #language-tag
+[Name server]:                  #name-server
+[net.ipv4]:                     https://metacpan.org/pod/Zonemaster::Engine::Profile#net.ipv4
+[net.ipv6]:                     https://metacpan.org/pod/Zonemaster::Engine::Profile#net.ipv6
 [Privilege levels]:             #privilege-levels
--- a/docs/API.md
+++ b/docs/API.md
@@ -807,8 +807,8 @@ In the case of a test created with `start_domain_test`:
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued.
 * `"id"`: An integer.
 * `"hash_id"`: A *test id*. The *test* in question. 
-* `"params"`: The `"params"` object sent to `start_domain_test` when the *test*
-  was started.
+* `"params"`: A normalized version `"params"` object sent to
+  `start_domain_test` when the *test* was started.
 * `"results"`: A list of *test result* objects.
 
 
@@ -816,9 +816,10 @@ In the case of a test created with `add_batch_job`:
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued.
 * `"id"`: An integer.
 * `"hash_id"`: A *test id*. The *test* in question. 
-* `"params"`: The `"params"` object sent to `start_domain_test` when the *test*
-  was started.
-* `"results"`: the result is a list of *test id* corresponding to each tested domain.
+* `"params"`: A normalized version `"params"` object sent to `add_batch_job`
+  when the *test* was started.
+* `"results"`: the result is a list of *test id* corresponding to each tested
+  domain.
 
 >
 > TODO: Change name in the API of `"hash_id"` to `"test_id"`
@@ -1129,7 +1130,7 @@ An object with the following properties:
 
 ## API method: `get_test_params`
 
-Return all *params* objects of a *test*.
+Return a normalized *params* objects of a *test*.
 
 Example request:
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -644,7 +644,7 @@ An object with the following properties:
 * `"nameservers"`: A list of *name server* objects, optional. (default: `[]`). Used to perform un-delegated test.
 * `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`). Used to perform un-delegated test.
 * `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated**. A string, optional. Ignored. Specify `"profile"` instead.
+* `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_id"`: A *client id*, optional. (default: unset). Used to monitor which client uses the API.
 * `"client_version"`: A *client version*, optional. (default: unset). Used to monitor which client use the API
 * `"priority"`: A *priority*, optional. (default: `10`)
@@ -1046,7 +1046,7 @@ The value of `"test_params"` is an object with the following properties:
 
 * `"client_id"`: A *client id*, optional. (default: unset)
 * `"profile"`: A *profile name*, optional (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated.** A string, optional. Ignored. Specify profile instead.
+* `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_version"`: A *client version*, optional. (default: unset)
 * `"nameservers"`: A list of *name server* objects, optional. (default: `[]`)
 * `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`)
--- a/docs/API.md
+++ b/docs/API.md
@@ -644,7 +644,6 @@ An object with the following properties:
 * `"nameservers"`: A list of [*name server*][Name server] objects, optional. (default: `[]`). Used to perform un-delegated test.
 * `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`). Used to perform un-delegated test.
 * `"profile"`: A *profile name*, optional. (default: `"default"`). Run the tests using the given profile.
-* `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_id"`: A *client id*, optional. (default: unset). Used to monitor which client uses the API.
 * `"client_version"`: A *client version*, optional. (default: unset). Used to monitor which client use the API
 * `"priority"`: A *priority*, optional. (default: `10`)
@@ -754,7 +753,6 @@ Example response:
       "domain": "zonemaster.net",
       "profile": "default",
       "ipv6": true,
-      "advanced": true,
       "nameservers": [
         {
           "ns": "ns3.nic.se",
@@ -865,13 +863,11 @@ Example response:
       "id": "c45a3f8256c4a155",
       "creation_time": "2016-11-15 11:53:13.965982",
       "overall_result": "error",
-      "advanced_options": null
     },
     {
       "id": "32dd4bc0582b6bf9",
       "creation_time": "2016-11-14 08:46:41.532047",
       "overall_result": "error",
-      "advanced_options": null
     },
     ...
   ]
@@ -1046,7 +1042,6 @@ The value of `"test_params"` is an object with the following properties:
 
 * `"client_id"`: A *client id*, optional. (default: unset)
 * `"profile"`: A *profile name*, optional (default: `"default"`). Run the tests using the given profile.
-* `"config"`: **No longer supported**. Use `"profile"` instead.
 * `"client_version"`: A *client version*, optional. (default: unset)
 * `"nameservers"`: A list of [*name server*][Name server] objects, optional. (default: `[]`)
 * `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`)
@@ -1157,7 +1152,6 @@ Example response:
          "domain": "zonemaster.net",
          "profile": "default",
          "client_id": "Zonemaster Dancer Frontend",
-         "advanced": true,
          "nameservers": [
             {
                 "ns": "ns3.nic.se",
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -209,6 +209,7 @@ sub _normalize_parameter_hash {
     my $array_nameservers = $$params{nameservers} // [];
     for my $nameserver (@$array_nameservers) {
         $$nameserver{ip} = "" if ( not defined $$nameserver{ip} );
+        $$nameserver{ns} = lc $$nameserver{ns};
     }
     my @array_nameservers_sort = sort {
         $a->{ip} cmp $b->{ip} or
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -189,7 +189,7 @@ sub generate_fingerprint {
     my $profile = Zonemaster::Engine::Profile->effective;
 
     my %to_encode = ();
-    $to_encode{domain}      = $$params{domain}      // "";
+    $to_encode{domain}      = lc $$params{domain}   // "";
     $to_encode{ds_info}     = $$params{ds_info}     // [];
     $to_encode{ipv4}        = $$params{ipv4}        // $profile->get( 'net.ipv4' );
     $to_encode{ipv6}        = $$params{ipv6}        // $profile->get( 'net.ipv6' );
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -190,13 +190,34 @@ sub generate_fingerprint {
 
     my %to_encode = ();
     $to_encode{domain}      = lc $$params{domain}   // "";
-    $to_encode{ds_info}     = $$params{ds_info}     // [];
     $to_encode{ipv4}        = $$params{ipv4}        // $profile->get( 'net.ipv4' );
     $to_encode{ipv6}        = $$params{ipv6}        // $profile->get( 'net.ipv6' );
-    $to_encode{nameservers} = $$params{nameservers} // [];
     $to_encode{priority}    = $$params{priority}    // 10;
     $to_encode{profile}     = $$params{profile}     // "default";
     $to_encode{queue}       = $$params{queue}       // 0;
+
+    my $array_ds_info = $$params{ds_info} // [];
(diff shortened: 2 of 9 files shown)


lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,30 @@ sub process_unfinished_tests {
     }
 }
 
+sub generate_deterministic_hash {
```
matsduf, line 136: 1. Now that a new method is added to this module, it should also be documented with POD. Please compare with e.g. Config.pm
2. The name of the method is misleading. It actually does two things:
   1. Normalize the parameter hash
   2. Create a finger print (hash) of the normalized parameter hash

The two operations do not need to coexist in the same method, and I suggest that the method is broken up into two methods, which will make it easier to understand the code.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,30 @@ sub process_unfinished_tests {
...
+    my $profile = Zonemaster::Engine::Profile->effective;
+
+    my %to_encode = ();
+    $to_encode{domain}      = lc $$params{domain}   // "";
+    $to_encode{ds_info}     = $$params{ds_info}     // [];
+    $to_encode{ipv4}        = $$params{ipv4}        // $profile->get( 'net.ipv4' );
+    $to_encode{ipv6}        = $$params{ipv6}        // $profile->get( 'net.ipv6' );
+    $to_encode{nameservers} = $$params{nameservers} // [];
+    $to_encode{priority}    = $$params{priority}    // 10;
+    $to_encode{profile}     = $$params{profile}     // "default";
```
matsduf, line 148: The API documentation should be then be updated to state that "default" will be used if unspecified.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,30 @@ sub process_unfinished_tests {
     }
 }
 
+sub generate_deterministic_hash {
+    my ( $self, $params ) = @_;
+
+    my $profile = Zonemaster::Engine::Profile->effective;
+
+    my %to_encode = ();
+    $to_encode{domain}      = lc $$params{domain}   // "";
```
matsduf, line 142: 1. Is it possible to come here with an undefined "domain"?
2. I am not sure that this is the correct place in the code to do normalization (down casing) of the domain name. Normalization of domain name is a bigger issue with IDN and should reside in a central place such as RPCAPI.pm

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,30 @@ sub process_unfinished_tests {
...
+    $to_encode{ds_info}     = $$params{ds_info}     // [];
+    $to_encode{ipv4}        = $$params{ipv4}        // $profile->get( 'net.ipv4' );
+    $to_encode{ipv6}        = $$params{ipv6}        // $profile->get( 'net.ipv6' );
+    $to_encode{nameservers} = $$params{nameservers} // [];
+    $to_encode{priority}    = $$params{priority}    // 10;
```
matsduf, line 147: This already gets a default value in RCP.pm. Needed here? -- The same applies for "queue" below.

lib/Zonemaster/Backend/DB.pm
pnax: > It actually does two things

Good point, I'll review my approach

lib/Zonemaster/Backend/DB.pm
pnax: > Is it possible to come here with an undefined "domain"?

Not sure, I need to investigate.

> I am not sure that this is the correct place in the code to do normalization (down casing) of the domain name.

I totally agree with you. However, we should start somewhere and I think this is a good place before moving it in an upper layer.

lib/Zonemaster/Backend/DB.pm
pnax: I choose to define a default value for all entries because I tried to be cautious. I need to make some tests and look deeper into the code to check which values are already set or not when reaching this part of the code.

lib/Zonemaster/Backend/DB.pm
pnax: > Is it possible to come here with an undefined "domain"?

Indeed, this is not possible. Same for `priority` and `queue`. I'll remove the default value.

pnax: I choose to keep the default value for now as I need it in the unit tests. This is not the best as this duplicate the value. I haven't find a proper solution so far. I'll think about it.

lib/Zonemaster/Backend/DB.pm
pnax: see https://github.com/zonemaster/zonemaster-backend/pull/798#discussion_r655516858

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,84 @@ sub process_unfinished_tests {
...
+            defined $a->{ip} and
+            defined $b->{ip} and
+            $a->{ip} cmp $b->{ip}
+        ) or
+        (
+            defined $a->{ns} and
+            defined $b->{ns} and
+            $a->{ns} cmp $b->{ns}
+        )
+    } @$array_nameservers;
```
matsduf, line 187: A nameserver object can be both `{"nameservers":[{"ns":"ns.nic.se","ip":""}` and `{"nameservers":[{"ns":"ns.nic.se"}`. Will your code normalize that? -- Maybe that is too ambitious.

mattias-p: Empty `ip` and omitted `ip` mean the same thing, right? In that case we should be normalized.

mattias-p:
Would you consider using the term "fingerprint" instead of "deterministic hash" when updating or adding new code? I realize the term "deterministic hash" isn't your idea, but I think we should move away from it and you're updating these parts anyway.

pnax:
> Would you consider using the term "fingerprint" instead of "deterministic hash" when updating or adding new code?

Good idea. Should we consider updating the database column name as well (if yes, this can be done in another PR) ?

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,84 @@ sub process_unfinished_tests {
...
+    $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
+    $normalized{profile}  = $$params{profile}   // "default";
+    $normalized{priority} = $$params{priority}  // 10;
+    $normalized{queue}    = $$params{queue}     // 0;
+
+    my $array_ds_info = $$params{ds_info} // [];
+    my @array_ds_info_sort = sort {
+        (
+            defined $a->{algorithm} and
+            defined $b->{algorithm} and
```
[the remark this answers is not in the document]
pnax: I added these checks everywhere because I got some warning on undefined values. Instead of looking for what should be defined or not when reaching this part, I decided to do the checks for all the keys indifferently.

mattias-p:
I plan to make a couple of PRs that move the database initialization from stand-alone scripts and into the database adapter modules among other things. I think it's probably better to defer updating the database schema until after that.

Also, when we update the schema we have to provide a migration for updating existing installations. And next time we do it I think we should also introduce a version number in the schema.

lib/Zonemaster/Backend/DB.pm
pnax: Done with 10a5dca.

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
pnax: This has been removed as those values are required as described in the [API.md document](https://github.com/zonemaster/zonemaster-backend/blob/v6.2.0/docs/API.md#ds-info).

lib/Zonemaster/Backend/DB.pm
pnax: It looks like the API already specifies the default values for all params.
https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/API.md#params-2

The `ipv4` and `ipv6` need to be updated though because we now accept unset values (and set them accordingly to the config).

docs/API.md
```diff
@@ -567,8 +567,8 @@ An object with the property:
 
 An object with the following properties:
 
-* `"ns_list"`: A list of *name server* objects representing the nameservers of the given *domain name*.
-* `"ds_list"`: A list of *DS info* objects representing delegated signer of the given *domain name*.
+* `"ns_list"`: A list of [*name server*][Name server] objects representing the nameservers of the given *domain name*.
+* `"ds_list"`: A list of [*DS info*][DS info] objects representing delegated signer of the given *domain name*.
```
matsduf, line 571: ...delegation signer (DS record data) of the given...

docs/API.md
```diff
@@ -631,12 +631,11 @@ Example response:
...
 
 * `"domain"`: A *domain name*, required. The zone to test.
-* `"ipv6"`: A boolean, optional. (default `true`). Used to configure the test and enable IPv4 tests.
-* `"ipv4"`: A boolean, optional. (default `true`). Used to configure the test and enable IPv6 tests.
-* `"nameservers"`: A list of *name server* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated**. A string, optional. Ignored. Specify `"profile"` instead.
+* `"ipv6"`: A boolean, optional. (default: [`net.ipv4`][net.ipv4] profile value). Used to configure the test and enable IPv4 tests.
+* `"ipv4"`: A boolean, optional. (default: [`net.ipv6`][net.ipv6] profile value). Used to configure the test and enable IPv6 tests.
```
matsduf, line 635:
```
Used to enable or disable testing over IPv4 transport.
```

lib/Zonemaster/Backend/DB.pm
```diff
@@ -130,6 +133,79 @@ sub process_unfinished_tests {
...
+
+    my $profile = Zonemaster::Engine::Profile->effective;
+
+    my %normalized = ();
+    $normalized{domain}   = lc $$params{domain} // "";
+    $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
+    $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
+    $normalized{profile}  = $$params{profile}   // "default";
+    $normalized{priority} = $$params{priority}  // 10;
+    $normalized{queue}    = $$params{queue}     // 0;
```
matsduf, line 147: Both "priority" and "queue" are set in RPCAPI.pm. They should not be set twice.

pnax:
I rebased again (I moved one commit up in the history, the one about removing a deprecated parameter).
I also refactored some code while rebasing. And I addressed some of your remarks at the same time (@matsduf).

lib/Zonemaster/Backend/DB.pm
pnax: Well, to be honest there is also a duplication in each SQL schema as those values are the default for their respective column.
I agree that there are too many duplicates of these values. However, I need them in this new method for the unit tests.
I haven't removed the initialization in the RPCAPI.pm file because I feel that even though they have the same value, they don't exactly have the same meaning (it is an initialization instead of a fall back to some default values).

I started some refactoring of all these values duplication.

docs/API.md
pnax: updated.

mattias-p:
Could you add a clear note to the PR description about the removal of deprecated parameters? It's important that we list all such removals in the release notes.

pnax:
> Could you add a clear note to the PR description about the removal of deprecated parameters? It's important that we list all such removals in the release notes.

Yes totally. Just a remark, the `config` property was not a valid property since the validation of the config file (see 1b84528 and [Config.pm](https://github.com/mattias-p/zonemaster-backend/blob/1b8452823c9d93ed2a52e545212adde216c05505/lib/Zonemaster/Backend/Config.pm#L222..L231))

pnax:
I think I missed something about deprecation. The documentation can have statements on deprecated properties. This can mean that the code has been updated (we no longer allow this property) or not (we allow it but it might disappear in a near future).

With this back in mind, I think we can either
* keep the documentation with the deprecated statement
* or update the _deprecated_ value to something stronger (I think there is an issue on how to deal with deprecation)
* or just remove the statement

I can totally revert my commit if you think we should keep the deprecation documented.

matsduf:
> I think I missed something about deprecation. The documentation can have statements on deprecated properties. This can mean that the code has been updated (we no longer allow this property) or not (we allow it but it might disappear in a near future).

As we have used it, deprecation means that it is allowed but it is a warning that it will be removed.

pnax:
> As we have used it, deprecation means that it is allowed but it is a warning that it will be removed in a near future.

If this is the case, then the property is not deprecated anymore since we die if it is present in the config file.
So I'll keep the documentation removal and update the PR description.

matsduf:
> > As we have used it, deprecation means that it is allowed but it is a warning that it will be removed in a near future.
> 
> If this is the case, then the property is not deprecated anymore since we die if it is present in the config file.
> So I'll keep the documentation removal and update the PR description.

If we think it is important to keep the information in the ducumentation we could something like "no longer supported", but in this case it should be be removed.

pnax:
> If we think it is important to keep the information in the ducumentation we could something like "no longer supported", but in this case it could just be removed.

TBH I don't know what is better either. I think it is good to keep documented that this is no longer supported. But I am wondering if this is really needed and not making our documentation heavier.

pnax:
To make Travis run, I rebased once more with a change: I added back the documentation of the "config" property and updated it to "no longer supported".
I also fixed a typo in SQLite.pm

matsduf:
> To make Travis run

I added an empty commit to another PR to make Travis start a test.

pnax:
I made the choice to always define the "ip" field in the "nameservers" object. However I found issue https://github.com/zonemaster/zonemaster-gui/issues/176 that suggests to remove the field if it is empty. I'll update this PR to remove the "ip" property if its value is empty.

docs/API.md
```diff
@@ -631,12 +631,12 @@ Example response:
...
-* `"nameservers"`: A list of *name server* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"ds_info"`: A list of *DS info* objects, optional. (default: `[]`). Used to perform un-delegated test.
-* `"profile"`: A *profile name*, optional. (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated**. A string, optional. Ignored. Specify `"profile"` instead.
+* `"ipv6"`: A boolean, optional. (default: [`net.ipv4`][net.ipv4] profile value). Used to enable or disable testing over IPv4 transport protocol.
+* `"ipv4"`: A boolean, optional. (default: [`net.ipv6`][net.ipv6] profile value). Used to enable or disable testing over IPv6 transport protocol.
+* `"nameservers"`: A list of [*name server*][Name server] objects, optional. (default: `[]`). Used to perform un-delegated test.
+* `"ds_info"`: A list of [*DS info*][DS info] objects, optional. (default: `[]`). Used to perform un-delegated test.
+* `"profile"`: A *profile name*, optional. (default: `"default"`). Run the tests using the given profile.
+* `"config"`: **No longer supported**. Use `"profile"` instead.
```
matsduf, line 639: In this case I think we can just remove it.

docs/API.md
```diff
@@ -1037,13 +1037,13 @@ An object with the following properties:
 The value of `"test_params"` is an object with the following properties:
 
 * `"client_id"`: A *client id*, optional. (default: unset)
-* `"profile"`: A *profile name*, optional (default `"default"`). Run the tests using the given profile.
-* `"config"`: **Deprecated.** A string, optional. Ignored. Specify profile instead.
+* `"profile"`: A *profile name*, optional (default: `"default"`). Run the tests using the given profile.
+* `"config"`: **No longer supported**. Use `"profile"` instead.
```
matsduf, line 1041: In this case I think we can just remove it.

lib/Zonemaster/Backend/DB.pm
matsduf: It could be OK to have these values here if it otherwise create other problems, but then it should be documented in the code that the values are also set in RPCAPI.pm. In a second step (later PR) it could be cleaned up.

matsduf (changes requested):
In API.pm under 
```
## API method: `get_test_results`
```
it says
```
* `"params"`: The `"params"` object sent to `start_domain_test` when the *test*
  was started.
```
which is not true if this PR is merged. Then it will be the normalized *params* object.

matsduf (changes requested):
```
$ make distcheck
Not in MANIFEST: t/db.t
```

matsduf:
The implementation works as expected.

pnax:
> which is not true if this PR is merged. Then it will be the normalized params object.

Good point.

I also realized that I am currently removing the `client_id` and `client_version` properties. Do you think these properties have value in the database? Is there any use case for these values? Are they used for debug or something else?
With the current PR's code, these properties won't be stored anymore. This is maybe a breaking change. I think I'll review this part to keep them stored in database (but leave them out in the computation of the fingerprint).

matsduf:
> I also realized that I am currently removing the `client_id` and `client_version` properties. Do you think these properties have value in the database? Is there any use case for these values? Are they used for debug or something else?
> With the current PR's code, these properties won't be stored anymore. This is maybe a breaking change. I think I'll review this part to keep them stored in database (but leave them out in the computation of the fingerprint).

I do not see any value of the `client_id` and  `client_version` properties in a production setting, but in  development situation where you update or develop client it could have some value. I think they should be kept as-is, but left out of the computation of the fingerprint.

docs/API.md
pnax: Removed in 52dfb17.

docs/API.md
pnax: Removed in 52dfb17.

pnax:
* I added the new test file to the MANIFEST
* I updated the logic to store the `client_id` and the `client_version` in database without using it to compute the fingerprint (this means that if someone uses a different client, it might get the result of a test created from another client, so the fields `client_id` and `client_version` sent out to a client might differ from the one used to start the test)
* I added a test in t/test01.t so that this can be tested against each database engine easily
* I updated the PR description with one more test scenario (where you update the `client_id` to see it being stored in database)

@matsduf and @mattias-p please re-review, I think I have addressed all the remarks so far.

docs/API.md
```diff
@@ -801,18 +799,19 @@ In the case of a test created with `start_domain_test`:
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued.
 * `"id"`: An integer.
 * `"hash_id"`: A *test id*. The *test* in question. 
-* `"params"`: The `"params"` object sent to `start_domain_test` when the *test*
-  was started.
+* `"params"`: A normalized `"params"` object sent to `start_domain_test` when
+  the *test* was started.
```
matsduf, line 803:
```
A normalized version of the `"params"` object sent to `start_domain_test` when
  the *test* was started.
```

docs/API.md
```diff
@@ -801,18 +799,19 @@ In the case of a test created with `start_domain_test`:
...
 
 In the case of a test created with `add_batch_job`:
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued.
 * `"id"`: An integer.
 * `"hash_id"`: A *test id*. The *test* in question. 
-* `"params"`: The `"params"` object sent to `start_domain_test` when the *test*
-  was started.
-* `"results"`: the result is a list of *test id* corresponding to each tested domain.
+* `"params"`: A normalized `"params"` object sent to `start_domain_test` when
+  the *test* was started.
```
matsduf, line 812: It should be `add_batch_job`, shouldn't it?

```
A normalized version of the `"params"` object sent to `add_batch_job` when
  the *test* was started.
```

lib/Zonemaster/Backend/DB.pm
matsduf: Is it documented somewhere that priority and queue get default values in two places of the code?

matsduf (changes requested):

lib/Zonemaster/Backend/DB.pm
pnax: Oops, I forgot this one. I'll do it.

docs/API.md
pnax: Done in 558f9c6.

docs/API.md
pnax: Yes, fixed in 558f9c6

matsduf:
@pnax, please resolve the conflicts between this PR and the branch and I can re-review and approve.

pnax:
I fixed the conflicts while rebasing on top of _develop_ once again.
@matsduf, please re-review.

matsduf (approved):

lib/Zonemaster/Backend/DB.pm
```diff
@@ -179,6 +183,97 @@ sub _new_dbh {
...
+    my %normalized = ();
+
+    # some of these values are already set in RPCAPI
+    # however setting them here again is required for testing purpose
+    $normalized{domain}   = lc $$params{domain} // "";
+    $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
+    $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
+    $normalized{profile}  = $$params{profile}   // "default";
+    $normalized{priority} = $$params{priority}  // 10;
+    $normalized{queue}    = $$params{queue}     // 0;
```
matsduf, line 200: @pnax, on a second thought, I do not see why "priority" and "queue" should be part of the fingerprint. It should only be those parameters that can affect the outcome. I do not see how those could affect the outcome.

pnax: This is a pretty fair point. When I thought about those values, I thought it might avoid mixing results from different queues. But I guess this is not strong enough to support the fact that we use them in the computation of the fingerprint. I guess, we might remove them safely.

PR #799, opened 2021-06-18
Refactor locale validation

mattias-p:
## Purpose

This PR refactors the implementation of LANGUAGE.locale.

## Context

This PR extends the use of patterns for API and validation introduced in #759 and #757 to also cover the LANGUAGE.locale property.

These changes were developed in parallel with #759 and #757 as part of resolving #685. These changes are not part for solving in a technical sense, because the validations were already there. But they are in an architectural sense because they make the Config module interface more cohesive. These changes also played an important role when designing the new Config module interface, its internal structure and the new validation framework.

## Changes

The methods Language_Locale_hash() and ListLanguageTags() are replaced by the LANGUAGE_locale() getter.
Call sites are updated to use the new getter instead.

The internal representation of the LANGUAGE.locale property within the Config is updated to reflect a little more of its structure as described in the configuration documentation.

The assignment of the default value for LANGUAGE.locale is placed together with the other properties.

## How to test this PR

This PR is a pure refactoring so it should not affect any functionality, and it does add a few new tests at the unit level. That said it modifies code used by the `get_language_tags` and `get_test_results` RPC-API methods and we don't have any automatic tests for those.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -187,7 +187,7 @@ sub parse {
     }
     if ( defined( my $value = $ini->val( 'LANGUAGE', 'locale' ) ) ) {
         if ( $value eq "" ) {
-            push @warnings, "Use of empty LANGUAGE.locale propery is deprecated. Remove the LANGUAGE.locale entry or specify LANGUAGE.locale = en_US instead.";
+            push @warnings, "Use of empty LANGUAGE.locale property is deprecated. Remove the LANGUAGE.locale entry or specify LANGUAGE.locale = en_US instead.";
         }
     }
     if ( defined( my $value = $get_and_clear->( 'ZONEMASTER', 'number_of_professes_for_frontend_testing' ) ) ) {
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -187,7 +187,7 @@ sub parse {
     }
     if ( defined( my $value = $ini->val( 'LANGUAGE', 'locale' ) ) ) {
         if ( $value eq "" ) {
-            push @warnings, "Use of empty LANGUAGE.locale propery is deprecated. Remove LANGUAGE.locale entry or specify LANUGAGE.locale = en_US instead.";
+            push @warnings, "Use of empty LANGUAGE.locale propery is deprecated. Remove the LANGUAGE.locale entry or specify LANGUAGE.locale = en_US instead.";
         }
     }
     if ( defined( my $value = $get_and_clear->( 'ZONEMASTER', 'number_of_professes_for_frontend_testing' ) ) ) {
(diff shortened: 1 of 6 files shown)


lib/Zonemaster/Backend/Config.pm
```diff
@@ -253,16 +254,13 @@ sub parse {
...
-        $locale_tag =~ /^[a-z]{2}_[A-Z]{2}$/
-          or die "Illegal locale tag in LANGUAGE.locale: $locale_tag\n";
-
-        !exists $obj->{_LANGUAGE_locale}{$locale_tag}
-          or die "Repeated locale tag in LANGUAGE.locale: $locale_tag\n";
-
-        $obj->{_LANGUAGE_locale}{$locale_tag} = 1;
+    if ( defined( my $value = $get_and_clear->( 'LANGUAGE', 'locale' ) ) ) {
+        if ( $value ne "" ) {
+            $obj->_clear_LANGUAGE_locale();
```
pnax, line 259: It looks like you are emptying the hash before populating it. If this is really the case, I think the name `_clear_LANGUAGE_locale` is conflicting with `get_and_clear` (which removes an entry from the ini file), both uses the word "clear". What do you think about renaming the method to something like `_reset_LANGUAGE_locales` ?

mattias-p: Good point. Fixing.

pnax (approved):
Looks good to me (and Travis has already been run on the last commit, and is happy)

pnax (approved):
looks good, beside 2 small typos.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -184,6 +185,11 @@ sub parse {
         $obj->_set_SQLITE_database_file( $value )
           if $obj->DB_engine eq 'SQLite';
     }
+    if ( defined( my $value = $ini->val( 'LANGUAGE', 'locale' ) ) ) {
+        if ( $value eq "" ) {
+            push @warnings, "Use of empty LANGUAGE.locale propery is deprecated. Remove LANGUAGE.locale entry or specify LANUGAGE.locale = en_US instead.";
```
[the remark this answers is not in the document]
matsduf: The documentation says: "Leaving the locale key empty or absent is deprecated. Always configure it with supported locale tags." Having no LANGUAGE.locale is deprecated.

matsduf, line 190: The documentation says: "Leaving the locale key empty or absent is deprecated. Always configure it with supported locale tags." Having no LANGUAGE.locale is deprecated.

matsduf (changes requested):

lib/Zonemaster/Backend/Config.pm
mattias-p: I removed the deprecation of absent LANGAUGE.locale. That way we don't have to force users who don't care about native language support to add this property to their configuration files when we finally remove the feature. Having a default value of "en_US" is perfectly reasonable.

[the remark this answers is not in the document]
mattias-p: Fixed

matsduf (approved):

mattias-p:
I've tested this v2021.2 and it LGTM.

I've run:
* `zmb get_language_tags` a couple of times with different LANGUAGE.locale settings
* `zmtest` both with and without the --lang option
Both seems to work fine.

PR #801, opened 2021-06-21
Validate profiles

### LINKED ISSUE #808: Private profiles do not work
RPCAPI will ignore private profiles. There is no code to check the profile in `start_domain_test` against the list of private profiles.

Scenario 1:
```
[PUBLIC PROFILES]
default-no-ipv6=/usr/local/etc/zonemaster/profile-no-ipv6.json
default=/usr/local/etc/zonemaster/profile.json
test=/usr/local/etc/zonemaster/profile-test1.json

[PRIVATE PROFILES]
#test=/usr/local/etc/zonemaster/profile-test1.json
```
```
# zmb --verbose start_domain_test --domain sunet.se --profile test
{"id":1,"method":"start_domain_test","jsonrpc":"2.0","params":{"profile":"test","domain":"sunet.se"}}
{"result":"5d94caf2d74d76f1","id":1,"jsonrpc":"2.0"}
```
Scenario 2:
```
[PUBLIC PROFILES]
default-no-ipv6=/usr/local/etc/zonemaster/profile-no-ipv6.json
default=/usr/local/etc/zonemaster/profile.json
#test=/usr/local/etc/zonemaster/profile-test1.json

[PRIVATE PROFILES]
test=/usr/local/etc/zonemaster/profile-test1.json
```
```
# zmb --verbose start_domain_test --domain sunet.se --profile test
{"params":{"profile":"test","domain":"sunet.se"},"jsonrpc":"2.0","id":1,"method":"start_domain_test"}
{"jsonrpc":"2.0","error":{"code":-32603,"message":"Internal error 009 \n"},"id":1}
```

mattias-p:
## Purpose

This PR mainly does two things:
* It implements the validation rules already described the PUBLIC PROFILES and PRIVATE PROFILES sections.
* It fixes a bug that prevented the use of profiles in the PRIVATE PROFILES section. (#808)

## Context

This PR currently contains all the commits from #802. After that PR is merged this one will be rebased onto the develop branch.

This is the last leg of #685.

Fixes #808.

## Changes

This PR contains lots of tiny changes that update the same parts of the code. To understand the changes I recommend going through the changes commit by commit. When adding review comments, keep in mind that lines changed in an earlier commit may well be updated in a later commit.

* It adds getters for the validation of the PUBLIC PROFILES and PRIVATE PROFILES sections.
* It simplifies the internal representation of the profiles settings in the Config and RPCAPI modules.
* It moves some profile merging operation to load-time as opposed to performing them for each request.
* It adds unit tests for the validation rules.

### Verify that it's possible to specify a private profile in start_domain_test

Create an empty profile: `echo {} > /tmp/empty.profile`

Update the PRIVATE PROFILES section in your backend_config.ini:
```
[PRIVATE PROFILES]
private = /tmp/empty.profile
```

Restart zm-rpcapi and zm-testagent.

Run: `zmtest --profile private example.com`

Make sure that you get a test result.

### Verify that both zm-rpcapi and zm-testagent refuse to start if there is an invalidly named profile in the configuration

Update the PRIVATE PROFILES section in your backend_config.ini:
```
[PRIVATE PROFILES]
-invalid-name- = /tmp/empty.profile
```
Restart zm-rpcapi and zm-testagent and make sure that they refuse to come up. Also make sure that a log entries are written containing details of the errors.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -767,7 +767,19 @@ sub _reset_LANGUAGE_locale {
 sub _add_public_profile {
     my ( $self, $name, $path ) = @_;
 
+    $name = untaint_profile_name( $name )    #
+      // die "Invalid profile name in PUBLIC PROFILES section: $name\n";
+
     $name = lc $name;
+
+    if ( defined $self->{_public_profiles}{$name} || exists $self->{_private_profiles}{$name} ) {
+        die "Profile name not unique: $name\n";
+    }
+
+    if ( defined $path ) {
+        $path = untaint_abs_path( $path )    #
+          // die "Path must be absolute for profile: $name\n";
+    }
 
     $self->{_public_profiles}{$name} = $path;
     return;
@@ -776,7 +788,21 @@ sub _add_public_profile {
 sub _add_private_profile {
     my ( $self, $name, $path ) = @_;
 
+    $name = untaint_profile_name( $name )    #
+      // die "Invalid profile name in PRIVATE PROFILES section: $name\n";
+
     $name = lc $name;
+
+    if ( $name eq 'default' ) {
+        die "Profile name must not be present in PRIVATE PROFILES section: $name\n";
+    }
+
+    if ( exists $self->{_public_profiles}{$name} || exists $self->{_private_profiles}{$name} ) {
+        die "Profile name not unique: $name\n";
+    }
+
+    $path = untaint_abs_path( $path )    #
+      // die "Path must be absolute for profile: $name\n";
 
     $self->{_private_profiles}{$name} = $path;
     return;
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -100,6 +100,8 @@ sub parse {
     my ( $class, $text ) = @_;
 
     my $obj = bless( {}, $class );
+    $obj->{_public_profiles}  = {};
+    $obj->{_private_profiles} = {};
 
     my $ini = Config::IniFiles->new( -file => \$text )
       or die "Failed to parse config: " . join( '; ', @Config::IniFiles::errors ) . "\n";
@@ -132,6 +134,7 @@ sub parse {
     $obj->_set_ZONEMASTER_lock_on_queue( '0' );
     $obj->_set_ZONEMASTER_age_reuse_previous_test( '600' );
     $obj->_add_LANGUAGE_locale( 'en_US' );
+    $obj->_add_public_profile( 'default', undef );
 
     # Assign property values (part 1/2)
     if ( defined( my $value = $get_and_clear->( 'DB', 'engine' ) ) ) {
@@ -269,15 +272,14 @@ sub parse {
         }
     }
 
-    $obj->{_public_profiles} = {
-        default => undef,
-    };
     for my $name ( $ini->Parameters( 'PUBLIC PROFILES' ) ) {
-        $obj->{_public_profiles}{lc $name} = $get_and_clear->( 'PUBLIC PROFILES', $name );
+        my $path = $get_and_clear->( 'PUBLIC PROFILES', $name );
+        $obj->_add_public_profile( $name, $path );
     }
-    $obj->{_private_profiles} = {};
+
     for my $name ( $ini->Parameters( 'PRIVATE PROFILES' ) ) {
-        $obj->{_private_profiles}{lc $name} = $get_and_clear->( 'PRIVATE PROFILES', $name );
+        my $path = $get_and_clear->( 'PRIVATE PROFILES', $name );
+        $obj->_add_private_profile( $name, $path );
     }
 
     # Check required propertys (part 2/2)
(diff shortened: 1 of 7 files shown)


matsduf:
@mattias-p, is this ready for review?

mattias-p:
If you like you can start looking through the diff. I expect the final version to look a lot like the current one. But I haven't tested it myself since the last changes so don't bother reporting any malfunctions.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -269,15 +272,14 @@ sub parse {
         }
     }
 
-    $obj->{_public_profiles} = {
-        default => undef,
-    };
     for my $name ( $ini->Parameters( 'PUBLIC PROFILES' ) ) {
-        $obj->{_public_profiles}{lc $name} = $get_and_clear->( 'PUBLIC PROFILES', $name );
+        my $path = $get_and_clear->( 'PUBLIC PROFILES', $name );
```
matsduf, line 276: Will this remove duplicates? On purpose?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -269,15 +272,14 @@ sub parse {
...
+        $obj->_add_public_profile( $name, $path );
     }
-    $obj->{_private_profiles} = {};
+
     for my $name ( $ini->Parameters( 'PRIVATE PROFILES' ) ) {
-        $obj->{_private_profiles}{lc $name} = $get_and_clear->( 'PRIVATE PROFILES', $name );
+        my $path = $get_and_clear->( 'PRIVATE PROFILES', $name );
```
matsduf, line 281: Will this remove duplicates? On purpose?

matsduf (changes requested):
The code will not discover duplicate profile definitions:
```
[PUBLIC PROFILES]
default-no-ipv6=/usr/local/etc/zonemaster/profile-no-ipv6.json
default=/usr/local/etc/zonemaster/profile-test3.json
test=/usr/local/etc/zonemaster/profile-test1.json
test=/usr/local/etc/zonemaster/profile-test2.json
```

lib/Zonemaster/Backend/Config.pm
mattias-p: It's Config::IniFiles that removes duplicates. I can't say if it's on purpose but I'd prefer if it didn't.

mattias-p:
> The code will not discover duplicate profile definitions

We need to replace out Config::IniFiles to fix that. Do you want to create a new issue about that?

matsduf:
> > The code will not discover duplicate profile definitions
> 
> We need to replace out Config::IniFiles to fix that. Do you want to create a new issue about that?

I understand, it is a limitation in Config::IniFiles. As I read `_add_public_profile` it tried to discover duplicates within "PUBLIC".

I think this can pass -- if it is documented. I guess this applies to all keys in the configuration file. Can you just add in Configuration.pm that duplicates keys in the same section will not be discovered and that it is not determined which valued that will be used? Or something?

mattias-p:
Let's say we specify the current behavior regarding duplicated entries. And then in a later release we decide to fix it so that duplicated entries are forbidden. This is a breaking change since we made promises about our behavior by specifying it.

On the other hand if we leave the behavior regarding duplicated entries unspecified, it isn't a breaking change to fix it because we're not breaking any promises.

I prefer the latter approach better because it affords us more room to avoid breaking changes.

matsduf:
> I prefer the latter approach better because it affords us more room to avoid breaking changes.

I think that is a bad approach. Instead it is better to say that duplicate keys are not permitted by specification but currently not prevented by the implementation, but that might change in the future without notice.

We should not promise that duplicates are permitted, but warn that having duplicates can create problem. When we know the limitation we should document it.

By the way, is that true for duplicate sections too?

mattias-p:
I had a closer look and it seems I was wrong. It is possible to get duplicated values out from Config::IniFiles, I just wasn't aware of it. I implemented a check to forbid repeated config properties.

matsduf (dismissed):
Great that duplicates can be captured.

mattias-p:
@pnax, @blacksponge Can I get another review?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -295,9 +295,10 @@ sub validate_syntax {
         }
 
         if ( defined $syntax_input->{profile} ) {
-            my @profiles = map lc, $self->{config}->ListPublicProfiles();
-            return { status => 'nok', message => encode_entities( "Invalid profile option format" ) }
-            unless ( grep { $_ eq lc $syntax_input->{profile} } @profiles );
+            my %profiles = ( $self->{config}->PUBLIC_PROFILES, $self->{config}->PRIVATE_PROFILES );
+            if ( !exists $profiles{ $syntax_input->{profile} } ) {
```
blacksponge, line 299: if the profile name is normalized in the config shouldn't the profile parameter be normalized as well?

pnax:
Sure, I'll have a look.

pnax (commented):
Beside the remark from @blacksponge, this looks good to me.

lib/Zonemaster/Backend/RPCAPI.pm
mattias-p: Your right. I fixed it.

blacksponge (approved):
lgtm

matsduf:
> lgtm

What is "lgtm"?

blacksponge:
> what is "lgtm"?

looks good to me

matsduf:
The limitation of permitted characters in a profile name and the fact that they are compared disregarding case is defined in the API document, but it is in the configuration file they are set. In the Configuration.md there is no word about it.

The API only permits profiles defined in the configuration file, and it is in the configuration that the limitations sit. At least there should be a reference from Configuration.md to the relevant section of API.md.

matsduf (approved):

pnax:
## Testing v2021.2

LGTM

PR #810, opened 2021-07-06
Corrects zmb

matsduf:
## Purpose

For RPCAPI method get_test_history zmb has a non-supported parameter "nameservers" and a missing parameter "filter". This PR corrects that.

This PR also clarifies a cryptic error message.

Before this PR from `zmb man`:
```
   get_test_history
        zmb [GLOBAL OPTIONS] get_test_history [OPTIONS]

        Options:
          --domain DOMAIN_NAME
          --nameserver true|false|null
          --offset COUNT
          --limit COUNT
```
Before this PR:
```
# zmb --verbose get_test_history --domain xb --nameserver true | jq .
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "get_test_history",
  "params": {
    "frontend_params": {
      "nameservers": true,
      "domain": "xb"
    }
  }
}
{
  "id": 1,
  "error": {
    "code": "-32602",
    "data": "/frontend_params: Properties not allowed: nameservers."
  },
  "jsonrpc": "2.0"
}
```

Issue #812 is related to this issue.

## Changes

This PR removes the incorrect parameter and implements the missing "filter" parameter. It also clarifies an unrelated error message.

## How to test this PR

* Run `get_test_history` for some domain with the incorrect "nameservers" parameter and make sure that the program outputs an error.
* Run `get_test_history` for some domain with the "filter" parameter. Use value "all". The other values work, but not fully correctly, in MySQL, but currently not in SQLite.
* Run `get_test_history` for some domain with the "filter" parameter. Use value "nisse" (illegal) to get an error message directly from `zmb`.
* Run `start_domain_test` with some domain and the parameter "ipv6" with the incorrect value "nisse" to get the clarified error message directly from `zmb`.

### DIFF
--- a/script/zmb
+++ b/script/zmb
@@ -464,7 +464,7 @@ sub json_tern {
         return undef;
     }
     else {
-        die "unknown ternary value";
+        die 'Illegal value. Expects "true", "false" or "null" ';
     }
 }
 
--- a/script/zmb
+++ b/script/zmb
@@ -368,7 +368,7 @@ sub cmd_get_test_history {
     );
 
     if ( $opt_filter ) {
-        unless ( $opt_filer =~ /^(?:all|delegated|undelegated)$/ ) {
+        unless ( $opt_filter =~ /^(?:all|delegated|undelegated)$/ ) {
             die 'Illegal filter value. Expects "all", "delegated" or "undelegated" ';
         }
         $params{filter} = $opt_filter;
--- a/script/zmb
+++ b/script/zmb
@@ -340,7 +340,7 @@ sub cmd_get_test_results {
 
  Options:
    --domain DOMAIN_NAME
-   --nameserver true|false|null
+   --filter all|delegated|undelegated
    --offset COUNT
    --limit COUNT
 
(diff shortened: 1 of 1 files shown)


script/zmb
```diff
@@ -367,8 +367,11 @@ sub cmd_get_test_history {
         },
     );
 
-    if ( $opt_nameserver ) {
-        $params{frontend_params}{nameservers} = json_tern( $opt_nameserver );
+    if ( $opt_filter ) {
+        unless ( $opt_filter =~ /^(?:all|delegated|undelegated)$/ ) {
+            die 'Illegal filter value. Expects "all", "delegated" or "undelegated" ';
```
mattias-p, line 372: Is the trailing space intentional?

mattias-p (approved):

script/zmb
matsduf: Yes, since there is no LF `die` will add a reference to the place in the code, but now I see that it also adds a space before that. I can remove the trailing space, but I guess that it really does not matter.

mattias-p: With an extra space we get a clearer separation between the message and the line reference. I don't believe I've seen that before but I think it's a nice feature.

matsduf: Yes, you are right.
```
# ~matsd/zmb --verbose get_test_history --domain iis.se --filter deleg | jq .
Illegal filter value. Expects "all", "delegated" or "undelegated"  at /home/matsd/zmb line 372.
```

PR #814, opened 2021-07-08
Corrects the description of the reuse time

matsduf:
## Purpose

The API document says that the reuse time is hard coded, which it is not. It also gives  an incorrect description of what parameters that are taken into account when comparing two requests.

## Changes

Description is corrected.

## How to test this PR

This PR only updates a document, not the implementation.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -658,8 +658,15 @@ An object with the following properties:
 
 A *test id*. 
 
-If the test has been run with the same domain name within an interval of 10 mins (hard coded), 
-then the new request does not trigger a new test, but returns with the results of the last test
+If a test has been run with the same parameters (as listed below) not more than
+"reuse time" ago, then a new request will not trigger a new test. Instead the
+`test id` of the previous test will be returned. The default value of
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -658,8 +658,15 @@ An object with the following properties:
 
 A *test id*. 
 
-If the test has been run with the same domain name within an interval of 10 mins (hard coded), 
-then the new request does not trigger a new test, but returns with the results of the last test
+If a test has been run with the same parameters (as listed below) not more than
```
mattias-p, line 661: There's an ambiguity here between when a test is requested and when it is started. I think "requested" would be a better word than "run".

mattias-p (approved):

docs/API.md
matsduf: I have updated. @mattias-p, please re-review.

mattias-p (approved):

matsduf:
@blacksponge, can you review this?

blacksponge (approved):

matsduf:
@blacksponge, I just resolved the conflict. Can you approve again?

blacksponge (approved):

PR #821, opened 2021-07-12
Custom error message

blacksponge:
## Purpose

Provide more human friendly error messages than the ones generated by the JSON validator.

## Context

#789

## Changes

The json validation error are mapped to a set of custom defined messages by using pattern matching on the `path`. The error is replaced with the one corresponding to the first `path` that matches. This is because the module we use for validation does not allow custom error messages (and internalization, in anticipation of #823 )  to be easily defined as it is not a specification of the JSON schema. See comment https://github.com/zonemaster/zonemaster-backend/pull/795#discussion_r667109704 and the following one for additional context .

Custom error messages implemented (with source of message):
* Domain names (`/domain`, `/nameserver/*/ns`)
  - The domain name is not a valid IDNA string and cannot be converted to an A-label (`_check_domain`)
  - The domain name contains non-ascii characters and IDNA is not installed (`_check_domain` and `json_schemas`)
  - The domain name character(s) are not supported (`_check_domain`)
* IPs ( `/nameserver/*/ip`)
  - Invalid IP address (`validate_syntax` for parsing check, `json_schemas` for pattern check)
* DS key tag (`/ds_info/*/keytag`)
  - Keytag should be a positive integer (`json_schemas` for type check and min(0) check)
* DS digest data (`/ds_info/*/digest`)
  - Invalid digest format (`json_schemas` for pattern check (length and characters allowed))
* Profile (`/profile`)
  - Unknown profile (`validate_syntax`)
* Language (`/language`)
  - Invalid language tag format
  - Language string not unique
  - Unkown language string
* Generic
  - Missing property

## How to test this PR
`curl -sH "Content-Type: application/json" --data '{"method":"start_domain_test","params":{"domain":"toto.com", "nameservers":[{"ip":"a"}], "ds_info":[{"algorithm":-1,"digest":"abc","digtype":-1,"keytag":"abc"}]},"id":1,"jsonrpc":"2.0"}' localhost:5000/api  | jq`
```json
{
  "id": 1,
  "error": {
    "data": [
      {
        "message": "Algorithm should be a positive integer",
      },
      {
        "path": "/ds_info/0/digest",
        "message": "Invalid digest format"
      },
      {
        "path": "/ds_info/0/digtype",
        "message": "Digest type should be a positive integer"
      },
      {
        "path": "/ds_info/0/keytag",
        "message": "Keytag should be a positive integer"
      },
      {
        "path": "/nameservers/0/ip",
        "message": "Invalid IP address"
      },
      {
        "path": "/nameservers/0/ns",
        "message": "Missing property"
      },
      {
        "path": "/nameservers/0/ns",
        "message": "Domain name required"
      },
      {
        "message": "Invalid IP address",
        "path": "/nameservers/0/ip"
      }
    ],
    "code": "-32602",
  },
  "jsonrpc": "2.0"
}
```


lib/Zonemaster/Backend/ErrorMessages.pm
```diff
@@ -0,0 +1,96 @@
...
+#  hash are the types of error.
+# This structure is similar to the one used internally in JSON::Validator,
+#  see https://github.com/jhthorsen/json-validator/blob/master/lib/JSON/Validator/Error.pm
+
+Readonly my @CUSTOM_MESSAGES_CONFIG => (
+    {
+        pattern => "/(domain|hostname)",
+        config => {
+            string => {
+                pattern => 'The domain name character(s) are not supported'
```
matsduf, line 20: 'The domain name contains a character or characters not supported'

lib/Zonemaster/Backend/ErrorMessages.pm
```diff
@@ -0,0 +1,96 @@
...
+            string => {
+                pattern => 'Invalid IP address'
+            }
+        }
+    },
+    {
+        pattern => "/nameservers/\\d+/ns",
+        config => {
+            string => {
+                pattern => 'The domain name character(s) are not supported'
```
matsduf, line 36: 'The name server name contains a character or characters not supported'

lib/Zonemaster/Backend/ErrorMessages.pm
```diff
@@ -0,0 +1,96 @@
...
+                minimum => 'Algorithm should be a positive integer'
+            }
+        }
+    },
+    {
+        pattern => "/ds_info/\\d+/digtype",
+        config => {
+            integer => {
+                type => 'Digest type should be a positive integer',
+                minimum => 'Digest type should be a positive integer'
```
matsduf, line 63: Why not "must" in all three cases?

matsduf (changes requested):

lib/Zonemaster/Backend/ErrorMessages.pm
blacksponge: This is better indeed

lib/Zonemaster/Backend/ErrorMessages.pm
blacksponge: done

lib/Zonemaster/Backend/ErrorMessages.pm
blacksponge: done

matsduf (dismissed):

blacksponge:
@mattias-p as you had already some thought about this from https://github.com/zonemaster/zonemaster-backend/pull/795#discussion_r667109704 , could you review this PR when you can so we can move forward with this feature?

mattias-p:
I have a suggestion for a completely different approach that I think has some merit. I feel that there is a strong coupling between the validation rules and the custom validation messages. For this reason I think they should be defined close together and be defined in a structurally similar way. That way it's easier to check which validation rules have custom messages and which ones are lacking. It's also harder to forget to add the custom message when you add a new validation rule, and you don't even have to think about it when adding a new parameter to a method with a path pattern that hasn't been used before.

The approach I'm suggesting is to appropriate the [description field](https://metacpan.org/pod/JSON::Validator#Basics) to store the custom error messages. I haven't fully explored the consequences of this approach but I can only see one potential downside. I'm not sure if it's possible to specify descriptions using the JSON::Validator::Joi interface, so maybe we'd have to migrate away from it, and I haven't checked how much convenience it's buying us. Maybe it's a lot. Maybe it's negligible.

Here's a POC for extracting the description value:

```perl
use JSON::Validator;
my $jv = JSON::Validator->new;

$jv->schema(
    {
        type       => "object",
        required   => [ "firstName", "lastName" ],
        properties => {
            firstName => { type => "string" },
            lastName  => { type => "string" },
            age       => { type => "integer", minimum => 0, description => "Must be a positive integer" }
        }
    }
);

my @errors = $jv->validate( { firstName => "Jan Henning", lastName => "Thorsen", age => -42 } );

for my $err ( @errors ) {
    my $message = $jv->get( "/properties" . $err->path )->{description};
    printf qq({"path": "%s", "message": "%s"}\n), $err->path, $message;
}
```

blacksponge:
Oh I must have missed the `description` field, I will make some tests to see if we could leverage that. I like this approach too, it would be much cleaner.

blacksponge:
As the changes were totally unrelated to the work here I have made an other PR implementing your idea: https://github.com/zonemaster/zonemaster-backend/pull/853

matsduf:
If this RP is replaced by #853, then this could be closed.

matsduf:
@blacksponge, has this been replaced by #853? If so, close it. Else, can you resolve the conflict and review the PR?

blacksponge:
> has this been replaced by #853?

If the design proposed in #853 is accepted, this one could be closed. I'd like to have @mattias-p thoughts on #853 as he was the one who asked for it.

blacksponge:
Replaced by #853

PR #822, opened 2021-07-13
Add table of contents to API and Configuration documents

matsduf:
## Purpose

The API.md and Configuration.md are both complex and hard to navigate.

## Changes

In both documents a table of contents have been added. Only two levels have been included in the table.

## How to test this PR

Nothing to test. Only editorial documentation update.


docs/API.md
```diff
@@ -1,5 +1,49 @@
...
+* [API method: get_data_from_parent_zone](#API-method-get_data_from_parent_zone)
+* [API method: start_domain_test](#API-method-start_domain_test)
+* [API method: test_progress](#API-method-test_progress)
+* [API method: get_test_results](#API-method-get_test_results)
+* [API method: get_test_history](#API-method-get_test_history)
+* [API method: add_api_user](#API-method-add_api_user)
+* [API method: add_batch_job](#API-method-add_batch_job)
+* [API method: get_batch_job_result](#API-method-get_batch_job_result)
+* [API method: get_test_params](#API-method-get_test_params)
+
```
pnax, line 45: https://github.com/zonemaster/zonemaster-backend/commit/e903cab0c143ff0e1c742c864be0c59994cb2dfa adds a new 2nd level heading "Undelegated and delegated". Should this be included in the toc? Or should the level of this new heading be 3 (###)?

docs/Configuration.md
```diff
@@ -1,5 +1,42 @@
...
+  * [port](#port)
+  * [user](#user)
+  * [password](#password)
+  * [database](#database)
+* [POSTGRESQL section](#POSTGRESQL-section)
+  * [host](#host)
+  * [port](#port)
+  * [user](#user)
+  * [password](#password)
+  * [database](#database)
```
pnax, line 24: each anchor should be suffixed
```
  * [host](#host-1)
  * [port](#port-1)
  * [user](#user-2)
  * [password](#password-2)
  * [database](#database-1)
```

docs/Configuration.md
```diff
@@ -1,5 +1,42 @@
...
+  * [user](#user)
+  * [password](#password)
+  * [database_host](#database_host)
+  * [database_name](#database_name)
+  * [polling_interval](#polling_interval)
+* [MYSQL section](#MYSQL-section)
+  * [host](#host)
+  * [port](#port)
+  * [user](#user)
+  * [password](#password)
```
pnax, line 17: missing suffix to avoid conflicts with the previous headings:
```
  * [user](#user-1)
  * [password](#password-1)
```

pnax (commented):
This makes it easier to move around each file indeed!

docs/API.md
matsduf: It should be level 3, and included. I will rebase this and undate. Good catch!

docs/Configuration.md
matsduf: Good that you checked and found this.

docs/Configuration.md
matsduf: And again. The model is a little bit clumsy. I fix all this.

matsduf:
@pnax, please rereview updated and corrected PR. And rebased.

docs/API.md
pnax: This is then the only level 3 heading in the toc. Is this ok if we do not include all other level 3 headings? (For me this is fine, but just asking if we want to be consistent)

matsduf: All level 3 headings are included. There are several level 3 headings at the start of the document. All other headings under the methods are level 4 headings (which is a little strange since there are no level 3 headings).

Or would it be better to make it a level 4 heading like the other headings under the methods?

pnax: Oops, you're right. I thought the `params`, `results`, ... were also level 3 headings. I think this is fine as it is now.

pnax (approved):

PR #826, opened 2021-07-14
Correct the handling of delegated and undelegated in historic tests

### LINKED ISSUE #812: Delegated/undelegated is incorrectly implemented in Backend
The field "undelegated" is supposed to hold the value 1 for an undelegated test and 0 for a normal test.

The value is supposed to be set indirectly by `start_domain_test` but for PostgreSQL it is not done. For SQLite and MySLQ wrong value is inserted. The value is always greater than 0.

With `get_test_history` it is possible to select to get all or only one type of tests with the "filter" parameter. The implementation in MySQL is correct, but since the value in the database is wrong it does not help. The implementations in PostgrSQL and SQLite, respectively, are wrong, and unless "filter" is set to "all" an error will be returned.

PR #810 is related to this issue.

matsduf:
## Purpose

The feature to separate delegated from undelegated tests (and vice versa) when listing historic tests does not work due to incorrect implementation.

## Context

Fixes #812.

## Changes

* Creates a clear specification in the API documentation of the meaning of "delegated" vs "undelegated".
* Creates a shared method in DB.pm that calculates if the test is delegated or undelegated.
* Updates the implementation in SQLite.pm and MySQL.pm to make it work and follow the specification. Also for tests started with `add_batch_job`.
* Updates the database schema for PostgreSQL to follow the other database engines and to make it possible implement all engines with the same structure.  
* Updates the implementation in PostgreSQL.pm to make it work and follow the specification. Also for tests started with `add_batch_job`.
* Adds instructions and script for updating a PostgreSQL database.

Since this PR changes the schema of the PostgreSQL database it should be consider to be a breaking change.

## How to test this PR

* Create tests, with and without name server or DS information. Use both `start_domain_test` and `add_batch_job`.
* Verify with `get_test_history` and the filter option that the tests are correctly classified.
* Repeat for all three database engines.
* For PostgreSQL, create the database before the change and use the patch script to update. Test before and after patching.


share/patch_postgresql_db_zonemaster_backend_ver_7.0.0.pl
```diff
@@ -0,0 +1,27 @@
...
+    die "The configuration file does not contain the PostgreSQL backend";
+}
+my $dbh = Zonemaster::Backend::DB::PostgreSQL->from_config( $config )->dbh;
+
+sub patch_db {
+
+    ####################################################################
+    # TEST RESULTS
+    ####################################################################
+    $dbh->do( 'ALTER TABLE test_results ADD COLUMN undelegated integer NOT NULL DEFAULT 0' );
```
blacksponge, line 23: Why is the `undelegated` column of type integer here and of type boolean in the creation script?

blacksponge: Also I see we are marking all previous tests as delegated even if they were not, is that okay?

matsduf: It should be integer in both cases. I missed to change in the creation script. Corrected now.

Good observation that all previous tests will classified as delegated. For previous tests in MySQL or SQLite databases, all tests are stored as undelegated even when they are not.

One option is to let it be as is since this classification has not been used. Another is to create a conversion script for all three database engines to rewrite the tests to make the classification to be correct.

What do you think?

blacksponge: I think that it would be nice to have the correct data in database, but I understand that it requires some work to make the update scripts.

You made me realize that the GUI does not allow to filter for undelegated / delegated tests (or else I missed it), is that something we could want? (And could make the update scripts worth writing). Should I open an issue on the GUI regarding that feature?

matsduf: > I think that it would be nice to have the correct data in database, but I understand that it requires some work to make the update scripts.

I agree to both points. I am not sure that it is worth the effort, but we could think about it before the release. When this PR is merged, I will create an issue that lifts this up and we can decide later.

> You made me realize that the GUI does not allow to filter for undelegated / delegated tests (or else I missed it), is that something we could want? (And could make the update scripts worth writing). Should I open an issue on the GUI regarding that feature?

That was my plan to do after this is merged, but go ahead! My idea is as follows:
* In the normal case, the undelegated tests are unwanted because the are really not comparable with any other test unless the domain name and DS settings are equal.
* By default only delegated tests are to be shown, with a buttons saying "all historic tests" and "only undelegated tests".
* I was thinking if the result JSON of method "get_test_history" should contain the delegated value to make it possible for the GUI to swap between the selections.

blacksponge: > When this PR is merged, I will create an issue that lifts this up and we can decide later.

Okay this is fine.

> I was thinking if the result JSON of method "get_test_history" should contain the delegated value to make it possible for the GUI to swap between the selections.

I think that could simplify the GUI implementation indeed.

I created this issue from your comments https://github.com/zonemaster/zonemaster-gui/issues/251 so the discussion can move there if needed.

matsduf:
@blacksponge, if you think this is good to go, please approve and I will merge.

After merging I will create two issues:
* Possibly create scripts for updating existing databases with correct values in the "undelegated" column.
* Possibly update the "get_test_history" method to return "undelegated" key with value in the JSON data.

blacksponge (approved):

matsduf:
@blacksponge, please rereview. I have added unit test.

blacksponge (approved):

matsduf:
>     * Possibly create scripts for updating existing databases with correct values in the "undelegated" column.
#829

>     * Possibly update the "get_test_history" method to return "undelegated" key with value in the JSON data.

#830

Issue #211, opened 2016-11-10
Create a config_file parameter allowing a backend to execute only tests of a certain priority


mtoma:
This is a temporary workaround until a proper queue+priority is implemented

mattias-p:
Could you update this issue to explicitly state the following?
* the name of the option
* under what section it's located
* its allowed values
* whether it's supposed to be optional or required
* possibly what default value to expect

matsduf:
What the backend is about to have is queue handling and make it possible to let different workers handle different queue. I see no reason why this should be temporary. And let us call this "queue" not "priority". That is less confusing for the users. I know that the field in the database is called "priority".

mtoma:
You are probably right Mats, I'll call this queue and add the another column to the database.

PR #831, opened 2021-07-26
Remove "priority" and "queue" parameters from fingerprint computation

### LINKED ISSUE #815: Remove queue and priority from fingerprint computation
PR #798 implements a fingerprint computation in lib/Zonemaster/Backend/DB.pm for comparison between a `start_domain_test` request and an already executed test. "queue" and "priority" have been included, which is not correct.

See https://github.com/zonemaster/zonemaster-backend/pull/798#pullrequestreview-702064076

pnax:
## Purpose

This PR removes the use of `queue` and `priority` in the fingerprint computation.

## Context

Fixes #815 

## Changes

Revert the use of `priority` and `queue` to compute the fingerprint added in #798.

**Remark**: this value is still kept in the JSON string stored into the `params` column.

## How to test this PR

1. create a test using `add_batch_job`:
    ```
    BATCHID=$(zmb add_batch_job --username zmb --api-key zmb --domain example.com | jq -r .result)
    ```
2. create a test using `start_domain_test` for the same domain:
    ```
    TESTID=$(zmb start_domain_test --domain example.com | jq -r .result)
    ```
3. wait for the batch job to finish:
    ```
    zmb get_batch_job_result --batch-id $BATCHID
    ```
4. check that `$TESTID` is the same as the id in the batch job
    ```
    diff <(echo $TESTID) <(zmb get_batch_job_result --batch-id $BATCHID | jq -r .result.finished_test_ids[0])
    ```

**Remark**: testing the other way around (starting with a single test and then the batch job) will result in the creation of 2 different tests. Currently there is no logic to reuse a previous test within the batch job.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -691,8 +691,8 @@ the `test id` of the previous test will be returned. The default value of
 key in the configuration file.
 
 The parameters that are compared when to determine if two requests are to be
-considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`
-and `profile`.
+considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`,
+`profile` and `queue`.
 
 
 #### `"error"`
--- a/docs/API.md
+++ b/docs/API.md
@@ -691,8 +691,8 @@ the `test id` of the previous test will be returned. The default value of
 key in the configuration file.
 
 The parameters that are compared when to determine if two requests are to be
-considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`,
-`profile` and `queue`.
+considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`
+and `profile`.
 
 
 #### `"error"`
--- a/docs/API.md
+++ b/docs/API.md
@@ -691,8 +691,8 @@ the `test id` of the previous test will be returned. The default value of
 key in the configuration file.
 
 The parameters that are compared when to determine if two requests are to be
-considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`,
-`profile`, `priority` and `queue`.
+considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`
+and `profile`.
 
 
 #### `"error"`
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -196,6 +196,7 @@ sub _normalize_params {
     $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
     $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
     $normalized{profile}  = $$params{profile}   // "default";
+    $normalized{queue}    = $$params{queue}     // 0;
 
     my $array_ds_info = $$params{ds_info} // [];
     my @array_ds_info_sort = sort {
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -190,13 +190,10 @@ sub _project_params {
 
     my %projection = ();
 
-    # some of these values are already set in RPCAPI
-    # however setting them here again is required for testing purpose
     $projection{domain}   = lc $$params{domain} // "";
     $projection{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
     $projection{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
     $projection{profile}  = $$params{profile}   // "default";
-    $projection{queue}    = $$params{queue}     // 0;
 
     my $array_ds_info = $$params{ds_info} // [];
     my @array_ds_info_sort = sort {
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -183,20 +183,20 @@ sub _new_dbh {
     return $dbh;
 }
 
-sub _normalize_params {
+sub _project_params {
     my ( $self, $params ) = @_;
 
     my $profile = Zonemaster::Engine::Profile->effective;
 
-    my %normalized = ();
+    my %projection = ();
 
     # some of these values are already set in RPCAPI
     # however setting them here again is required for testing purpose
-    $normalized{domain}   = lc $$params{domain} // "";
-    $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
-    $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
-    $normalized{profile}  = $$params{profile}   // "default";
-    $normalized{queue}    = $$params{queue}     // 0;
+    $projection{domain}   = lc $$params{domain} // "";
+    $projection{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
+    $projection{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
+    $projection{profile}  = $$params{profile}   // "default";
+    $projection{queue}    = $$params{queue}     // 0;
 
     my $array_ds_info = $$params{ds_info} // [];
     my @array_ds_info_sort = sort {
@@ -206,7 +206,7 @@ sub _normalize_params {
         $a->{keytag}    <=> $b->{keytag}
     } @$array_ds_info;
 
-    $normalized{ds_info} = \@array_ds_info_sort;
+    $projection{ds_info} = \@array_ds_info_sort;
 
     my $array_nameservers = $$params{nameservers} // [];
     for my $nameserver (@$array_nameservers) {
@@ -220,9 +220,9 @@ sub _normalize_params {
         ( defined $a->{ip} and defined $b->{ip} and $a->{ip} cmp $b->{ip} )
     } @$array_nameservers;
 
-    $normalized{nameservers} = \@array_nameservers_sort;
+    $projection{nameservers} = \@array_nameservers_sort;
 
-    return \%normalized;
+    return \%projection;
 }
 
 sub _params_to_json_str {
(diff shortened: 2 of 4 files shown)


matsduf:
> **Remark**: testing the other way around (starting with a single test and then the batch job) will result in the creation of 2 different tests. Currently there is no logic to reuse a previous test within the batch job.

I have noted that too. I do not know if that is on purpose or that is by accident. I cannot think of any good reason why there should be a difference.

matsduf (dismissed):

pnax:
> I do not know if that is on purpose or that is by accident. I cannot think of any good reason why there should be a difference.

Me neither.
However I noted that if you add thousands of tests in a batch job, this is a bottleneck if you want to run a single test on one of the domain.
Say you are running a batch on the whole domains from the .fr zone. Then all the domains are queued in the database. Now you want to test "afnic.fr". It is already in database, so you get its id. But there is a high probability that the test hasn't been run yet. So your only option is to wait `ZONEMASTER.age_reuse_previous_test` seconds before being able to add the test with a higher priority (because on its own). And then you'll get your results before the end of the batch job.

So far I don't know how to deal with this use case.

pnax:
What I wanted to express is the following use case:

P1 and P2 use the same database.
P1 has enough rights to run batch jobs. P1 creates a batch job with thousands of domains including "domain.example".
P2 would like to test "domain.example". This domain is already in database because P1 added it with its batch job. But the test has not started yet because it comes later in the queue.
So P2 gets the _test-id_ associated with the test added by P1 for domain "domain.example". Here if there are thousands of tests to run, there is little chance that P2 gets its results in a decent time. A workaround would be for P2 to wait `age_reuse_previous_test` seconds before running its test again, then a new test will be created for "domain.example" with a higher priority.

Speaking of `test_start_time`, I realized only SQLite uses it. MySQL and PostgreSQL use `creation_time`.

#### MySQL

https://github.com/zonemaster/zonemaster-backend/blob/079dadbec18e23f89d3b76a0525257305a053d5a/lib/Zonemaster/Backend/DB/MySQL.pm#L134

#### PostgreSQL

https://github.com/zonemaster/zonemaster-backend/blob/079dadbec18e23f89d3b76a0525257305a053d5a/lib/Zonemaster/Backend/DB/PostgreSQL.pm#L146..L152

#### SQLite

https://github.com/zonemaster/zonemaster-backend/blob/079dadbec18e23f89d3b76a0525257305a053d5a/lib/Zonemaster/Backend/DB/SQLite.pm#L191

And by default SQLite uses the current timestamp for `test_start_time` when adding a test.

https://github.com/zonemaster/zonemaster-backend/blob/079dadbec18e23f89d3b76a0525257305a053d5a/lib/Zonemaster/Backend/DB/SQLite.pm#L74

I think we need to review this part as well. At least to make it consistent across databases engine.

matsduf:
Yes, it would be good if the three databases had the same behavior. 

In SQLite `test_start_time` is explicitly set for `create_new_test` but not for for `add_batch_job`, but since it has a default value it will not matter. `test_start_time` is reset several times, so it is really not a "start test time", at least not in SQLite.

The problem that you have identified is something that should be resolved if we should keep the batch function. There is probably nobody using it.

mattias-p:
> A workaround would be for P2 to wait age_reuse_previous_test seconds before running its test again, then a new test will be created for "domain.example" with a higher priority.

Good thinking!

With regard to priority I think the proper behavior when a testid is reused is to update its priority to the maximum old and new.

pnax:
> With regard to priority I think the proper behavior when a testid is reused is to update its priority to the maximum old and new.

This was what I had in mind as well (and expressed in https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-895301703). We tried to come to a solution in #840. Please have a look and tell us what you think (to summarize briefly: we came to the idea to create a new test when it has a higher priority than the one already in database).

pnax:
I pushed an update that looks for the priority and creates a new test if it has a higher priority than the one in database. Depending on what we chose here, I totally can revert this change and set the `priority` to the maximum between the old and new tests instead.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -196,8 +196,6 @@ sub _normalize_params {
     $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
     $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
     $normalized{profile}  = $$params{profile}   // "default";
-    $normalized{priority} = $$params{priority}  // 10;
-    $normalized{queue}    = $$params{queue}     // 0;
```
mattias-p, line 200: Why are priority and queue removed from the normalization? That seems wrong to me. Maybe I'm missing something?

pnax: They are removed because they shouldn't affect the fingerprint computation. And this normalization layer has been added to filter the parameters kept to compute the fingerprint.
We decided not to keep these parameters because they don't affect the outcome of a test. Having 2 tests with all parameters equal beside their priority (or queue) should not give a different fingerprint.

mattias-p (dismissed):

blacksponge:
Regarding the `queue` parameter, I wonder if this is the wanted behavior. Let's say that I have two pools of test agents in different testing environment running each with their own queue id. Then if I start the same test for each `queue` I don't want the test to be executed only in one of the testing pools but I want it to be executed in both of them.

pnax:
@blacksponge this is a very good remark. I didn't think about that at all. And indeed I think you are right, we need to consider the `queue` as this can affect the test's outcome (since the TestAgent can be in a different environment).

lib/Zonemaster/Backend/DB.pm
pnax: I see, I added this method while working on the fingerprint. I now realize that its name might be ambivalent. This is not exactly a _normalization_ per se, but more a _projection_ of the parameters. This might indeed need some refactoring.

matsduf (approved):
This works as expected.

pnax:
I added back the `queue` parameter to the fingerprint computation and I made some refactoring to better express the fact that we are not performing the normalization of all the parameters here.

Please re-review.

matsduf:
> Regarding the `queue` parameter, I wonder if this is the wanted behavior. Let's say that I have two pools of test agents in different testing environment running each with their own queue id. Then if I start the same test for each `queue` I don't want the test to be executed only in one of the testing pools but I want it to be executed in both of them.

That is of course theoritically possible that testing from different probes will give different results, but if that is the case, and results from those probes are collected to the same database then that is a problem of the test system set up. When you ask for previous tests Zonemaster will not separate on queues.

The queue example is a corner case that we should disregard. The assumption behind reusing the result is that if tests are run in a short time frame they will probably give the same result and the result of the previous test can be reused.

I object to include the queue because it is not relevant for the test as such.

docs/API.md
```diff
@@ -692,7 +692,7 @@ key in the configuration file.
 
 The parameters that are compared when to determine if two requests are to be
 considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`,
-`profile`, `priority` and `queue`.
+`profile` and `queue`.
```
matsduf, line 695: Queue should be exluced here.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -183,21 +183,20 @@ sub _new_dbh {
...
-    $normalized{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
-    $normalized{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
-    $normalized{profile}  = $$params{profile}   // "default";
-    $normalized{priority} = $$params{priority}  // 10;
-    $normalized{queue}    = $$params{queue}     // 0;
+    $projection{domain}   = lc $$params{domain} // "";
+    $projection{ipv4}     = $$params{ipv4}      // $profile->get( 'net.ipv4' );
+    $projection{ipv6}     = $$params{ipv6}      // $profile->get( 'net.ipv6' );
+    $projection{profile}  = $$params{profile}   // "default";
+    $projection{queue}    = $$params{queue}     // 0;
```
matsduf, line 199: Queue should be excluded here.

pnax:
> That is of course theoritically possible that testing from different probes will give different results, but if that is the case, and results from those probes are collected to the same database then that is a problem of the test system set up.

I can't get why this would be "a problem of the test system set up". Could you elaborate a little please, I think I am missing something.

Because I wasn't sure what the purpose of the `queue` parameter was, I went back to where this parameter was added to have a better view of what was its purpose, and I found the issue #211 and PR #223.

> What the backend is about to have is queue handling and make it possible to let different workers handle different queue. I see no reason why this should be temporary. And let us call this "queue" not "priority". That is less confusing for the users. I know that the field in the database is called "priority".
> https://github.com/zonemaster/zonemaster-backend/issues/211#issuecomment-259675814

This part is now clearer to me. And the way I see it, adding the `queue` to the fingerprint would expand the usage of this `queue` parameter. Why would it be that bad?

matsduf:
If we consider a delegated test run by different Zonemaster implementations, but using the same profile, then we would expect all to have the same result. If not, we would look for some error in testing even though time and net difference could result in differences. If we then consider a Backend installation with multiple probes (Test Agent sessions) for a dedicated queue where the same delegated test is run at the same time and with the same profile, we would definitely expect those to give the same result. When the two tests are presented in the same history and giving different results we would look for errors in testing.

The only use case that exist today for queues are running different probes (Test Agent sessions) to increase the capacity of Backend (but if that is used or not is unknown). If that is used, the expectation is that all Test Agent sessions will test in the same way to make sure the different test results are comparable.

As the system looks today, all tests that come from GUI will end up in the same queue and being run by the same Test Agent session.

The purpose of the fingerprint is very limited. If the same test is restarted within a short time frame, the previous result will be reused. The function is further more not used by tests started in batches. If two identical batches are started directly after each other all tests will be run twice even if the individual tests are run in a short time frame. The only real usage of the fingerprint and the function is when an end-user starts the same test within that short time frame. This saves some time for the end-user, and possibly saves some loads for the system (but it is no over-load protection for the system).

If the fingerprint excludes relevant data, result from a previous, non-identical test can be presented to the end-user. If the fingerprint includes non-relevant data, the test will be rerun even if not needed.

The queue information is non-relevant for the fingerprint, but since the the only relevant use case, tests from GUI, always have the same queue setting, it will in practice not have any affect. Furthermore, the probability that the same test comes with a different queue setting, such as in a batch test, within the time frame is very low.

Even if I still consider it to be wrong to include the queue parameter in the fingerprint, I will not insist since it is irrelevant in practice if it is included or not.

matsduf (dismissed):
It is wrong to keep queue value in fingerprint, but in practice is does not give any difference due to the way queues are set and the way fingerprint is used.

pnax:
You convinced me. I reverted the change. The `queue` is no more in the fingerprint computation.

matsduf (approved):

Issue #805, opened 2021-06-29
Better database initialization


mattias-p:
Today the installation instruction prescribes database initialization in different ways for different database engines. For MySQL and PostgreSQL a combination of `perl`, `cd` and `mysql`/`psql` is used. For SQLite, a combination of `perl`, `cd`, `install` and `chown` is used. The same approach is used in the section about cleaning up the database. 

Today the definitions themselves are located in different places. For MySQL and PostgreSQL the definitions are duplicated across the share/ and script/ directories in the repository. For SQLite the definition is located in the database adapter module.

There is already an issue #689 to deduplicate the database definitions for MySQL and PostgreSQL. This issue goes beyond that.

Requirements:
* A command line tool should be created for initializing and cleaning up the database.
* The tool should be used in the installation instruction.
* The database definitions for all database engines should be located in their respective database adapter modules.

matsduf:
@mattias-p, can we close this?

PR #834, opened 2021-07-28
Use a unique set of definitions to initialize the database

### LINKED ISSUE #689: Duplicated database definitions
There are two sources for creating the MySQL (Maria DB) or PostgreSQL database, respectively:

```
share/initial-mysql.sql
share/initial-postgres.sql
```
```
script/create_db_mysql.pl
script/create_db_postgresql_9.3.pl
```

The installation instructions refer to the `.sql` files, whereas Travis scripts use the Perl scripts. Do we need both or could we decide on one variant? With two there is a high risk of deviation.

We already have a deviation for MySQL (Maria DB):

In `script/create_db_mysql.pl`: `results blob DEFAULT NULL,`

In `share/initial-mysql.sql`: `results mediumblob DEFAULT NULL,`

The latter was introduced to resolve an issue with large results.

pnax:
## Purpose

This PR is about removing duplicate in database initialization as expressed in #689.

## Context

Fixes #689.
Partially addresses #805 

## Changes

* Database and user creations are done manually on Backend installation (as documented in the installation document)

* Creates a new script `share/create_db.pl` that can be used to create the tables for any database engine.

* Removes files `share/initial-mysql.sql`, `share/initial-postgres.sql`, `script/create_db_mysql.pl`, `script/create_db_postgresql_9.3.pl` and `share/create_db_sqlite.pl`. The SQL commands are now part of a `create_db()` routine in the MySQL.pm and PostgreSQL.pm files respectively.

* Do not drop the tables before creating them anymore.

I left the whole commit history. Some commits were only needed to make the code easier to refactor later in the PR.

## How to test this PR

Follow the [installation instructions (develop)][installation] and make sure you have a running Backend.

[installation]: https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/Installation.md


blacksponge:
Nice work, this looks promising. I do have a few comments.

I think you should use the connection settings provided in the configuration as much as possible, the database can be on a different host or port, this should be handled. Also I think the user creation should be optional. Maybe only use the superuser to create the local user and grant privileges if required, but then use the new user to create the tables; this can even be done using the database handler in perl. What do you think?

pnax:
@blacksponge good remarks. I tried to only do some refactoring and keep the old behavior as much as possible. If we agree on this PR, I see few updates here and there (as well as your remarks).

matsduf:
>     * Creates a new script `share/create_db.pl` that can be used to automatically create the database for all database engine. The config file is used to know which engine to use. It's also used to get specific values such as the database name, the database user and its password.

Is it really easier to maintain one script that should handle all database engines instead three different script? For the user it will be the same.

share/create_db.pl
```diff
@@ -0,0 +1,119 @@
...
+
+my $OPT_USERNAME = 0;
+my $OPT_ASKPASS  = 0;
+
+sub usage {
+    print "Usage: " . basename($0) . " [-u] [-p] [-h]\n";
+    print "\n";
+    print "Automatically initialize the Zonemaster database based on the Zonemaster Backend config\n";
+    print "file. Use the values to create the database and a user with enough privileges to\n";
+    print "access this database. Also create all the Zonemaster tables inside this database.\n";
```
matsduf, line 19: For FreeBSD this depends on #835 to find the configuration file.

To be consistent: "Zonemaster Backend" --> "Zonemaster-Backend" or "Zonemaster::Backend".

pnax:
> Is it really easier to maintain one script that should handle all database engines instead three different script?

Maybe not.

TBH I am still not completely satisfied with this PR. I need to think about it a little more.

pnax:
I took a 180 degrees shift. I choose to remove the `.sql` files and keep the `create_db` subroutines (moving them in the MySQL.pm and PostgreSQL.pm file).
Database and user creations are back to the documentation. There is no wrapper script beside the new `share/create_db.pl` script that is only about initializing the tables independently from the engine.

I also updated the installation process with notes to update the "backend_config.ini" file based on the database name chosen, and its user/password.

The new installation procedure needs to be tested on FreeBSD.

The way I see it in the near future is to remove the call to `share/create_db.pl` in the installation process and make the Backend call `create_db()` on startup instead.

blacksponge:
Tested with postgres 12 and mariadb 10, it is working well. I think it is cleaner this way.

blacksponge (approved):

pnax:
I realized that the `create_db()` routine in SQLite was dropping the tables. I removed this part and included it into the related commit "Do not DROP tables on creation".

matsduf (commented):
I do not see the gain of this change. For the user (installation instructions) it will require more steps to create the database, especially for MariaDB/MySQL, but also for PostgreSQL and SQLite. We should try to make the installation step easy.

pnax:
> We should try to make the installation step easy.

I agree on this. And I think this might help go that way even if this PR adds more steps to the installation process.

When you install the Backend, you only need to create a database and a user (for MySQL and PostgreSQL). You also need to set these values into the `backend_config.ini` file. That is all what is required, and this might require certain privileges. Once you have performed these steps, with the needed privileges, you can drop them and let the Backend perform the rest of the operations (the tables' creation).

This PR is about splitting the process in two. One part creates the database and the user (if applicable). The other part is about creating the tables. Currently everything has to be made by hand during installation, but a future work I am planning is to let the Backend automatically create the tables on startup (if they don't exist). This would then remove the "Database configuration" section for each engine. *
Then this might also give the possibility to develop a script that creates the database and the user (if applicable). I think that such a script should takes its inputs from the `backend_config.ini` file. But it should also be possible to bypass the database and/or user creation.

Finally I think that having the commands to create the database (and user if applicable) inside the installation instruction makes it easier for a user to tweak the installation process to its needs (for instance use a different database name, user, password, or even the database's host or port).

\* _I didn't add the tables' creation by the Backend to this PR to avoid overloading this PR_

matsduf:
If the second step can be done before the next release I am fine with this PR, but if we risk having such complex installation instructions at next release I am not fine with it.

> * _I didn't add the tables' creation by the Backend to this PR to avoid overloading this PR_

That is a good ambition.

pnax:
> If the second step can be done before the next release I am fine with this PR

This second step is done in #846

matsduf:
> > If the second step can be done before the next release I am fine with this PR
> 
> This second step is done in #846

if #846 is the end result, then the end result gives a installation instruction that is too complex. I cannot see why it should be more complex than today. Command line commands (shell) is simpler than having to give SQL commands directly to the server or editing the configuration file.

pnax:
> if #846 is the end result, then the end result gives a installation instruction that is too complex. I cannot see why it should be more complex than today. Command line commands (shell) is simpler than having to give SQL commands directly to the server or editing the configuration file.

Yes #846 is the end result.
Indeed it is easier, however not easily adaptable. Say you want to use a different database, you need to update the "share/initial-mysql.sql" or "share/initial-postgres.sql" file, or manually create it and run "script/create_db_mysql.pl" or "script/create_db_postgresql_9.3.pl" as done with Travis.
Since the goal is about removing duplicates in the database definition, I realized that Travis that uses a different database that requires to run commands different from the one in the installation document in order to create the database. So I tried to merge everything to come with a unique solution. Maybe this can be improved somehow to avoid having all these new SQL commands in the installation process.

mattias-p:
I really like that you put the database definitions in the adapter modules!

I agree with @matsduf that it's nicer to have only shell commands in the installation instruction, rather than jumping in and out of special purpose shells (e.g. mysql).

I can see four ways forward:
1. Merge this PR as is.
2. Make each SQL statement into its own command (e.g. `echo "CREATE DATABASE zonemaster" | sudo mysql`).
3. Put these SQL statements into an SQL file and have the user run that.
4. Move the database and user creation statements into the `create_db` procedure and make them use the database name and user name from the config file.

The fourth option has the advantage that it automatically coordinates the database name, user name and password between the SQL statements and the configuration file.

pnax:
> I agree with @matsduf that it's nicer to have only shell commands in the installation instruction, rather than jumping in and out of special purpose shells (e.g. mysql).

Oh, I think I now understand what was meant by "too complex", thanks for the light.
To be honest, when I was working on this PR I found that we had a very detailed instructions set for MySQL for FreeBSD, so I was inspired by it.

The thing with the 4th option is that you cannot reuse the `data_source_name` string as is because it takes the database name which does not exist yet. Hence you can't use `dbh()` either or `_new_db()` because they take a "user" and "password" (for MySQL and PostgreSQL) that are not yet defined. We could update the code to create the database and user when calling `create_db`, but I think this is too much effort for a small gain (but I might be wrong).

About the 3rd option, this will be slightly harder to tweak the parameters to someone needs.

So I'll go with option 2 to address your remarks.

pnax:
> So I'll go with option 2 to address your remarks.

I realized that this might leak the password.
```
sudo mysql -e "CREATE USER 'zonemaster'@'localhost' IDENTIFIED BY 'zonemaster';"
```

Should I make such update in the installation document?

pnax:
I've rebased the branch on top of develop and fixed the conflicts within the Installation file.

pnax:
@blacksponge, @mattias-p, @matsduf could you re-review?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -95,6 +95,92 @@ sub dbh {
     return $self->dbhandle;
 }
 
+sub create_db {
```
blacksponge, line 98: Maybe include the `create_db` method in the `requires` of DB.pm ?

blacksponge (commented):
Nice work! Just a small comment

docs/Installation.md
```diff
@@ -276,7 +283,17 @@ See appendices for [MariaDB][MariaDB instructions for Debian] and
...
 
-### 4.3 Service configuration and startup (Debian/Ubuntu)
+### 4.3 Database configuration (Debian/Ubuntu)
+
+Finally initialize the database:
+
+```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
+sudo -u zonemaster perl create_db.pl
+```
```
matsduf, line 293: Suggestion to make it easier to just copy and paste:

1. Make `create_db.pl` executable
2. Make the command a one-liner:
``` sh
sudo -u zonemaster $(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")')/create_db.pl
```

docs/Installation.md
```diff
@@ -390,7 +405,17 @@ See appendices for [MariaDB][MariaDB instructions for FreeBSD] and
...
 
-### 5.3 Service startup (FreeBSD)
+### 5.3 Database configuration (FreeBSD)
+
+Finally initialize the database:
+
+```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
+su -m zonemaster -c "perl create_db.pl"
+```
```
matsduf, line 415: Suggestion to make it easier to just copy and paste:

1. Make `create_db.pl` executable
2. Make the command a one-liner:
``` sh
su -m zonemaster -c "$(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")')/create_db.pl"
```

docs/Installation.md
```diff
@@ -477,15 +502,25 @@ sudo systemctl enable mariadb
...
+Create a new user and give it all permissions on the newly created database:
+```sql
+CREATE USER 'zonemaster'@'localhost' IDENTIFIED BY 'zonemaster';
+GRANT ALL ON zonemaster.* TO 'zonemaster'@'localhost';
 ```
 
-> **Note:** This creates a database called `zonemaster`, as well as a user
-> called "zonemaster" with the password "zonemaster" (as stated in the config
-> file). This user has just enough permissions to run the backend software.
+Then update the `/etc/zonemaster/backend_config.ini` file accordingly.
```
matsduf, line 523: What is the expected update? The above setting is the default value, isn't it? Should it be
```
Then update the `/etc/zonemaster/backend_config.ini` file with username and password unless the default values are used.
```

docs/Installation.md
```diff
@@ -504,15 +539,25 @@ sudo sed -i '/\bengine\b/ s/=.*/= MySQL/' /etc/zonemaster/backend_config.ini
...
```
matsduf, line 560: What is the expected update? The above setting is the default value, isn't it? Should it be
```
Then update the `/etc/zonemaster/backend_config.ini` file with username and password unless the default values are used.
```

docs/Installation.md
```diff
@@ -477,15 +502,25 @@ sudo systemctl enable mariadb
...
+Create the database:
+```sql
+CREATE DATABASE zonemaster;
+```
+
+Create a new user and give it all permissions on the newly created database:
+```sql
+CREATE USER 'zonemaster'@'localhost' IDENTIFIED BY 'zonemaster';
+GRANT ALL ON zonemaster.* TO 'zonemaster'@'localhost';
 ```
```
matsduf, line 522: I suggest (see FreeBSD):

Logout from database:

```sql
exit;
```

docs/Installation.md
```diff
@@ -504,15 +539,25 @@ sudo sed -i '/\bengine\b/ s/=.*/= MySQL/' /etc/zonemaster/backend_config.ini
...
```
matsduf, line 559: I suggest (see FreeBSD):

Logout from database:

```sql
exit;
```

docs/Installation.md
```diff
@@ -554,23 +599,24 @@ Reset root password in MySQL (required by MySQL). Replace
...
-cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
-mysql -u root -p < ./initial-mysql.sql
+```sql
+exit;
 ```
 
-> **Note:** This creates a database called `zonemaster`, as well as a user
-> called "zonemaster" with the password "zonemaster" (as stated in the config
-> file). This user has just enough permissions to run the backend software.
+Then update the `/etc/zonemaster/backend_config.ini` file accordingly.
```
matsduf, line 619: What is the expected update? The above setting is the default value, isn't it? Should it be
```
Then update the `/etc/zonemaster/backend_config.ini` file with username and password unless the default values are used.
```

share/create_db.pl
```diff
@@ -0,0 +1,13 @@
+use strict;
```
matsduf, line 1: I suggest the following to make it an executable script (and the bits have to be set).
```
#!/usr/bin/env perl
use strict;
```

matsduf (changes requested):

docs/Installation.md
pnax: Good point, I'll do the update.

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: done

docs/Installation.md
```diff
@@ -145,7 +145,17 @@ See appendices for [MariaDB][MariaDB instructions for CentOS] and
...
 
-### 3.3 Service configuration and startup (CentOS)
+### 3.3 Database configuration (CentOS)
+
+Finally initialize the database:
+
+```sh
+cd `perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backend")'`
+sudo -u zonemaster perl create_db.pl
+```
```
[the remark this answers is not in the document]
pnax: Good suggestion. I've made the changes.

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

docs/Installation.md
pnax: done

share/create_db.pl
pnax: done

docs/Installation.md
matsduf: I'm sorry, for FreeBSD it was a bad suggestion from me. Default shell for FreeBSD root is C shell (or tcsh), and that does not support "$( )". Bourne shell (or sh) is the default shell for other users, and that supports "$( )". Back ticks have to be used, and that affects <"> too.

``` sh
su -m zonemaster -c "`perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir(qw(Zonemaster-Backend))'`/create_db.pl"
```

matsduf (changes requested):

docs/Installation.md
pnax: This should be fixed (and I realized that I forgot to remove the call to perl since the script is now executable).

matsduf (approved):

PR #833, opened 2021-07-28
add scripts to backfill undelegated fields in db

blacksponge:
## Purpose

Update the old 'undelegated' values in the database.
- [x] PostgreSQL tested with ~1% of the records of the zonemaster.net database on Postgres 9.3 and 13.3
- [x] SQLite
- [x] MySQL tested with  ~1% of the records of the zonemaster.net database on MariaDB 10.3

## Context

Fixes #829

## Changes

Add patch scripts.

### Postgres

* Check the value of delegated vs undelegated tests (all test are delegated, so the second query should return nothing except if you have the patch for undelegated tests in which case there can be a few).

  ```pgsql
  select DISTINCT params->>'ds_info' as ds_info, params->>'nameservers'  as nameservers from test_results where undelegated = 0 ;
  
  select DISTINCT params->>'ds_info' as ds_info, params->>'nameservers' as nameservers from test_results where undelegated = 1 ;
  ```
* Apply update script `perl zonemaster-backend/share/patch_postgresql_db_zonemaster_backend_ver_7.0.0.pl` (ignore error if the table has already been changed, "if not exists" is not compatible with PG 9.3)

* Repeat the first step to verify that all tests has been classified correctly.

### MySQL

* Check the value of delegated vs undelegated tests (all test are delegated, so the second query should return nothing except if you have the patch for undelegated tests in which case there can be a few).

  ```sql
   select distinct json_extract(params, '$.nameservers'), json_extract(params, '$.ds_info') from test_results where undelegated = 0;
  
   select distinct json_extract(params, '$.nameservers'), json_extract(params, '$.ds_info') from test_results where undelegated = 1;
  ```
* Apply update script `perl zonemaster-backend/share/patch_mysql_db_zonemaster_backend_ver_7.0.0.pl`

* Repeat the first step to verify that all tests has been classified correctly.

PR #835, opened 2021-07-29
improve config search and add default freebsd config location

blacksponge:
## Purpose

Improve the config search mechanism 

## Context
Following the comment https://github.com/zonemaster/zonemaster-backend/pull/833#issuecomment-888965227

## Changes

* iterate through a default config location list to find an existing file
* add default freebsd location to the search list (`/usr/local/etc/zonemaster/backend_config.ini`)
* update travis file to correctly install the share directory

## How to test this PR

* Create the file `/etc/zonemaster/backend_config.ini`, run the backend, check that this is the config file loaded, then delete it
* Create the file `/usr/local/etc/zonemaster/backend_config.ini`, run the backend, check that this is the config file loaded, then delete it
* Create the file `/tmp/my_config.ini`, set the environment variable `ZONEMASTER_BACKEND_CONFIG_FILE` to `/tmp/my_config.ini`, run the backend, check that this is the config file loaded, then delete it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,6 +107,10 @@ before_install:
   - git clone --depth=1 --branch=$TRAVIS_BRANCH https://github.com/zonemaster/zonemaster-engine.git
   - ( cd zonemaster-engine && cpanm --verbose --notest . ) && rm -rf zonemaster-engine
 
+    # Install share files
+  - mkdir -p ./lib/auto/share/dist/
+  - ln -s ../../../../share ./lib/auto/share/dist/Zonemaster-Backend
+
 before_script:
     - if [[ "$TARGET" == "PostgreSQL" ]]; then psql -c "create user travis_zonemaster WITH PASSWORD 'travis_zonemaster';" -U postgres; fi
     - if [[ "$TARGET" == "PostgreSQL" ]]; then psql -c 'create database travis_zonemaster OWNER travis_zonemaster;' -U postgres; fi
@@ -122,4 +126,3 @@ before_script:
 
 script:
     - perl Makefile.PL && make test
-
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -17,16 +17,16 @@ use Zonemaster::Backend::DB;
 
 our $path;
 
-my @search_paths = (
-    '/etc/zonemaster/backend_config.ini',
-    '/usr/local/etc/zonemaster/backend_config.ini',
-    dist_file('Zonemaster-Backend', "backend_config.ini")
-);
-
 if ($ENV{ZONEMASTER_BACKEND_CONFIG_FILE}) {
     $path = $ENV{ZONEMASTER_BACKEND_CONFIG_FILE};
 }
 else {
+    my @search_paths = (
+        '/etc/zonemaster/backend_config.ini',
+        '/usr/local/etc/zonemaster/backend_config.ini',
+        dist_file('Zonemaster-Backend', "backend_config.ini")
+    );
+
     for my $default_path (@search_paths) {
         if ( -e $default_path ) {
             $path = $default_path;
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -17,7 +17,7 @@ use Zonemaster::Backend::DB;
 
 our $path;
 
-my @seatch_paths = (
+my @search_paths = (
     '/etc/zonemaster/backend_config.ini',
     '/usr/local/etc/zonemaster/backend_config.ini',
     dist_file('Zonemaster-Backend', "backend_config.ini")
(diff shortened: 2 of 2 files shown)


matsduf:
Is it really good to add the number of places that `Config.pm` "magically" looks for the configuration file? Isn't it better to let e.g. the start script point out by environment variable where to find it?

blacksponge:
The environment variable can still be used, I am just adding one more common place to search to the already existing list `Config.pm` is using. I think that searching for a configuration in standard places is a common pattern but I might be wrong.

Also those are pretty common places to have the configuration file, I think I prefer to have it specified once here than having to deal with edge cases in multiple location because the ones listed here don't work for an OS that we say we are supporting.

matsduf:
> Also those are pretty common places to have the configuration file, I think I prefer to have it specified once here than having to deal with edge cases in multiple location because the ones listed here don't work for an OS that we say we are supporting.

I understand what you mean and I see the benefit of it too. There were some issues before when unit tests used the configuration files under /etc which made them fail when the updates were not backward compatible. I guess that @mattias-p has fixed that issue.

I agree that is simplifies.

.travis.yml
```diff
@@ -107,6 +107,10 @@ before_install:
   - git clone --depth=1 --branch=$TRAVIS_BRANCH https://github.com/zonemaster/zonemaster-engine.git
   - ( cd zonemaster-engine && cpanm --verbose --notest . ) && rm -rf zonemaster-engine
 
+    # Install share files
+  - mkdir -p ./lib/auto/share/dist/
+  - ln -s ../../../../share ./lib/auto/share/dist/Zonemaster-Backend
```
matsduf, line 112: How does this change relate to this PR?

matsduf (approved):

.travis.yml
blacksponge: Otherwise the tests fail as `dist_file` cannot find the dist directory for Zonemaster-Backend

matsduf: This PR adds a second path to look for `backend_config.ini`, else Config.pm still falls back to what `dist_file` gives. What has changed that requires a change for Travis? Or is it a bug that you fix at the same time?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -16,14 +16,23 @@ use Zonemaster::Backend::Validator qw( :untaint );
...
-elsif ( -e '/etc/zonemaster/backend_config.ini' ) {
-    $path = '/etc/zonemaster/backend_config.ini';
-}
 else {
-    $path = dist_file('Zonemaster-Backend', "backend_config.ini");
+    for my $default_path (@search_paths) {
+        if ( -e $default_path ) {
+            $path = $default_path;
+            last;
+        }
```
matsduf, line 34: In the current code, `$path` is always defined. In the new code it will be undefined if the environment variable is undefinded and `backend_config.ini` does not exist in any of the three paths. That will probably give a different error message. It should probably be defined.

matsduf (changes requested):

lib/Zonemaster/Backend/Config.pm
blacksponge: The error will be the same actually. If `dist_file('Zonemaster-Backend', "backend_config.ini")` fails to find the file, then the backend fails to start with error `Failed to find shared file 'backend_config.ini' for dist 'Zonemaster-Backend'` in both cases, otherwise `$path` will be defined as the file exists.

.travis.yml
blacksponge: Before, `dist_file` was not executed as we are using the environment variable, now it is always executed, that is the difference.

pnax: If I may,  instead of having to install the share files I'd rather avoid executing `dist_file` if the `ZONEMASTER_BACKEND_CONFIG_FILE` variable is set. This would prevent failure such as the Travis one. Would it be possible to execute `dist_file` only if the environment variable is unset?

blacksponge: That should be done

pnax: Nice, thanks. Then the Travis patch could be removed I guess.

blacksponge: > Then the Travis patch could be removed I guess.

I would prefer to keep it so that the share directory is correctly installed.

pnax: All right.

pnax (approved):

.travis.yml
matsduf: > Before, `dist_file` was not executed as we are using the environment variable, now it is always executed, that is the difference.

That change is not included in the description. Please add that.

lib/Zonemaster/Backend/Config.pm
matsduf: Please update the description of this PR to include the change for Travis, and I am fine with the changes.

.travis.yml
blacksponge: This has been reverted, `dist_file` is now executed only if the variable is not set, as it was before.

lib/Zonemaster/Backend/Config.pm
blacksponge: I've updated the description.

matsduf (approved):

matsduf:
## Testing v2021.2
LGTM.

As far as I can see it works as it should.

Issue #830, opened 2021-07-21
New value to the "filter" property in get_test_history method


matsduf:
The filter property in method `get_test_history` has three legal values:
* all -- all tests are listed (default if the property is empty or not included)
* delegated -- delegated tests are listed
* undelegated -- undelegated tests are listed

Issue zonemaster/zonemaster-gui#251 asks for the return JSON to be updated so that for each test returned, the value of "undelegated" (0 or 1) is included.

This issue suggests a new value for the property, `tagged`, which will as `all` return all tests with the addition of a extra property in the result, i.e. "undelegated" with its value.

matsduf:
Resolved by #837

PR #837, opened 2021-08-04
add undelegated value in results history

blacksponge:
## Purpose

Add more information on the tests in history.

## Context

#830

## Changes

Return an `undelegated` in the test history items.

## How to test this PR

Create a delegated and undelegated test, then query the history and check the `undelegated` property:    
`./script/zmb get_test_history --domain toto.com --filter all | jq`

```json
{
  "jsonrpc": "2.0",
  "result": [
    {
      "creation_time": "2021-08-04 08:13:16",
      "id": "9aa193588f53b8d8",
      "overall_result": "error",
      "undelegated": 1
    },
    {
      "id": "b60fb8f881b199f8",
      "creation_time": "2021-07-20 08:39:01",
      "undelegated": 0,
      "overall_result": "error"
    }
  ],
  "id": 1
}
```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -944,12 +944,12 @@ Example response:
     {
       "id": "c45a3f8256c4a155",
       "creation_time": "2016-11-15 11:53:13.965982",
-      "undelegated": 1,
+      "undelegated": true,
       "overall_result": "error",
     },
     {
       "id": "32dd4bc0582b6bf9",
-      "undelegated": 0,
+      "undelegated": false,
       "creation_time": "2016-11-14 08:46:41.532047",
       "overall_result": "error",
     },
--- a/docs/API.md
+++ b/docs/API.md
@@ -1000,7 +1000,7 @@ An object with the following properties:
     `"ERROR"`, but none with `"CRITICAL"`.
   * `"critical"`, if there is at least one message with *severity level*
     `"CRITICAL"`.
-* `"undelegated"`: `1` if the test is undelegated, `0` otherwise.
+* `"undelegated"`: `true` if the test is undelegated, `false` otherwise.
 
 #### `"error"`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -255,7 +255,7 @@ Properties:
 ### Non-negative integer
 
 Basic data type: number (integer)
- 
+
 A non-negative integer is either zero or strictly positive.
 
 
@@ -371,7 +371,7 @@ The object has three keys, `"module"`, `"message"` and `"level"`.
 
 Sometimes additional keys are present.
 
-* `"ns"`: a *domain name*. The name server used by the *test module*. 
+* `"ns"`: a *domain name*. The name server used by the *test module*.
 This key is added when the module name is `"NAMESERVER"`.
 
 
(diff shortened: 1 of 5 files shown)


pnax (commented):
Works as expected.
Could you add the new property in the "result" section in the [documentation](https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/API.md#result-8)?

pnax (dismissed):

docs/API.md
```diff
@@ -998,7 +1000,7 @@ An object with the following properties:
     `"ERROR"`, but none with `"CRITICAL"`.
   * `"critical"`, if there is at least one message with *severity level*
     `"CRITICAL"`.
-
+* `"undelegated"`: `1` if the test is undelegated, `0` otherwise.
```
matsduf, line 1003: This is a breaking change. My suggestion was to add a new value on "filter" that would then include `undelegated` in the response. That would not be breaking.

Would that be expensive? What are the downsides of that?

blacksponge: I am not sure it is such a breaking change, clients that don't support that could just ignore the property.

In any cases`filter` should be kept to filtering only, making this argument change the content of the response looks like a very counter intuitive thing to do. Furthermore you could want to have the value in the response while filtering the result (because it could be required later in the client or because the filter could be augmented to filter on other properties). What you propose does not make that possible.

mattias-p: I agree with @blacksponge. Adding a new field in the response shouldn't be considered breaking.

But why is it `1` and `0` rather than `true` and `false`?

Another option to consider is to have a "type" field instead, with the two possible values `delegated` and `undelegated`.

blacksponge: I changed undelegated to be a boolean.

matsduf: 1. If you add a property, is that not a breaking change? Should the client just ignore unknown properties according to RPC?
2. In the document it still says >>`"undelegated"`: `1` if the test is undelegated, `0` otherwise.<<

blacksponge: > If you add a property, is that not a breaking change? Should the client just ignore unknown properties according to RPC?

I think the client should ignore unknown properties, yes. This allows to extend the API without breaking.

> In the document it still says >>"undelegated": 1 if the test is undelegated, 0 otherwise.<<

should be fixed.

matsduf: > > If you add a property, is that not a breaking change? Should the client just ignore unknown properties according to RPC?
> 
> I think the client should ignore unknown properties, yes. This allows to extend the API without breaking.

In the API document it says
> * Extra top-level properties in request objects are allowed but ignored.
> * Extra properties in the `"params"` object are allowed for some methods but ignored for others.

Neither in the API document  nor in the [JSON-RPC 2.0] specification I can find any support for unknown properties in responses. If I am wrong, it is fine.

[JSON-RPC 2.0]:  https://www.jsonrpc.org/specification

mattias-p: It's very reasonable to assume that clients only look for properties that they know about in responses. If we add a new property to a response without changing the semantics of the properties, the client should is able to continue as if nothing happened. This kind of a change should result in an incremented second digit in semantic versioning.

mattias-p (approved):

docs/API.md
matsduf: I do not see any problem to have it like that, but then the API document should state that since it is not clear from JSON-RPC 2.0.

matsduf (approved):

PR #839, opened 2021-08-05
use plack reverse proxy middleware to get real remote ip

### LINKED ISSUE #838: add_api_user remote ip check don't work behind reverse proxy
When running the API behind a reverse proxy on the same machine (like it is using the [configuration example](https://github.com/zonemaster/zonemaster-gui/blob/master/zonemaster.conf-example) provided by the GUI) the remote ip might always be localhost even if the query was done from elsewhere so the check done by `add_api_user ` is useless.

A better solution would be to use the [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For), and only default to the remote ip if it is not present.

blacksponge:
## Purpose

Secure the add_api_user method.

## Context

Fixes #838 

## Changes

* Use `Plack::Middleware::ReverseProxy`  to set the environment according to the proxy headers
* Add two configuration keys `API.enable_add_batch_job` and `API.enable_add_api_user` to enable the corresponding api methods (disabled by default).

## How to test this PR

* Running `./script/zmb add_api_user --username toto --api-key secret` should return the error `"Procedure 'add_api_user' not found"`.
* Running `./script/zmb add_batch_job --domain example.com --username toto --api-key secret` should return the error `"Procedure 'add_batch_job' not found"`.
* Enable the add_api_user method.
  * run the API behind a reverse proxy on the same host
    ```
    curl -s "Content-Type: application/json" --data '{"method":"add_api_user","params":{"username":"toto","api_key":"toto"},"id":1,"jsonrpc":"2.0"}' not.localhost/api  | jq
    ```
  * Try to add a user from a another host
  * If the query return `0` as result it works.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -59,13 +59,15 @@ Available keys: `enable_batch_jobs`, `enable_add_api_user`.
 
 Boolean value to enable the batch jobs methods of the API.
 
-Accpected values: 0 or 1, default to 0 (disabled).
+Accpected values: `yes` (or `true`) or `no` (or `false`),
+default to `no` (disabled).
 
 ### enable_add_api_user
 
 Boolean value to enable the `add_api_user` method of the API.
 
-Accpected values: 0 or 1, default to 0 (disabled).
+Accpected values: `yes` (or `true`) or `no` (or `false`),
+default to `no` (disabled).
 
 
 ## DB section
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -53,11 +53,11 @@ Each section in `backend_config.ini` is documented below.
 
 ## RPCAPI section
 
-Available keys: `enable_batch_jobs`, `enable_add_api_user`.
+Available keys: `enable_add_batch_job`, `enable_add_api_user`.
 
-### enable_batch_jobs
+### enable_add_batch_job
 
-Boolean value to enable the batch jobs methods of the API.
+Boolean value to enable the `enable_add_batch_job` methods of the API.
 
 Accpected values: `yes` (or `true`) or `no` (or `false`),
 default to `no` (disabled).
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -51,6 +51,23 @@ Repeating a key name in one section is forbidden.
 
 Each section in `backend_config.ini` is documented below.
 
+## API section
+
+Available keys: `enable_batch_jobs`, `enable_add_api_user`.
+
+### enable_batch_jobs
+
+Boolean value to enable the batch jobs methods of the API.
+
+Accpected values: 0 or 1, default to 0 (disabled).
+
+### enable_add_api_user
+
+Boolean value to enable the `add_api_user` method of the API.
+
+Accpected values: 0 or 1, default to 0 (disabled).
+
+
 ## DB section
 
 Available keys : `engine`, `user`, `password`, `database_name`,
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -59,14 +59,14 @@ Available keys: `enable_add_batch_job`, `enable_add_api_user`.
 
 Boolean value to enable the `enable_add_batch_job` methods of the API.
 
-Accpected values: `yes` (or `true`) or `no` (or `false`),
+Accepted values: `yes` (or `true`) or `no` (or `false`),
 default to `no` (disabled).
 
 ### enable_add_api_user
 
 Boolean value to enable the `add_api_user` method of the API.
 
-Accpected values: `yes` (or `true`) or `no` (or `false`),
+Accepted values: `yes` (or `true`) or `no` (or `false`),
 default to `no` (disabled).
 
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -60,14 +60,14 @@ Available keys: `enable_add_batch_job`, `enable_add_api_user`.
 Boolean value to enable the `enable_add_batch_job` methods of the API.
 
 Accepted values: `yes` (or `true`) or `no` (or `false`),
-default to `no` (disabled).
+default to `yes` (enabled).
 
 ### enable_add_api_user
 
 Boolean value to enable the `add_api_user` method of the API.
 
 Accepted values: `yes` (or `true`) or `no` (or `false`),
-default to `yes` (enabled).
+default to `no` (disabled).
 
 
 ## DB section
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -67,7 +67,7 @@ default to `no` (disabled).
 Boolean value to enable the `add_api_user` method of the API.
 
 Accepted values: `yes` (or `true`) or `no` (or `false`),
-default to `no` (disabled).
+default to `yes` (enabled).
 
 
 ## DB section
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -51,7 +51,7 @@ Repeating a key name in one section is forbidden.
 
 Each section in `backend_config.ini` is documented below.
 
-## API section
+## RPCAPI section
 
 Available keys: `enable_batch_jobs`, `enable_add_api_user`.
 
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -146,7 +146,7 @@ sub parse {
     $obj->_set_ZONEMASTER_lock_on_queue( '0' );
     $obj->_set_ZONEMASTER_age_reuse_previous_test( '600' );
     $obj->_set_RPCAPI_enable_add_api_user( 'no' );
-    $obj->_set_RPCAPI_enable_add_batch_job( 'yes' );
+    $obj->_set_RPCAPI_enable_add_batch_job( 'no' );
     $obj->_add_LANGUAGE_locale( 'en_US' );
     $obj->_add_public_profile( 'default', undef );
 
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -146,7 +146,7 @@ sub parse {
     $obj->_set_ZONEMASTER_lock_on_queue( '0' );
     $obj->_set_ZONEMASTER_age_reuse_previous_test( '600' );
     $obj->_set_RPCAPI_enable_add_api_user( 'no' );
-    $obj->_set_RPCAPI_enable_batch_jobs( 'no' );
+    $obj->_set_RPCAPI_enable_add_batch_job( 'yes' );
     $obj->_add_LANGUAGE_locale( 'en_US' );
     $obj->_add_public_profile( 'default', undef );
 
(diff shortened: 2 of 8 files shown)


pnax (dismissed):

matsduf (approved):

matsduf (changes requested):
The protection of `add_api_user` might be week even after this change. If RPCAPI is behind `stunnel` to create https access localhost would be used to the RPCAPI. Since batch is in limited use and before we have a better solution, I suggest that we add a key to the configuration file that must be set to enable the use of batch. It could one key covering both, or one key for `add_api_user`("enable_add_api_user") and one for `add_batch_job`("enable_add_batch_job"). Both should be off by default.

A long term solution could be to move `add_api_user` out of RPCAPI:

* Command line script adding and removing API users.
* Configuring the API user in the configuration file, both user and password.

blacksponge:
> A long term solution could be to move add_api_user out of RPCAPI.

I think this is a good idea, if we were to use a command line script, should it be a dedicated one or could it be integrated to an existing script?

matsduf:
> > A long term solution could be to move add_api_user out of RPCAPI.
> 
> I think this is a good idea, if we were to use a command line script, should it be a dedicated one or could it be integrated to an existing script?

As far as I can see we have no script to use for that.

matsduf:
Can this PR implement a configuration key to disable/enable batch to really have it solved?

blacksponge:
> Can this PR implement a configuration key to disable/enable batch to really have it solved?

Sure, I'll do that.

blacksponge:
> Can this PR implement a configuration key to disable/enable batch to really have it solved?

I've added two configuration key to enable / disable the batch and user api.

matsduf (dismissed):

matsduf:
We should also provide a patch for the current release. Could this PR also work directly on current master?

blacksponge:
> We should also provide a patch for the current release. Could this PR also work directly on current master?

I agree, I will try and test it on master Monday but I don't see any reason why it would not work.

matsduf:
> > We should also provide a patch for the current release. Could this PR also work directly on current master?
> 
> I agree, I will try and test it on master Monday but I don't see any reason why it would not work.

My question should rather be: Could the changes in this PR be done on current master? In the patch we do not want anything else that has been merged in develop branch.

blacksponge:
> My question should rather be: Could the changes in this PR be done on current master? In the patch we do not want anything else that has been merged in develop branch.

They are, see https://github.com/zonemaster/zonemaster-backend/pull/850

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -96,6 +98,9 @@ Readonly my $RELAXED_DOMAIN_NAME_RE => qr/^[.]$|^.{2,254}$/;
 Readonly my $TEST_ID_RE             => qr/^[0-9a-f]{16}$/;
 Readonly my $USERNAME_RE            => qr/^[a-z0-9-.@]{1,50}$/i;
 
+# Boolean 0 or 1
+Readonly my $BOOL_RE            => qr/^(0|1)$/;
```
mattias-p, line 102: What do you think about making the values `true` and `false`? That way the configuration file itself is more readable. (I believe the best way to implement it would be to have untaint_bool perform the conversion to 1 and 0.)

blacksponge: I have extended the accepted values for boolean to support `true`, `false`, `yes`, `no`, `1` and `0`.

mattias-p: Could you update the configuration documentation too?

docs/Configuration.md
```diff
@@ -51,6 +51,23 @@ Repeating a key name in one section is forbidden.
...
 
+## API section
+
+Available keys: `enable_batch_jobs`, `enable_add_api_user`.
+
+### enable_batch_jobs
+
+Boolean value to enable the batch jobs methods of the API.
+
+Accpected values: 0 or 1, default to 0 (disabled).
```
mattias-p, line 62: Could you update this too after the change to the allowed values?

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -96,6 +98,11 @@ Readonly my $RELAXED_DOMAIN_NAME_RE => qr/^[.]$|^.{2,254}$/;
 Readonly my $TEST_ID_RE             => qr/^[0-9a-f]{16}$/;
 Readonly my $USERNAME_RE            => qr/^[a-z0-9-.@]{1,50}$/i;
 
+# Boolean
+Readonly my $BOOL_TRUE_RE           => qr/^(true|yes|1)$/i;
+Readonly my $BOOL_FALSE_RE          => qr/^(false|no|0)$/i;
```
mattias-p, line 103: Would you consider removing the 1 and 0? I think it's a mistake to have both integers and booleans together using numeric representations.

mattias-p (commented):
I think this looks good. I have a couple of remarks.

docs/Configuration.md
blacksponge: Done

lib/Zonemaster/Backend/Validator.pm
blacksponge: Done

mattias-p (commented):
It seems the scope of this PR has changed a bit. Could you update its title to better reflect what it does now?

I'm not sure if it's good or bad to have both "true"/"false" and "yes"/"no" be valid values for booleans. I'm leaning towards bad. I'd like to hear more opinions on that. And I'm sorry for dwelling on this.

matsduf:
I do not think it is needed to have both "yes" ("no") and "true" ("false") but it will work fine too.

matsduf (changes requested):
I do not see the need of the changes to `docs/upgrade_db_zonemaster_backend_ver_7.0.0.md`. That is already included in the installation instructions.

blacksponge:
> 
> 
> I do not see the need of the changes to `docs/upgrade_db_zonemaster_backend_ver_7.0.0.md`. That is already included in the installation instructions.

see https://github.com/zonemaster/zonemaster-backend/pull/850#issuecomment-914060935

matsduf:
> > I do not see the need of the changes to `docs/upgrade_db_zonemaster_backend_ver_7.0.0.md`. That is already included in the installation instructions.
> 
> see [#850 (comment)](https://github.com/zonemaster/zonemaster-backend/pull/850#issuecomment-914060935)

See https://github.com/zonemaster/zonemaster-backend/pull/850#issuecomment-915021194

docs/Configuration.md
```diff
@@ -51,6 +51,25 @@ Repeating a key name in one section is forbidden.
 
 Each section in `backend_config.ini` is documented below.
 
+## RPCAPI section
+
+Available keys: `enable_batch_jobs`, `enable_add_api_user`.
```
matsduf, line 56: To be consistent, shouldn't the first key be `enable_add_batch_jobs`. Also see comment in #850.

share/backend_config.ini
```diff
@@ -31,6 +31,12 @@ database_file = /var/lib/zonemaster/db.sqlite
 #
 #maximal_number_of_retries=3
 
+[RPCAPI]
+
+# Uncomment to enable the user and batch RPCAPI methods.
+# enable_add_api_user = yes
+# enable_batch_jobs = yes
+
```
matsduf, line 39: To be consistent with #850 and comments there:
```
# Uncomment to enable API method "add_api_user"
# enable_add_api_user = yes
# Uncomment to disable API method "add_batch_job"
# enable_batch_jobs = yes
```

matsduf (changes requested):
Now we have a risk that this and #850 diverge. I suggest #850 is completed first, and then this is adjusted to that. The content of this is what will end up in next full release.

blacksponge:
I adjusted this PR with the changes in #850. Please rereview

matsduf:
@blacksponge, does this add the same changes as #850, except that `Changes` should not be updated here (updated in #870)?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -610,6 +638,8 @@ UNITCHECK {
     _create_setter( '_set_ZONEMASTER_number_of_processes_for_frontend_testing', '_ZONEMASTER_number_of_processes_for_frontend_testing', \&untaint_strictly_positive_int );
     _create_setter( '_set_ZONEMASTER_number_of_processes_for_batch_testing',    '_ZONEMASTER_number_of_processes_for_batch_testing',    \&untaint_non_negative_int );
     _create_setter( '_set_ZONEMASTER_age_reuse_previous_test',                  '_ZONEMASTER_age_reuse_previous_test',                  \&untaint_strictly_positive_int );
+    _create_setter( '_set_RPCAPI_enable_add_api_user',                          '_RPCAPI_enable_add_api_user',                          \&untaint_bool );
+    _create_setter( '_set_RPCAPI_enable_add_batch_job',                         '_RPCAPI_enable_add_batch_job',                         \&untaint_bool );
 }
```
matsduf, line 643: Is this equivalent to the following code in #850?
```
sub _set_RPCAPI_enable_add_api_user {
    my ( $self, $value ) = @_;

    $value = untaint_bool( $value ) // die "Invalid value for RPCAPI.enable_add_api_user: $value\n";
    $self->{_RPCAPI_enable_add_api_user} = $value;
    return;
}

sub _set_RPCAPI_enable_add_batch_job {
    my ( $self, $value ) = @_;

    $value = untaint_bool( $value ) // die "Invalid value for RPCAPI.enable_add_batch_job: $value\n";
    $self->{_RPCAPI_enable_add_batch_job} = $value;
    return;
}
```

share/backend_config.ini
```diff
@@ -31,6 +31,13 @@ database_file = /var/lib/zonemaster/db.sqlite
 #
 #maximal_number_of_retries=3
 
+[RPCAPI]
+
+# Uncomment to enable API method "add_api_user"
+#enable_add_api_user = yes
+# Uncomment to enable API method "add_batch_job"
+#enable_add_batch_job = yes
```
matsduf, line 39: It should be
```
# Uncomment to disable API method "add_batch_job"
#enable_add_batch_job = no
```

matsduf (commented):
A few adjustments are needed to match #850.

docs/Configuration.md
```diff
@@ -51,6 +51,25 @@ Repeating a key name in one section is forbidden.
...
+## RPCAPI section
+
+Available keys: `enable_add_batch_job`, `enable_add_api_user`.
+
+### enable_add_batch_job
+
+Boolean value to enable the `enable_add_batch_job` methods of the API.
+
+Accepted values: `yes` (or `true`) or `no` (or `false`),
+default to `no` (disabled).
```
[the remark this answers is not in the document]
matsduf: This is not handled. It must be `yes`.

docs/Configuration.md
```diff
@@ -51,6 +51,25 @@ Repeating a key name in one section is forbidden.
...
+
+### enable_add_api_user
+
+Boolean value to enable the `add_api_user` method of the API.
+
+Accepted values: `yes` (or `true`) or `no` (or `false`),
+default to `yes` (enabled).
+
```
matsduf, line 71: This should be `no`.

lib/Zonemaster/Backend/Config.pm
matsduf: Any comment here?

matsduf (changes requested):
One change in the wrong place.

lib/Zonemaster/Backend/Config.pm
blacksponge: yes it is

matsduf (approved):

matsduf:
## Testing v2021.2
LGTM.

This has been tested several time during review and after.

Issue #840, opened 2021-08-05
Choose between "test_start_time" vs "creation_time" to compute the age of a test


pnax:
Currently SQLite uses `test_start_time` (see [SQLite.pm (develop)](https://github.com/zonemaster/zonemaster-backend/blob/develop/lib/Zonemaster/Backend/DB/SQLite.pm#L185..L192)):

```perl
# Search for recent test result with the test same parameters, where "$seconds"
# gives the time limit for how old test result that is accepted.
my ( $recent_hash_id ) = $dbh->selectrow_array(
    "SELECT hash_id FROM test_results WHERE params_deterministic_hash = ? AND test_start_time > DATETIME('now', ?)",
    undef,
    $fingerprint,
    "-$seconds seconds"
);
```

But MySQL and PostgreSQL use `creation_time`:
* [MySQL.pm (develop)](https://github.com/zonemaster/zonemaster-backend/blob/develop/lib/Zonemaster/Backend/DB/MySQL.pm#L171..L176):
    ```perl
    my ( $recent_hash_id ) = $dbh->selectrow_array(
        q[
        SELECT hash_id FROM test_results WHERE params_deterministic_hash = ? AND (TO_SECONDS(NOW()) - TO_SECONDS(creation_time)) < ?
        ],
        undef, $fingerprint, $seconds_between_tests_with_same_params,
    );
    ```

* [PostgreSQL.pm (develop)](https://github.com/zonemaster/zonemaster-backend/blob/develop/lib/Zonemaster/Backend/DB/PostgreSQL.pm#L178..L185):
    ```perl
    my $sth = $dbh->prepare( "
        INSERT INTO test_results (batch_id, priority, queue, params_deterministic_hash, params, undelegated)
        SELECT ?, ?, ?, ?, ?, ?
        WHERE NOT EXISTS (
            SELECT * FROM test_results
            WHERE params_deterministic_hash = ?
              AND creation_time > NOW() - ?::interval
    )" );
    ```

It would be nice to use the same value for all database engines.

Now which one should we choose?

pnax:
> Now which one should we choose?

The [current API documentation (develop)](https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/API.md#api-method-start_domain_test) says:

```
If an identical test was already enqueued and hasn't been started or
was enqueued less than 10 minutes earlier, no new test is enqueued. 
```

[...]

Issue #832, opened 2021-07-27
Rename column "params_deterministic_hash" into "fingerprint"


pnax:
With #798 we changed all code occurrences of "params deterministic hash" into "fingerprint". Since we are updating the database with #826 I suggest that we also update the column name "params_deterministic_hash" in the database to "fingerprint".

This will be a breaking change.

matsduf:
I think that could be fine, but then we should investigate if there are more such changes that should be done at the same time. And make sure that all three engines are synchronize. Migration scripts are of course necessary.

matsduf:
This has been resolved.

Issue #331, opened 2017-12-14
The current implementation of the create_batch_job bulk API method needs to be improved


mtoma:
Its current implementation is dangerous for the DB under certain very specific conditions

mtoma:
Related to:
https://github.com/dotse/zonemaster-backend/issues/318

sandoche2k:
Not able to remember the source of this issue. Hence closed.

PR #841, opened 2021-08-05
Harmonize databases

pnax:
## Purpose

This PR tries to improve and add consistency between database engines. There are PRs to update the database in a way that breaks things (new `undelegated` column in PosgreSQL for instance). It might be nice to take the opportunity of this new Backend release to improve and fix some behaviors in the databases.

A first part in this PR is to improve the `create_db()` routine in order to be able to call it at daemon startup (which will be handle in another PR #846 ).

## Context

Based on top of #833 and #834
Addresses #832, #840

## Changes

There are several changes. This PR can easily be split in smaller PRs.

* (1/11) Create indexes in SQLite -> #865

* (2/11) Always check that a table, index or trigger does not exists before creation. For some cases, the clause `IF NOT EXISTS` is sufficient, for others, we need workaround because it is not present in all supported versions of the database engine. (069732e) -> #866 

* (3/11) Rename column `params_deterministic_hash` into `fingerprint` (26c0511)

* (4/11) Correctly use `creation_time` in SQLite to check the age of a test (a8a4cee)

* (5/11) Set `test_start_time` to NULL by default and when the test is rescheduled (6aafdeb)

* (6/11) Move the `schedule_for_retry` routine to "DB.pm" (it now uses a SQL syntax that is valid for all database engines) (29e7eb2)

* (7/11) Fix a nonexistent index key field in PosgreSQL (7145d4f)

* (8/11) Add a new `domain` column in PostgreSQL to be consistent with other database engines (4e4d563)

* (9/11) Update the "users" table: (8563602)
   * new `username` and `api_key` columns for PostgreSQL
   * remove the `user_info` column for SQLite, MySQL and PostgreSQL

* (10/11) Made some refactoring: (576dc86)
    * New method `recent_test_hash_id' to search for pre-existing test
    * Move `create_new_batch_job' method to DB.pm
    * Move `get_test_params' method to DB.pm
    * Use generic DBI `last_insert_id'

* (11/11) Compute the `hash_id` before its insertion in database in order to use the same `create_new_test` method across database engines (60da83a)

* For all the changes that affect the database, each patch script has been updated accordingly
 
 

## How to test this PR

Do one of the following:
* upgrade an existing database with the corresponding patch script
* install and initialize the database with the instructions in the [installation document](https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/Installation.md)

Then check for the following things:
* SQLite `test_results` table has indexes, `PRAGMA index_info(test_results)` should result in 5 indexes
* there is no `params_deterministic_hash` column, it should be named `fingerprint`
* there is no index `test_results__params_deterministic_hash`, it should be named `test_results__fingerprint`
* for PostgreSQL, a column `domain` should exist in the `test_results` table (if migrating, the field should be properly initialized)
* for PostgreSQL, the columns `username` and `api_key` should exist in the `users` table (if migrating, the fields should be properly initialized)
* the `hash_id` field can't be null and has no default value
* there should not be any `user_info` field in the `users` table anymore

Finally run some checks, you should be able to properly use the following methods:
* `start_domain_test` (and you should be able to reuse a pre-existing test)
* `add_api_user`
* `add_batch_job`
* `get_test_params`

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -17,7 +17,6 @@ use Zonemaster::Backend::Errors;
 requires qw(
   add_batch_job
   create_db
-  create_new_test
   from_config
   get_test_history
   process_unfinished_tests_give_up
@@ -111,6 +110,48 @@ sub add_api_user {
         unless ( $result );
 
     return $result;
+}
+
+# Standard SQL, can be here
+sub create_new_test {
+    my ( $self, $domain, $test_params, $seconds_between_tests_with_same_params, $batch_id ) = @_;
+
+    my $dbh = $self->dbh;
+
+    $test_params->{domain} = $domain;
+
+    my $fingerprint = $self->generate_fingerprint( $test_params );
+    my $encoded_params = $self->encode_params( $test_params );
+    my $undelegated = $self->undelegated ( $test_params );
+
+    my $hash_id;
+
+    my $priority    = $test_params->{priority};
+    my $queue_label = $test_params->{queue};
+
+    my $recent_hash_id = $self->recent_test_hash_id( $seconds_between_tests_with_same_params, $fingerprint );
+
+    if ( $recent_hash_id ) {
+        # A recent entry exists, so return its id
+        $hash_id = $recent_hash_id;
+    }
+    else {
+        $hash_id = substr(md5_hex(time().rand()), 0, 16);
+        $dbh->do(
+            "INSERT INTO test_results (hash_id, batch_id, priority, queue, fingerprint, params, domain, undelegated) VALUES (?,?,?,?,?,?,?,?)",
+            undef,
+            $hash_id,
+            $batch_id,
+            $priority,
+            $queue_label,
+            $fingerprint,
+            $encoded_params,
+            $test_params->{domain},
+            $undelegated,
+        );
+    }
+
+    return $hash_id;
 }
 
 # Standard SQL, can be here
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -216,6 +216,19 @@ sub process_dead_test {
     }
 }
 
+=head2 schedule_for_retry
+
+For the test with the given "hash_id" increments its number of retries by 1,
+resets its progress to 0 and its start time to NULL.
+
+=cut
+
+sub schedule_for_retry {
+    my ( $self, $hash_id ) = @_;
+
+    $self->dbh->do("UPDATE test_results SET nb_retries = nb_retries + 1, progress = 0, test_start_time = NULL WHERE hash_id=?", undef, $hash_id);
+}
+
 # A thin wrapper around DBI->connect to ensure similar behavior across database
 # engines.
 sub _new_dbh {
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -14,7 +14,6 @@ use Log::Any qw( $log );
 use Zonemaster::Engine::Profile;
 
 requires qw(
-  add_api_user_to_db
   add_batch_job
   create_db
   create_new_batch_job
@@ -26,8 +25,6 @@ requires qw(
   select_unfinished_tests
   test_progress
   test_results
-  user_authorized
-  user_exists_in_db
   get_relative_start_time
 );
 
@@ -114,6 +111,50 @@ sub add_api_user {
         unless ( $result );
 
     return $result;
+}
+
+# Standard SQL, can be here
+sub user_exists_in_db {
+    my ( $self, $user ) = @_;
+
+    my $dbh = $self->dbh;
+    my ( $id ) = $dbh->selectrow_array(
+        "SELECT id FROM users WHERE username = ?",
+        undef,
+        $user
+    );
+
+    return $id;
+}
+
+# Standard SQL, can be here
+sub add_api_user_to_db {
+    my ( $self, $user_name, $api_key  ) = @_;
+
+    my $dbh = $self->dbh;
+    my $nb_inserted = $dbh->do(
+        "INSERT INTO users (username, api_key) VALUES (?,?)",
+        undef,
+        $user_name,
+        $api_key,
+    );
+
+    return $nb_inserted;
+}
+
+# Standard SQL, can be here
+sub user_authorized {
+    my ( $self, $user, $api_key ) = @_;
+
+    my $dbh = $self->dbh;
+    my ( $id ) = $dbh->selectrow_array(
+        "SELECT id FROM users WHERE username = ? AND api_key = ?",
+        undef,
+        $user,
+        $api_key
+    );
+
+    return $id;
 }
 
 # Standard SQL, can be here
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -6,22 +6,22 @@ use Moose::Role;
 
 use 5.14.2;
 
-use JSON::PP;
 use Digest::MD5 qw(md5_hex);
 use Encode;
+use JSON::PP;
 use Log::Any qw( $log );
 
 use Zonemaster::Engine::Profile;
+use Zonemaster::Backend::Errors;
 
 requires qw(
   add_batch_job
   create_db
-  create_new_batch_job
   create_new_test
   from_config
   get_test_history
-  get_test_params
   process_unfinished_tests_give_up
+  recent_test_hash_id
   select_unfinished_tests
   test_progress
   test_results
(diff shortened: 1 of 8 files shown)


matsduf:
> Addresses #831, #840

#331 is a pull request. I expect a pull request to address issues. If it is something mentioned in the PR, then please create an issue pointing that out, or include that in this PR.

pnax:
> I expect a pull request to address issues.

Oops, my mistake. I wanted to point at #832. I updated the description.

pnax:
I made some updates and improved the git history so that each chunk can be easily identified.

pnax:
I made some tests and fixed some errors in the patch scripts.

pnax:
I added a new `domain` column for PostgreSQL since it is present in SQLite and MySQL.
I also fixed an update I missed with the `undelegated` index.

pnax:
I updated the "users" table by adding the "username" and "api_key" columns in PostgreSQL and removing the "user_info" column for all database engine.
I am continuing to add changes here in order to make testing somehow easier (the patch script needs to be tested once for instance).

pnax:
Fixed an error with the new `domain` column with batch jobs.

pnax:
I updated how we create the `hash_id`. We don't rely on database engine to generate it so that the `create_new_test` method can be the same for all databases. Furthermore this allows to remove the need to retrieve the `hash_id` after the insertion, since it is already known (hence the removal of the database lock).

mattias-p:
This PR is growing pretty huge. I'm not sure how to review it. Do you think you could break it up into smaller parts?

pnax:
I tried to make it easier to review by linking intermediary commits.
Yes I can totally break it up in smaller PRs (one for each change I think). However since it is based on top of #833 and #834, I think it will be easier to wait for these 2 PRs to be merged first. Then I can open a new PR for each change. Is that okay for you?

You can also see each intermediary change within the git history with the following commands:
```
git checkout pnax/harmonize-database
git log --oneline --graph develop..
```

pnax:
This PR is now ready for review. Please have a look @blacksponge, @matsduf and @mattias-p

matsduf:
This PR is very complex to review.

matsduf:
I really think it is good that the databases are harmonized, but now we need a definition of the fields. Specifically I wonder how the following fields are defined:
* creation_time
* test_start_time
* test_end_time

pnax:
> This PR is very complex to review.

I started to split this PR. See #865 for a start.

matsduf:
> I started to split this PR. See #865 for a start.

Does that mean that we should not review this PR? Will this PR just be closed?

pnax:
> Does that mean that we should not review this PR? Will this PR just be closed?

Well, not really since splitting the PR will result in a lot of similar testing. So I'd like to keep this one because of the complete "How to test this PR" section.

And it seems that Travis is unhappy in #865. So maybe splitting it will require more work because I need to make the splits where Travis will be happy.

pnax:
> Specifically I wonder how the following fields are defined:

The way I see it (and implemented it):

* `creation_time`: time when the test is created (added to the database)
* `test_start_time`: time when the test is started, ie when the TestAgent begin using it. This value is NULL until the TestAgent get the associated test from database
* `test_end_time`: time when the test ends (successfullly or not, the TestAgent has finished with this test). This value is NULL until the end

matsduf:
>     * `creation_time`: time when the test is created (added to the database)
> 
>     * `test_start_time`: time when the test is started, ie when the TestAgent begin using it. This value is NULL until the TestAgent get the associated test from database
> 
>     * `test_end_time`: time when the test ends (successfullly or not, the TestAgent has finished with this test). This value is NULL until the end

That looks fine, and I think we should have such documentation for the tables.

When the ages is determined to find out if the result can be "reused" then `test_start_time` is the relevant time, not `creation_time`.

pnax:
> When the ages is determined to find out if the result can be "reused" then `test_start_time` is the relevant time, not `creation_time`.

Well, you could create a test (`creation_time` is then set and `test_start_time` is NULL). The TestAgent has a small queue. So you'll have to wait a little bit for the test to be run. But not a long wait. Then you send the same query to the RPCAPI. If we check on the `test_start_time`, we compare with a NULL value, so the test might be created again. I don't think this is something we want here.

> That looks the wrong way for me.

This might be wrong but it is aligned with PostgreSQL and MySQL, which was the primary goal. (even though I don't think this is particularly wrong to check against the `creation_time`).

matsduf:
> This might be wrong but it is aligned with PostgreSQL and MySQL, which was the primary goal. (even though I don't think this is particularly wrong to check against the `creation_time`).

We should have a specification to match against, not the implementations.

If we check against `creation_time` and the match test has a lower priority we can have to wait a long time. If a new test is requested it is OK to reuse an older test if that test has started and maybe completed. `creation_time` is the wrong field to compare against.

pnax:
> We should have a specification to match against, not the implementations.

Agreed, but without such specification, I think we need to have the same implementations for all database engines.

> If we check against `creation_time` and the match test has a lower priority we can have to wait a long time. If a new test is requested it is OK to reuse an older test if that test has started and maybe completed. creation_time is the wrong field to compare against.

Yes this is something I tried to bring to light in #840. This is why I suggested to have a way to check the priority of a test (or update it in database).

matsduf:
> Yes this is something I tried to bring to light in #840. This is why I suggested to have a way to check the priority of a test (or update it in database).

If we base reuse on `start_test_time` we might get some extra tests started, but we do not risk that someone get stuck waiting for a test to be started.

What is the problem of using `start_test_time`?

matsduf:
> Agreed, but without such specification, I think we need to have the same implementations for all database engines.

Think we should have the same database layout for all engines, unless there is some restriction in the engine. Could that specification be included with this PR? I could help on that.

> Yes this is something I tried to bring to light in #840. This is why I suggested to have a way to check the priority of a test (or update it in database).

If we base reuse on `start_test_time` we might get some extra tests started, but we do not risk that someone get stuck waiting for a test to be started.

What is the problem of using `start_test_time`?

pnax:
> If we base reuse on start_test_time we might get some extra tests started

I don't think this is something we should allow.

To avoid being stuck, I really think we need to look at the priority of the test.

> What is the problem of using start_test_time?

I don't think there is a problem to use `start_test_time` per se, it's just that this needs some work to have it properly implemented regarding a specification that we need to agree on.

matsduf:
> > If we base reuse on start_test_time we might get some extra tests started
> 
> I don't think this is something we should allow.
> 
> To avoid being stuck, I really think we need to look at the priority of the test.

1. You want to look at the priority, but that would also result in an extra test started.
2. Having an extra test started now and then is not a problem. If a GUI user has to wait too long it is a problem.

> I don't think there is a problem to use `start_test_time` per se, it's just that this needs some work to have it properly implemented regarding a specification that we need to agree on.

Agree. We should have a specification.

pnax:
> 1. You want to look at the priority, but that would also result in an extra test started.

Not if we agree on updating the `priority` field as suggested [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-895301703) and [here](https://github.com/zonemaster/zonemaster-backend/pull/831#issuecomment-899552257).

matsduf:
> > 1. You want to look at the priority, but that would also result in an extra test started.
> 
> Not if we agree on updating the `priority` field as suggested [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-895301703) and [here](https://github.com/zonemaster/zonemaster-backend/pull/831#issuecomment-899552257).

We should not over-estimate the problem of an extra test being started, i.e. the same domain with the same setting being started more than once within the reuse time. The most likely source of that is when a user re-runs it via GUI. The likelyhood of being run both from batch and GUI at the same time is not likely. If a large batch is started, e.g. an entire TLD, it is a considerable problem if the same domains cannot be started from GUI.

Changing priority is a possible solution, but it is simpler, and makes more sense, to set `start_test_time` at start of test and then use that to determine if a test should be reused.

I can different solutions, but I prefer simple solution that are easy to describe and comprehend. One simple solution would be to drop the reuse feature. It has its down-sides that we would then avoid.

matsduf (commented):
I get the following, which I have not seen before. I have got it twice on some 20 tests:
```
WARNING - DBD::mysql::db selectrow_array failed: Lost connection to MySQL server during query at /usr/local/lib/perl5/site_perl/Zonemaster/Backend/DB/MySQL.pm line 211.
```

pnax:
> I get the following, which I have not seen before. I have got it twice on some 20 tests:

I rebased on top of develop, and applied the changes from #866. @matsduf could you test again to see if rebasing on top of develop fixed your issue (maybe thanks to #859)

matsduf:
> I rebased on top of develop, and applied the changes from #866. @matsduf could you test again to see if rebasing on top of develop fixed your issue (maybe thanks to #859)

I have tested the rebased branch and it looks good with SQLite and MySQL. I will test with Postgresql too.

matsduf (approved):

blacksponge (approved):

blacksponge:
Nice work!

matsduf:
@pnax, please merge.

PR #846, opened 2021-08-12
Automatically create tables on damons' startup

pnax:
## Purpose

This PR removes the need to explicitly run a script to create the tables during installation. Each daemon creates the tables on startup if they are not present.

## Context

This is following #834 and #841 which lay the groundwork for this PR.
This PR is based on top of #841 (which is based on top of other PRs).

## Changes

* Each daemon now run the code to create the tables and indexes at startup.

* Call `create_db()` for all engines before running the tests in `t/test01.t`.

* Remove the `share/create_db.pl` script that was introduced in #834

* Remove obsolete installation instructions introduced in #834

## How to test this PR

Try installing the Backend with the new set of instructions. You should only have to create the database (and the user, if applicable). The tables should be automatically created after starting any of the Zonemaster's daemons (`zm-rpcapi` or `zm-testagent`).


docs/Installation.md
```diff
@@ -138,15 +138,25 @@ sudo systemctl enable mariadb
...
+Create the database:
+```sql
+CREATE DATABASE zonemaster;
+```
+
+Create a new user and give it all permissions on the newly created database:
+```sql
+CREATE USER 'zonemaster'@'localhost' IDENTIFIED BY 'zonemaster';
+GRANT ALL ON zonemaster.* TO 'zonemaster'@'localhost';
+```
```
matsduf, line 157: This is added complexity to the installation process. It should not be made harder to install Zonemaster-Backend.

docs/Installation.md
```diff
@@ -138,15 +138,25 @@ sudo systemctl enable mariadb
...
+
+> Update the `backend_config.ini` file accordingly.
```
matsduf, line 159: Why is that need now when it has not been needed before. It is also unclear what to update.

docs/Installation.md
```diff
@@ -183,15 +193,17 @@ Install, configure and start database engine:
...
 
 ```sh
-sudo -u postgres psql -f $(perl -MFile::ShareDir=dist_dir -E 'say dist_dir("Zonemaster-Backend")')/initial-postgres.sql
+sudo -u postgres createuser -P zonemaster
 ```
 
-> **Note:** This creates a database called `zonemaster`, as well as a user called
-> "zonemaster" with the password "zonemaster" (as stated in the config file).
-> This user has just enough permissions to run the backend software.
+This will ask for a password that should be copied in `backend_config.ini`.
```
matsduf, line 202: Why is that needed now when it was not needed before?

matsduf (changes requested):

mattias-p:
I've also been working on database initialization but in a completely different direction. I've been working towards removing privileges from the daemons so they cannot modify the database schema. The idea is to have a CLI tool for administrative tasks such as initializing or upgrading the database schema or querying the current database schema version. This tool could be run with sudo to gain the necessary privileges.

I like this approach better because it's easier for the user to know what state the system is in when things aren't done magically behind the curtains (e.g. a schema upgrade when you start a daemon). You also don't run the risk of having two daemons trying to initialize or upgrade the schema at the same time.

Any thoughts on this?

pnax:
Having a tool for administrative tasks sounds like a nice feature indeed. In a way this was my first approach in #834. However I changed my mind as I wasn't satisfied with my solution. I found it too complex, it needs to support all OSes, all database engines and it would be even better if it can be easily tweaked to only create the schemas and not the database and/or user (see https://github.com/zonemaster/zonemaster-backend/pull/834#issuecomment-888914080). So I chose to split the process in two. One part to create the database, the other one to create the schemas.

Now remains, how and when do we create the database and the schemas?
Looking at other projects it seems fair to let the user create the database and let the program initialize it. This is what I chose to do here. I found it easier compared to developing a new tool.

matsduf:
@pnax, can you resolve the conflicts and review the PR?

pnax:
This PR is based on top of #841. I am making this one a draft for now until #841 is approved.

pnax:
Now that #841 is merged, this has been rebased on top of develop and I set it ready for reviews.

script/zonemaster_backend_rpcapi.psgi
```diff
@@ -66,7 +66,10 @@ builder {
         my $app = shift;
 
         # Make sure we can connect to the database
-        $config->new_DB();
+        my $db = $config->new_DB();
+
+        $log->notice( "Initializing the database" );
+        $db->create_db();
```
mattias-p, line 72: Could we detect if the database is initialized and only initialize it if it isn't? I don't think it's helpful to have log lines that claim that the database was initialized when in fact nothing was changed. The same goes for zonemaster_backend_testagent. Also there seems to be a race condition where multiple daemons compete to perform the same initialization.

mattias-p (commented):
Do you have a plan for how database migrations should tie into this solution? If those are performed with separated commands by the administrator, wouldn't it make sense to perform the initialization manually too?

script/zonemaster_backend_testagent
```diff
@@ -197,7 +197,11 @@ eval {
     Zonemaster::Backend::TestAgent->new( { config => $initial_config } );
 
     # Connect to the database
-    $initial_config->new_DB();
+    my $db = $initial_config->new_DB();
+
+    $log->notice( "Initializing the database" );
+    $db->create_db();
```
matsduf, line 203: Are both daemons supposed to create the database? Isn't it more reasonable to let one of them do the job before forking, and then exit and ask the operator to restart both daemons?

matsduf:
Today there is a short script `create_db.pl` that you run when you create or reset the Zonemaster database. That script is run when the daemons are not running. In many cases you will run `create_db.pl` less often than running the upgrade scripts.

What is the gain of having the database tables being created magically by the daemon instead of having the a separate create script? We can strip a few lines in the installation instructions. Those lines are quite simple and straight forward.

What will be easiest to maintain? The script or the function in the daemon?

script/zonemaster_backend_rpcapi.psgi
pnax: I don't think we currently can. And I agree that this would be better to have this kind of detection. This also goes with your question about whether I have a plan about database migrations. Since database migration is about running a Perl script, I would like to have them being run automatically after an update. This would require to be able to check the version of the previous installed Backend and compare it to the newly installed version. Then we can automatically run the corresponding upgrade routines. However I haven't found a solution on how to compare these versions using our current code yet. Maybe we need to store it in database so that we can perform this comparison.
If we have a way to compare versions, then we will be able to detect if the database is initialized or not.

About the race condition, I have to review the code, but I think that all clauses using `IF NOT EXISTS` should not be affected.

matsduf: > Maybe we need to store it in database so that we can perform this comparison. If we have a way to compare versions, then we will be able to detect if the database is initialized or not.

I definitely think that we should store the database structure version in the database.

mattias-p: > Also there seems to be a race condition where multiple daemons compete to perform the same initialization.
>
> > About the race condition, I have to review the code, but I think that all clauses using IF NOT EXISTS should not be affected.

Actually my point is more about the architecture than the implementation.

All daemons will attempt to perform every part of the initialization. Not only in this version but in every future version. Since I think we agree that schema initialization should be performed using the same approach, it needs to work for upgrades too.

The fewer failure modes I have to worry about during schema upgrades (and initialization too) the happier I am. I would really prefer to have these operations be performed in a friendlier single threaded context.

docs/Installation.md
pnax: not related to this PR

docs/Installation.md
pnax: not related to this PR

docs/Installation.md
pnax: not related to this PR

pnax:
The proposed implementation is too light and error prone (possible race condition). There is currently no gain in merging it. I'm closing it. But the idea to be able to easily prepare the database and upgrade it between releases using an administration tool is interesting and might need to be explored in a future work.

matsduf:
@pnax, I agree that the idea was nice. :-)

PR #844, opened 2021-08-11
Add metrics

blacksponge:
## Purpose

Add application level metrics to enable better observability.

## Changes

Use Net::Statsd to send metrics to a statsd compatible receiver. I chose a pushed based metrics system over a pull based metrics collection (e.g. Prometheus / Opentelemetry) because both the API and test agent are forking, making it hard to aggregate the metrics.

The change is purely optional and it is not required to installed any dependencies for users that don't need this feature.

NOTE: This PR is still a proof of concept, I wanted some insight before going further with it.

## How to test this PR

* Install Net::Statsd `cpanm Net::Statsd` 
* Set / export the variables `ZM_STATS_HOST` and `ZM_STATS_PORT` to the statsd receiver host and port (upd). The port can be omitted in which case it will default to 8125.
* Start the API and Test Agent.

Using the [statsd exporter](https://github.com/prometheus/statsd_exporter/releases) as receiver with the following mapping:
```yaml
mappings:
- match: zonemaster.rpcapi.requests.*.*
  name: zonemaster_rpcapi_requests_total
  labels:
    method: $1
    status: $2
```
we should see some metrics starting with `zonemaster_*`, e.g.:
```
# HELP zonemaster_rpcapi_requests_total Metric autogenerated by statsd_exporter.
# TYPE zonemaster_rpcapi_requests_total counter
zonemaster_rpcapi_requests_total{method="get_test_results",status="RPC_INTERNAL_ERROR"} 1
zonemaster_rpcapi_requests_total{method="get_test_results",status="RPC_SUCCESS"} 1
zonemaster_rpcapi_requests_total{method="start_domain_test",status="RPC_SUCCESS"} 3
zonemaster_rpcapi_requests_total{method="test_progress",status="RPC_SUCCESS"} 63
# HELP zonemaster_testagent_maximum_processes Metric autogenerated by statsd_exporter.
# TYPE zonemaster_testagent_maximum_processes gauge
zonemaster_testagent_maximum_processes 40
# HELP zonemaster_testagent_running_processes Metric autogenerated by statsd_exporter.
# TYPE zonemaster_testagent_running_processes gauge
zonemaster_testagent_running_processes 1
# HELP zonemaster_testagent_tests_completed Metric autogenerated by statsd_exporter.
# TYPE zonemaster_testagent_tests_completed counter
zonemaster_testagent_tests_completed 1
# HELP zonemaster_testagent_tests_died Metric autogenerated by statsd_exporter.
[...]

PR #843, opened 2021-08-11
use logger instead of warn

blacksponge:
## Purpose

Get better log files.

## Changes

Use logger instead of warn.

## How to test this PR

```
./script/zmb get_test_results --test-id 123456789abcdef0 --lang fr
```
This trigger an exception (it should not but that for another PR), check that the message is correctly formatted:
```
2021-08-11T10:31:23Z [63992] ERROR - Internal error 012: Unexpected error in the get_test_results API call: [malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /home/gbm/workspace/zonemaster/zonemaster-backend/lib/Zonemaster/Backend/DB/PostgreSQL.pm line 235.]
```

Issue #819, opened 2021-07-09
Application specific error codes


mattias-p:
The JSONRPC specification allows applications to specify their own error codes. Today all handler methods in the RPCAPI module catch exceptions at their top level and generate errors with the generic code -32603 (Internal error).

We should:
* Pick a small set of error states that we want to generate application defined error responses for.
* Update the "Error reporting" section in the API document to include those error states.

For each error state we should also:
1. Create a class method named after its message string in lower snake case. This method should return a data structure that represents an JSONRPC error object for this error state. If the error state has a described data type, the method should take appropriate arguments so the data member of the error object can be properly populated.
2. Make sure this error state throws an exception that can be recognized in a reliable way.
3. Determine all handler methods where this exception can end up.
4. Update exactly those handler methods to recognize the exception. When such an exception is recognized, the class method should be used to create an error object so the application defined error response can be returned instead of the generic -32603 response.

matsduf:
@mattias-p and @blacksponge, how much of this issue has been resolved by #847? Could this issue be updated to reflect that?

blacksponge:
#847 adds all the logic for application error codes but because JSON::RPC can not handle custom error code it does not work (related PR https://github.com/lestrrat-p5/JSON-RPC/pull/14)

matsduf:
As I understand it, #847 has resolved as much as is possible withing the frame work. If so, I suggest that we close this issue.

mattias-p:
I created #917 covering the remaining parts of this issue.

matsduf:
In most parts resolved by #847. The remainder moved to #917.

PR #847, opened 2021-08-12
Add custom error handling

blacksponge:
## Purpose

Make error management a bit more manageable. 

## Context

Partially addresses #819 (with limitations due to downstream dependencies; codes definition is not covered).

## Changes

Built on top of #843, planned to be integrated with #844 to provide server error counters.

Adds a base error type `Zonemaster::Backend::Error` extended to the following classes:
* `Zonemaster::Backend::Error::Internal` (covers code -32603, but subclassed to add more meaning when applicable)
  * `Zonemaster::Backend::Error::JsonError`
* `Zonemaster::Backend::Error::ResourceNotFound` (custom)

This type is used to differentiate *actual* unexpected server error and "normal" exceptions (a test that does not exist in the database should not result in a server error, even less in a json parsing error). Also I noticed that sometimes `null` is returned when a test is not found (like with `get_test_progress`) and sometimes an error is returned, what should be the correct behavior? Returning a custom error seems to be an accepted pattern from what I have seen in other implementations, it can also have some context about the error.

It also adds the possibility to add meaning and context to the error, both to the operator, in the logs, and to the user, in the response.

It theoretically adds the ability to set a custom error code but that is not yet usable due to a limitation of the [JSON::RPC](https://github.com/lestrrat-p5/JSON-RPC) module. I have made a [pull request](https://github.com/lestrrat-p5/JSON-RPC/pull/14) to address that matter but the maintainer of the repo that not seem active anymore (the package on cpan has been marked for adoption).

If / when custom code are available in the JSON RPC module, the parameter validation should be moved after the method routing, relying on the external module to parse and validate the method (what it is already doing).

## How to test this PR

```sh
% ./script/zmb get_test_params --test-id a260536b98076a81 | jq
{
  "jsonrpc": "2.0",
  "error": {
    "message": "Test not found",
    "data": {
      "test_id": "a260536b98076a81"
    },
    "code": -32603
  },
  "id": 1
}
# Should be converted to a application error
% ./script/zmb add_batch_job --domain example.com --username toto --api-key notsecret | jq
{
  "error": {
    "code": -32603,
    "data": null,
  },
  "id": 1,
  "jsonrpc": "2.0"
}
# Test result existence has been disable to show case a JsonError
% ./script/zmb get_test_results --test-id 123456789abcdef0 --lang fr | jq {
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32603,
    "data": {
      "test_id": "123456789abcdef0"
    },
  }
}
```
Generated logs:
```
2021-08-12T13:58:45Z [47491] ERROR - Internal error 015 (Zonemaster::Backend::Error::Internal): Unexpected error in the `Zonemaster::Backend::RPCAPI::add_batch_job` method: [User toto not authorized to use batch mode]
2021-08-12T13:58:49Z [47492] INFO - Test not found (code -32000); Context: {'test_id' => 'a260536b98076a81'}
2021-08-12T13:59:07Z [47493] ERROR - Internal error 000 (Zonemaster::Backend::Error::JsonError): Unexpected error in the `Zonemaster::Backend::DB::SQLite::test_results` method: [malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "(end of string)") at /home/gbm/workspace/zonemaster/zonemaster-backend/lib/Zonemaster/Backend/DB/SQLite.pm line 276.]; Context: {'test_id' => '123456789abcdef0'}
```

### DIFF
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -10,6 +10,7 @@ use DBI qw(:utils);
 use JSON::PP;
 
 use Zonemaster::Backend::Validator qw( untaint_ipv6_address );
+use Zonemaster::Backend::Errors;
 
 with 'Zonemaster::Backend::DB';
 
@@ -228,12 +229,14 @@ sub get_test_params {
 
     my ( $params_json ) = $self->dbh->selectrow_array( "SELECT params FROM test_results WHERE hash_id=?", undef, $test_id );
 
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Test not found", data => { test_id => $test_id } ) unless defined $params_json;
+
     my $result;
     eval {
         $result = decode_json( $params_json );
     };
 
-    warn "decoding of params_json failed (testi_id: [$test_id]):".Dumper($params_json) if $@;
+    die Zonemaster::Backend::Error::JsonError->new( reason => "$@", data => { test_id => $test_id } ) if $@;
 
     return decode_json( $params_json );
 }
@@ -249,8 +252,20 @@ sub test_results {
     my $result;
     my ( $hrefs ) = $self->dbh->selectall_hashref( "SELECT id, hash_id, CONVERT_TZ(`creation_time`, \@\@session.time_zone, '+00:00') AS creation_time, params, results FROM test_results WHERE hash_id=?", 'hash_id', undef, $test_id );
     $result            = $hrefs->{$test_id};
-    $result->{params}  = decode_json( $result->{params} );
-    $result->{results} = decode_json( $result->{results} );
+
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Test not found", data => { test_id => $test_id } ) unless defined $result;
+
+    eval {
+        $result->{params}  = decode_json( $result->{params} );
+
+        if (defined $result->{results}) {
+            $result->{results} = decode_json( $result->{results} );
+        } else {
+            $result->{results} = [];
+        }
+    };
+
+    die Zonemaster::Backend::Error::JsonError->new( reason => "$@", data => { test_id => $test_id } ) if $@;
 
     return $result;
 }
--- a/lib/Zonemaster/Backend/Errors.pm
+++ b/lib/Zonemaster/Backend/Errors.pm
@@ -1,0 +1,141 @@
+package Zonemaster::Backend::Error;
+use Moose;
+use Data::Dumper;
+
+
+has 'message' => (
+    is => 'rw',
+    isa => 'Str',
+    required => 1,
+);
+
+has 'code' => (
+    is => 'rw',
+    isa => 'Int',
+    required => 1,
+);
+
+has 'data' => (
+    is => 'rw',
+    isa => 'Any',
+    default => undef,
+);
+
+sub as_hash {
+    my $self = shift;
+    my $error = {
+        code => $self->code,
+        message => $self->message
+    };
+    $error->{data} = $self->data if defined $self->data;
+    return $error;
+}
+
+sub as_string {
+    my $self = shift;
+    my $str = sprintf "%s (code %d)", $self->message, $self->code;
+    if (defined $self->data) {
+        $str .= sprintf "; Context: %s", $self->_data_dump;
+    }
+    return $str;
+}
+
+sub _data_dump {
+    my $self = shift;
+    local $Data::Dumper::Indent = 0;
+    local $Data::Dumper::Terse = 1;
+    my $data = Dumper($self->data);
+    $data =~ s/[\n\r]/ /g;
+    return $data ;
+}
+
+package Zonemaster::Backend::Error::Internal;
+use Moose;
+
+extends 'Zonemaster::Backend::Error';
+
+has '+message' => (
+    default => 'Internal server error'
+);
+
+has '+code' => (
+    default => -32603
+);
+
+has 'reason' => (
+    isa => 'Str',
+    is => 'rw',
+    initializer => 'reason',
+);
+
+has 'method' => (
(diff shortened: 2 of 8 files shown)


lib/Zonemaster/Backend/Errors.pm
```diff
@@ -0,0 +1,160 @@
...
+
+around 'as_hash' => sub {
+    my $orig = shift;
+    my $self = shift;
+
+    my $href = $self->$orig;
+
+    $href->{exception_id} = $self->id;
+    $href->{reason} = $self->reason;
+    $href->{method} = $self->method;
```
mattias-p, line 126: Did you intend to add these into `$href->{data}` instead of into `$href` directly?

I don't think we should leak `reason` and `method` to the end user. It's great that they're being logged, but we shouldn't include them in the response.

Instead maybe we should assign a sufficiently unique id to each request and return that instead of the exception id? I'm not saying that needs to be included in this PR.

mattias-p (commented):
I like this design!

lib/Zonemaster/Backend/Errors.pm
blacksponge: No, I intended to add it to the `$href` and not data to not leak the info in the response as those properties are unused for the moment. 

This actually is anticipation work for future structured logging which, if JSON, could used the `$href` object directly. If you want I can remove it and only add it when / if needed later.

lib/Zonemaster/Backend/Errors.pm
```diff
@@ -0,0 +1,160 @@
+package Zonemaster::Backend::Error;
+use Moose;
+use Data::Dumper;
+
+
+has 'message' => (
+    is => 'rw',
```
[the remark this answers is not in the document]
blacksponge: I guess I could make them read-only; I'll try that.

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -228,12 +229,14 @@ sub get_test_params {
 
     my ( $params_json ) = $self->dbh->selectrow_array( "SELECT params FROM test_results WHERE hash_id=?", undef, $test_id );
 
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Test not found", data => { test_id => $test_id } ) unless defined $params_json;
```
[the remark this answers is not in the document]
blacksponge: Sure, I'll do that.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -51,22 +51,25 @@ sub get_db_class {
 sub user_exists {
     my ( $self, $user ) = @_;
 
-    die "username not provided to the method user_exists\n" unless ( $user );
+    die Zonemaster::Backend::Error::Internal->new( reason => "username not provided to the method user_exists")
+    unless ( $user );
```
pnax, line 55: Could add an indentation here before the `unless` to make it easier to read?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -67,18 +68,27 @@ sub _init_db {
...
+
     if ($@) {
-        handle_exception('_init_db', "Failed to initialize the [$dbtype] database backend module: [$@]", '002');
+        handle_exception('_init_db', "Failed to initialize the [$dbtype] database backend module: [$@]");
     }
 }
 
 sub handle_exception {
-    my ( $method, $exception, $exception_id ) = @_;
+    my ( $_method, $exception ) = @_;
```
pnax, line 78: Maybe you can remove the `$_method` since it is not used, my understanding is that `_build_method()` retrieve the method's name automatically.

pnax (commented):
This looks great!

lib/Zonemaster/Backend/Errors.pm
mattias-p: Ah, I see. Yeah, I think it's better to leave it out for now and reintroduce it when it's actually needed.

lib/Zonemaster/Backend/DB.pm
blacksponge: Done

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: Done

mattias-p (commented):
This looks really nice. I have one question/suggestion and a couple of formatting nits.

mattias-p (approved):
Really nice work!

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -43,7 +44,7 @@ sub new {
     bless( $self, $type );
 
     if ( ! $params || ! $params->{config} ) {
-        handle_exception('new', "Missing 'config' parameter", '001');
+        handle_exception('new', "Missing 'config' parameter");
```
pnax, line 47: Since the method name is not used anymore, is it still required to pass it to `handle_exception()`?

mattias-p (dismissed):

pnax (approved):

matsduf (changes requested):
```
$ make distcheck
Not in MANIFEST: lib/Zonemaster/Backend/Errors.pm
```

matsduf:
When MANIFEST is fixed, I will test.

matsduf (approved):
Looks fine.

mattias-p:
I've tested this for v2021.2 and it LGTM.

The last test case listed under How to test this PR didn't work out of the box because it relies on a bug that has been fixed. Instead I injected a bug into Zonemaster::Backend::DB::PostgreSQL::test_progress() that ensures that the decoded JSON blob is invalid before attempting to parse it. After this I was able to successfully run that test case too.

PR #848, opened 2021-08-12
Using decorators to handle errors

blacksponge:
## Purpose

This is very much a work in progress, built on top of #847 to generalize the way error handling is done. This could be extending to other common operation. (I included a commented example of how params validation could be achieved this way.)

If you think this could be an improvement I can finish this work.

## Context

TODO

## Changes

TODO

## How to test this PR

TODO


matsduf:
Could you write a better description? I do not get it.

blacksponge:
> Could you write a better description? I do not get it.

I try to make it clearer

matsduf:
@blacksponge, do you plan to work on this?

Issue #664, opened 2020-11-05
FreeBSD cannot switch to MariaDB


matsduf:
Zonemaster-Backend requires `DBD::mysql` but that cannot be installed together with MariaDB on FreeBSD (using FreeBSD packages).

There is `DBD::MariaDB` but it is unclear if that could be used on FreeBSD.

pnax:
> There is `DBD::MariaDB` but it is unclear if that could be used on FreeBSD.

It seems that the `DBD::MariaDB` driver is available for FreeBSD:
https://www.freshports.org/databases/p5-DBD-MariaDB/

matsduf:
If Backend does not require `DBD::mysql`, but `DBD::MariaDB` then it is likely that it will work with MariaDB sever,

matsduf:
There is no reason to keep this issue open.

PR #849, opened 2021-08-12
Switch to DBD::MariaDB driver (drop MySQL)

pnax:
## Purpose

This PR is about dropping support to MySQL and use MariaDB on all supported OSes.

I'd like to have your opinion on this.

## Context

Addresses #664.

## Changes

* For all OSes: use `DBD::MariaDB` instead of `DBD::mysql`.

* On FreeBSD: update the installation instructions to install a MariaDB server instead of MySQL.

* Provide instructions to upgrade the Backend for all OSes.

## How to test this PR

Follow the instructions in the upgrade document, restart the daemons and do some tests. The Backend should work as before, the current database should be properly accessible.


docs/Installation.md
```diff
@@ -507,7 +507,7 @@ sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= MySQL/' /usr/local/etc/zonemaster/back
 Install, configure and start database engine (and Perl bindings):
 
 ```sh
-pkg install mysql57-server p5-DBD-mysql
+pkg install mysql57-server p5-DBD-MariaDB
```
matsduf, line 510: It must be MariaDB server package too.

docs/upgrade_db_zonemaster_backend_ver_5.0.2.md
```diff
@@ -1,4 +1,4 @@
-If your zonemaster database was created by a Zonemaster-Backend version smaller than
+IF your zonemaster database was created by a Zonemaster-Backend version smaller than
```
matsduf, line 1: What is wrong with "If"?

matsduf (changes requested):
We call it MySQL but we install MariaDB. I think we should use "MariaDB" in the configuration, but make "MySQL" an alias for "MariaDB". At some time it should be deprecated.

docs/Installation.md
```diff
@@ -507,7 +507,7 @@ sed -i '' '/[[:<:]]engine[[:>:]]/ s/=.*/= MySQL/' /usr/local/etc/zonemaster/back
 Install, configure and start database engine (and Perl bindings):
 
 ```sh
-pkg install mysql57-server p5-DBD-mysql
+pkg install mysql57-server p5-DBD-MariaDB
 sysrc mysql_enable="YES"
 service mysql-server start
```
matsduf, line 512: Also here it must be updated.

matsduf:
I think the change as such is fine.

docs/upgrade_db_zonemaster_backend_ver_5.0.2.md
pnax: Nothing, I shouldn't have touched this file. I removed this update.

docs/Installation.md
pnax: fixed

docs/Installation.md
pnax: looks like this is a duplicate of https://github.com/zonemaster/zonemaster-backend/pull/849#discussion_r687908555

pnax:
> We call it MySQL but we install MariaDB. I think we should use "MariaDB" in the configuration, but make "MySQL" an alias for "MariaDB". At some time it should be deprecated.

Agreed. This will be done in a later work I think as there are a lot of places where we use "MySQL" (especially in the config file).


Travis is failing and I can't see why. It seems it comes from the LDNS installation command but this same command runs just fine for all the other jobs. Anyone has an idea why?

docs/Installation.md
matsduf: It was not me duplicating.

matsduf: No, this is
```
sysrc mysql_enable="YES"
service mysql-server start
```
It should be mariadb something.

matsduf:
> > We call it MySQL but we install MariaDB. I think we should use "MariaDB" in the configuration, but make "MySQL" an alias for "MariaDB". At some time it should be deprecated.
> 
> Agreed. This will be done in a later work I think as there are a lot of places where we use "MySQL" (especially in the config file).

Fine.

> Travis is failing and I can't see why. It seems it comes from the LDNS installation command but this same command runs just fine for all the other jobs. Anyone has an idea why?

I think it is Travis. It says "Can't link/include C library 'idna.h', 'idn', aborting.". I suggest rerun, e.g. with an empty commit.

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
```diff
@@ -30,3 +33,40 @@ cd $(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backe
...
+With this new release, we are dropping MySQL to fully use MariaDB. This applies
+to the driver as weel, [`DBD::mysql`] is replace by [`DBD::MariaDB`].
+
+### FreeBSD
+
+```sh
+pkg install mariadb105-server p5-DBD-MariaDB
+```
+
+This will ask to remove `mysql57-server` and `p5-DBD-mysql`, answer `yes`.
```
matsduf, line 48: I think this should be in the installation instruction. This file is for the upgrade of the database schema. Even if you decide to reset the database you have to do the upgrade.

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
```diff
@@ -30,3 +33,40 @@ cd $(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backe
...
+
+You may need to restart the `mysql-server`.
```
matsduf, line 50: MySQL server or MariaDB server?

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
```diff
@@ -30,3 +33,40 @@ cd $(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backe
...
+
+Check that it is properly running with:
+
+```sh
+service mysql-server status
```
matsduf, line 55: MySQL server or MariaDB server?

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
```diff
@@ -30,3 +33,40 @@ cd $(perl -MFile::ShareDir -le 'print File::ShareDir::dist_dir("Zonemaster-Backe
 perl patch_postgresql_db_zonemaster_backend_ver_7.0.0.pl
 ```
 
+
+## New dependencies
+
+With this new release, we are dropping MySQL to fully use MariaDB. This applies
+to the driver as weel, [`DBD::mysql`] is replace by [`DBD::MariaDB`].
```
mattias-p, line 40: weel -> well

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -60,7 +60,11 @@ sub from_config {
         $host = "[$host]";
     }
 
-    my $data_source_name = "DBI:mysql:database=$database;host=$host;port=$port";
+    my $data_source_name = "DBI:MariaDB:database=$database;host=$host";
+
+    if ( $host ne 'localhost' and $host ne 'embedded' ) {
```
mattias-p, line 65: The `embedded` value should be mentioned under MYSQL.host and MYSQL.port sections in Configuration.md.

pnax: After all I am wondering if we really need to check for the `embedded` value. I added it because I found it was also a value used to bypass the port configuration. Maybe I can remove it.
I also don't know how an embedded server is usually used, maybe it requires to pass option with [_mariadb_embedded_options_](https://metacpan.org/pod/DBD::MariaDB#mariadb_embedded_options) which we don't support.

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
pnax: > This file is for the upgrade of the database schema.

Right, I haven't seen that. Since we might need other packages for the next release, can't we rename this file to be more generic with a section for the database and another one for the dependencies?

docs/Installation.md
pnax: It seems not. I tried on FreeBSD 13, and installing MariaDB with `pkg install mariadb105-server` will install `mysql-server` in `/usr/local/etc/rc.d/`.
I can't find anything with `mariadb-`.
This is also the same procedure in this wiki: https://wiki.freebsd.org/MariaDB

lib/Zonemaster/Backend/DB/MySQL.pm
mattias-p: Maybe we should raise an error if MYSQL.host = embedded? I haven't tried but I have a hunch embedded won't work for Backend.

pnax: Good hunch, I get an error when setting MYSQL.host to `embedded`.
```
DBI connect(...) failed: Connection error: Embedded server is not supported
```
I'll remove this check since the DBI fails on its own. If someone needs it, we can easily re-add the condition.

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
pnax: I made a commit to rename all upgrade files to be more generic: 2407446

docs/upgrade_db_zonemaster_backend_ver_7.0.0.md
pnax: should be fixed

pnax:
Few updates:
* I rebased the PR on top of develop.
* I fixed the typo spotted by @mattias-p 
* I removed the `embedded` check
* I added notes to inform that on FreeBSD the service is called `mysql-server`.

matsduf:
On FreeBSD: p5-DBD-MariaDB requires mysql57-client and "mysql57-client-5.7.35_1 conflicts with mariadb105-client-10.5.12".

If we really want to move to MariaDB we could consider this to be a bug, and remove the support for MariaDB/MySQL for FreeBSD. I will open an issue with FreeBSD ports.

pnax:
I thought the idea was to get rid of MySQL client and server. Would you say that this is not possible? We will need to keep the MySQL server anyway?

matsduf:
> I thought the idea was to get rid of MySQL client and server. Would you say that this is not possible? We will need to keep the MySQL server anyway?

Correct. It is not possible to install MariaDB server or MariaDB client on the same computer as MariaDB DBD perl library. The MariaDB server requires MariaDB client. MariaDB DBD perl library requires MySQL client. MariaDB client and MySQL client are evil sisters that cannot coexist on the same computer under FreeBSD.

matsduf:
@pnax, could you resolve the conflicts and possibly the proposal?

pnax:
I'll need to rework on this, since I thought this would allow to get rid of MySQL. If I understand correctly this won't be the case because of how MariaDB and MySQL are packaged for FreeBSD. I guess we'll continue to have to support MySQL because of that. So I'm wondering if there is a gain to completely switch to DBD::MariaDB if we can't get rid of MySQL. I'll do an update but I really think this is a burden to support both MariaDB and MySQL.

matsduf:
> I'll need to rework on this, since I thought this would allow to get rid of MySQL. If I understand correctly this won't be the case because of how MariaDB and MySQL are packaged for FreeBSD. I guess we'll continue to have to support MySQL because of that. So I'm wondering if there is a gain to completely switch to DBD::MariaDB if we can't get rid of MySQL. I'll do an update but I really think this is a burden to support both MariaDB and MySQL.

I think we should leave things as they are now. It should be possible  to switch to DBD::MariaDB, but that would still require MySQL in FreeBSD, both MySQL client and server.

pnax:
The idea behind this PR was to reduce the amount of official database support to lift some work when it comes to database modification, and keep only MariaDB as official. But it seems this cannot be solved because of how FreeBSD deals with MySQL and MariaDB.
I'm closing this issue so that it does not stay around forever.

PR #850, opened 2021-08-16
Backport fix user api

blacksponge:
## Purpose

Secure the add_api_user method bit more and make it an optional feature.

## Context

Backport fix #839 for #838 to master

## Changes

* Use Plack::Middleware::ReverseProxy to set the environment according to the proxy headers
* Add two configuration keys API.enable_add_api_user (disabled by default) and API.enable_batch_jobs (enabled by default to minimize the change)

## How to test this PR

* Disabled the batch / user api.
* Running `./script/zmb add_api_user --username toto --api-key secret` should return the error `"Procedure 'add_api_user' not found"`.
* curl -s "Content-Type: application/json" --data '{"method":"add_batch_job","params":{"username":"toto","api_key":"toto", "domains":["example.com"]},"id":1,"jsonrpc":"2.0"}' 127.0.0.1:5000
* Enable the add_api_user method.
  * run the API behind a reverse proxy on the same host
    ```
    curl -s "Content-Type: application/json" --data '{"method":"add_api_user","params":{"username":"toto","api_key":"toto"},"id":1,"jsonrpc":"2.0"}' not.localhost/api  | jq
    ```
  * Try to add a user from a another host
  * If the query return `0` as result it works.


pnax (commented):
I know that the batch feature is not widely used, however aren't the new API keys introducing breaking changes? Are we okay with this to land into master?

To make the history cleaner, could you squash commit abfa742 and 0dbdbd4 together?

docs/Configuration.md
```diff
@@ -10,6 +10,22 @@ Each section in `backend_config.ini` is documented below.
...
 
+## API section
+
+Available keys: `enable_batch_jobs`, `enable_add_api_user`.
+
+### enable_batch_jobs
+
+Boolean value to enable the batch jobs methods of the API.
+
+Accpected values: 0 or 1, default to 0 (disabled).
```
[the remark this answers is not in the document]
blacksponge: fixed.

docs/Configuration.md
```diff
@@ -10,6 +10,22 @@ Each section in `backend_config.ini` is documented below.
...
+
+### enable_add_api_user
+
+Boolean value to enable the `add_api_user` method of the API.
+
+Accpected values: 0 or 1, default to 0 (disabled).
```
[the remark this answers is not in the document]
blacksponge: fixed.

docs/upgrade_db_zonemaster_backend_ver_6.2.1.md
```diff
@@ -0,0 +1,53 @@
+If your zonemaster database was created by a Zonemaster-Backend version smaller
+than v7.0.0, and not upgraded, use the instructions in this file.
```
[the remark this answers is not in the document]
blacksponge: The version number has been updated.

docs/upgrade_db_zonemaster_backend_ver_6.2.1.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Database upgrade
```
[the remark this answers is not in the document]
blacksponge: Oops, I removed it.

blacksponge:
> To make the history cleaner, could you squash commit abfa742 and 0dbdbd4 together?

Done.

blacksponge:
I have change the defaults to only disable the user api method by default so that existing installs using the batch method do not break. Changing the default for the batch api methods can be done in next release.

matsduf:
I do not see the need of `docs/upgrade_db_zonemaster_backend_ver_6.2.1.md`. The changes to the `Changes` should not be in this PR. That will be a separate PR for the release.

matsduf (changes requested):
I do not see the need of docs/upgrade_db_zonemaster_backend_ver_6.2.1.md. The changes to the Changes should not be in this PR. That will be a separate PR for the release.

blacksponge:
> I do not see the need of `docs/upgrade_db_zonemaster_backend_ver_6.2.1.md`. The changes to the `Changes` and ` lib/Zonemaster/Backend.pm` should not be in this PR. That will be a separate PR for the release.

I think it is easier for a user to upgrade if they have the specific upgrade instructions instead of just having the installation process (that they might skip thinking it has not changed). I can rename that file if needed.

blacksponge:
Tests are failing as they are trying to fetch the `fix-release-v2021.1.3` branch on https://github.com/zonemaster/zonemaster-ldns.
They seem to pass when run locally.

matsduf:
> > I do not see the need of `docs/upgrade_db_zonemaster_backend_ver_6.2.1.md`. The changes to the `Changes` and ` lib/Zonemaster/Backend.pm` should not be in this PR. That will be a separate PR for the release.
> 
> I think it is easier for a user to upgrade if they have the specific upgrade instructions instead of just having the installation process (that they might skip thinking it has not changed). I can rename that file if needed.

Those upgrade files are only used when the database files have to be updated (changed). We have never included information about updated dependencies. We do not need to include that information there and it does not add value.

matsduf:
The updates to the `Changes` file and `Backend.pm` file should not be in this PR.

matsduf:
> Tests are failing as they are trying to fetch the `fix-release-v2021.1.3` branch on https://github.com/zonemaster/zonemaster-ldns.

Now it it there, equal to master.

share/backend_config.ini
```diff
@@ -31,6 +31,13 @@ database_file = /var/lib/zonemaster/db.sqlite
 #
 #maximal_number_of_retries=3
 
+[RPCAPI]
+
+# Uncomment to enable the user API methods.
```
matsduf, line 36:
```
# Uncomment to enable API method "add_api_user"

File: share/backend_config.ini [hunk 6]
```diff
@@ -31,6 +31,13 @@ database_file = /var/lib/zonemaster/db.sqlite
...
+# enable_add_api_user = yes
+# Uncomment to disable the batch API methods.
```
matsduf, line 38:
```
# Uncomment to disable API method "add_batch_job"

File: docs/Configuration.md [hunk 7]
```diff
@@ -10,6 +10,25 @@ Each section in `backend_config.ini` is documented below.
 Restart the `zm-rpcapi` and `zm-testagent` daemons to load the changes
 made to the `backend_config.ini` file.
 
+## RPCAPI section
+
+Available keys: `enable_batch_jobs`, `enable_add_api_user`.
```
matsduf, line 15: To be consistent, shouldn't the first key be "enable_add_batch_jobs"? It will not affect running batch jobs.

matsduf (changes requested):

docs/Configuration.md
blacksponge: Actually this enable both `add_batch_job` and `get_batch_job_result` as I don't see a use case for the second one if the first is disabled.

blacksponge:
> We do not need to include that information there and it does not add value.

I think it does add value, it should be easy for a user to upgrade their Zonemaster installation without having to redo all the step of the installation instructions. Just having what has changed from a version to another and being able to just perform those few steps is much easier to follow and debug. I think that the `upgrade_db_zonemaster_backend_ver_6.2.1.md` should be much more generic, like `upgrade_zonemaster_backend_ver_6.2.1.md` and contain any additional step needed to go from a version to another, that include database migration but can also include new dependencies and such.

pnax (approved):

matsduf:
> I think it does add value, it should be easy for a user to upgrade their Zonemaster installation without having to redo all the step of the installation instructions. Just having what has changed from a version to another and being able to just perform those few steps is much easier to follow and debug. I think that the filees `upgrade_db_zonemaster_backend_ver_*.md` should be much more generic, like `upgrade_zonemaster_backend_ver_*.md` and contain any additional step needed to go from a version to another, that include database migration but can also include new dependencies and such.

I do not say that it is wrong to have upgrade information, but

1. Such a change requires more updates than this PR contains.
2. Such a change will increase the amount of updates for releases.
3. Such a change should be not be part of a patch.
4. If we introduce this for Backend then we should also be consistent to have that in other components.

Please remove it from this PR and bring it up on a meeting.

docs/Configuration.md
matsduf: There is no reason to disable method `get_batch_job_result`. You can want to see that even if you have disabled `add_batch_job`.

matsduf (approved):

docs/Configuration.md
```diff
@@ -10,6 +10,25 @@ Each section in `backend_config.ini` is documented below.
...
 
+## RPCAPI section
+
+Available keys: `enable_add_batch_job`, `enable_add_api_user`.
+
+### enable_add_batch_job
+
+Boolean value to enable the `enable_add_batch_job` methods of the API.
+
+Accpected values: `yes` (or `true`) or `no` (or `false`),
```
pnax, line 21: small typo "Accpected" -> "Accepted"

docs/Configuration.md
```diff
@@ -10,6 +10,25 @@ Each section in `backend_config.ini` is documented below.
...
+default to `yes` (enabled).
+
+### enable_add_api_user
+
+Boolean value to enable the `add_api_user` method of the API.
+
+Accpected values: `yes` (or `true`) or `no` (or `false`),
```
pnax, line 28: same typo

pnax (approved):

docs/Configuration.md
blacksponge: fixed

docs/Configuration.md
blacksponge: fixed

pnax (approved):

matsduf (commented):
1. I think `Plack::Middleware::ReverseProxy` should be added to Makefile.PL since it is a mandatory dependency.
2. I expected a `use Plack::Middleware::ReverseProxy` in `zonemaster_backend_rpcapi.psgi`. Is that because it is loaded indirectly by the other `Plack` modules.

blacksponge:
> I think Plack::Middleware::ReverseProxy should be added to Makefile.PL since it is a mandatory dependency.

Done

> Is that because it is loaded indirectly by the other Plack modules.

It is loaded by the plack builder, yes

matsduf:
Please merge.

PR #851, opened 2021-08-17
Load config only once and always log STDERR

### LINKED ISSUE #813: Load config in zm-rpcapi master process
Today the zm-rpcapi master process doesn't parse the config file. Instead each worker process parses the config the first thing they do. If the config is invalid they fail, and the master process starts spawning new replacements for the crashed workers.

The config should be parsed once and for all in the master process, and then the Config instance should be inherited by the worker processes.

mattias-p:
## Purpose

This PR does two things.
1. It makes zm-rpcapi not reload its config file for each worker on FreeBSD.
2. It makes zm-rpcapi write STDERR output to the log file on Linux.

## Context

Fixes #813.

## Changes

The options to starman in zm-rpcapi.lsb and zm_rpcapi.bsd are updated to be the same. The order of the options is also updated to make it easier to compare the two files.

## How to test this PR

This used to work only on Linux but not on FreeBSD:
1. Restart zm-rpcapi to make sure you have a fresh daemon started.
2. In one terminal run `tail -fn0 zm-rpcapi.log`.
3. In another terminal run `zmb version_info`.
4. Make sure the messages "NOTICE - Loading config:" is only logged once.

This used to work only on FreeBSD but not on Linux:
1. Update lib/Zonemaster/Backend/RPCAPI.pm to include the line `say STDERR "HERE";` in the `version_info` procedure.
2. Restart zm-rpcapi to make sure the modified code is loaded.
3. In one terminal run `tail -fn0 zm-rpcapi.log`.
4. In another terminal run `zmb version_info`.
5. Make sure the messages "HERE" is shows up in the output from the log file.


share/zm-rpcapi.lsb
```diff
@@ -27,7 +27,7 @@ STARMAN=`PATH="$PATH:/usr/local/bin" /usr/bin/which starman`
 . /lib/lsb/init-functions
 
 start () {
-    $STARMAN --user=$USER --group=$GROUP --error-log=$LOGFILE --pid=$PIDFILE --listen=$LISTENIP:$LISTENPORT --preload-app --daemonize $BINDIR/zonemaster_backend_rpcapi.psgi || exit 1
+    $STARMAN --daemonize --preload-app --user=$USER --group=$GROUP --pid=$PIDFILE --error-log=$LOGFILE --listen=$LISTENIP:$LISTENPORT /usr/local/bin/zonemaster_backend_rpcapi.psgi || exit 1
```
pnax, line 30: This removes the use of `$BINDIR`, is this on purpose? If so could you remove the variable definition?

mattias-p: No, that was a mistake. Thanks!

share/zm-rpcapi.lsb
```diff
@@ -27,7 +27,7 @@ STARMAN=`PATH="$PATH:/usr/local/bin" /usr/bin/which starman`
 . /lib/lsb/init-functions
 
 start () {
-    $STARMAN --user=$USER --group=$GROUP --error-log=$LOGFILE --pid=$PIDFILE --listen=$LISTENIP:$LISTENPORT --preload-app --daemonize $BINDIR/zonemaster_backend_rpcapi.psgi || exit 1
+    $STARMAN --daemonize --preload-app --user=$USER --group=$GROUP --pid=$PIDFILE --error-log=$LOGFILE --listen=$LISTENIP:$LISTENPORT $BINDIR/usr/local/bin/zonemaster_backend_rpcapi.psgi || exit 1
```
pnax, line 30: I think this should be
`$BINDIR/zonemaster_backend_rpcapi.psgi`
and not
`$BINDIR/usr/local/bin/zonemaster_backend_rpcapi.psgi`

pnax (commented):
To be honest, I don't really understand what is the fix on Linux, it seems to me that you only have reordered the options. Sorry if I missed something obvious.

mattias-p:
Your absolutely right. For some reason I thought zm-rpcapi.lsb was lacking the --error-log option and that I needed to add it back. I must have had a dirty working directory without realizing it. Hopefully this time I got things right.

Please review again.

pnax (approved):

blacksponge (approved):

matsduf:
> ## Purpose
> 
> This PR makes zm-rpcapi not reload its config file for each worker on FreeBSD.

@mattias-p, something is updated in the non-Freebsd file  share/zm-rpcapi.lsb. Can you document that?

mattias-p:
I expanded the Changes section with the rationale for the changes in zm-rpcapi.lsb.

pnax:
## Testing v2021.2

LGTM.
Just a remark, I was expecting to find the message in the logfile which default to `/var/log/zonemaster/zm-rpcapi` but it never lands in this file. I've found the message in the journal using `journalctl -u zm-rpcapi`. I don't know if this is the wanted behavior.

pnax:
Looking at the other daemon zm-testagent, it seems that some of the logs are duplicated between syslog and logfiles.

PR #853, opened 2021-08-18
add custom error message in json schema

blacksponge:
## Purpose

Provide more human friendly error messages than the ones generated by the JSON validator.

## Context

* Addresses issue #789 
* Replaces  #821 

## Changes

> TODO

I still need to test a bit more but this PR uses a custom attribute in the schema (`x-message-error`) to defined custom messages error directly in the schema. The exception being that when a required field is missing this custom message is ignored.

To make that possible I had to walk away from the Joi DSL, even the `description` field is not available with Joi. Not all schema have been replaced with raw JSON schema, only the necessary parts. I can replace the remaining bits later if needed.

## How to test this PR

`curl -sH "Content-Type: application/json" --data '{"method":"start_domain_test","params":{"domain":"toto.com", "nameservers":[{"ip":"a"}], "ds_info":[{"algorithm":-1,"digest":"abc","digtype":-1,"keytag":"abc"}]},"id":1,"jsonrpc":"2.0"}' localhost:5000/api  | jq`
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": "-32602",
    "data": [
      {
        "message": "Algorithm must be a positive integer",
      },
      {
        "path": "/ds_info/0/digest",
        "message": "Invalid digest format"
      },
      {
        "path": "/ds_info/0/digtype",
        "message": "Digest type must be a positive integer"
      },
      {
        "path": "/ds_info/0/keytag",
        "message": "Keytag must be a positive integer"
      },
      {
        "path": "/nameservers/0/ns",
        "message": "Missing property"
      },
      {
        "path": "/nameservers/0/ip",
        "message": "Invalid IP address"
      },
      {
        "message": "Domain name required",
        "path": "/nameservers/0/ns"
      },
      {
        "message": "Invalid IP address",
        "path": "/nameservers/0/ip"
      }
    ],
  }
}
```

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -13,6 +13,9 @@ use JSON::PP;
 use JSON::Validator::Joi;
 use Log::Any qw($log);
 use String::ShellQuote;
+use Mojo::JSON::Pointer;
+use Scalar::Util qw(blessed);
+use JSON::Validator::Schema::Draft7;
 
 # Zonemaster Modules
 use Zonemaster::LDNS;
@@ -131,9 +134,14 @@ sub get_language_tags {
     return \@lang_tags;
 }
 
-$json_schemas{get_host_by_name} = joi->object->strict->props(
-    hostname   => $zm_validator->domain_name->required
-);
+$json_schemas{get_host_by_name} = {
+    type => 'object',
+    additionalProperties => 0,
+    required => [ 'hostname' ],
+    properties => {
+        hostname => $zm_validator->domain_name
+    }
+};
 sub get_host_by_name {
     my ( $self, $params ) = @_;
     my @adresses;
@@ -162,9 +170,14 @@ $extra_validators{get_data_from_parent_zone} = sub {
     }
     return @errors;
 };
-$json_schemas{get_data_from_parent_zone} = joi->object->strict->props(
-    domain   => $zm_validator->domain_name->required
-);
+$json_schemas{get_data_from_parent_zone} = {
+    type => 'object',
+    additionalProperties => 0,
+    required => [ 'domain' ],
+    properties => {
+        domain => $zm_validator->domain_name
+    }
+};
 sub get_data_from_parent_zone {
     my ( $self, $params ) = @_;
 
@@ -297,24 +310,31 @@ $extra_validators{start_domain_test} = sub {
     }
 };
 
-$json_schemas{start_domain_test} = joi->object->strict->props(
-    domain => $zm_validator->domain_name->required,
-    ipv4 => joi->boolean,
-    ipv6 => joi->boolean,
-    nameservers => joi->array->items(
-        # NOTE: Array items are not compiled automatically, so to enforce all properties we need to do it by hand
-        $zm_validator->nameserver->compile
-    ),
-    ds_info => joi->array->items(
-        $zm_validator->ds_info->compile
-    ),
-    profile => $zm_validator->profile_name,
-    client_id => $zm_validator->client_id,
-    client_version => $zm_validator->client_version,
-    config => joi->string,
-    priority => $zm_validator->priority,
-    queue => $zm_validator->queue,
-);
+$json_schemas{start_domain_test} = {
+    type => 'object',
+    additionalProperties => 0,
+    required => [ 'domain' ],
+    properties => {
+        domain => $zm_validator->domain_name,
+        ipv4 => joi->boolean->compile,
+        ipv6 => joi->boolean->compile,
+        nameservers => {
+            type => 'array',
+            items => $zm_validator->nameserver
+        },
+        ds_info => {
+            type => 'array',
+            items => $zm_validator->ds_info
+        },
+        profile => $zm_validator->profile_name->compile,
+        client_id => $zm_validator->client_id->compile,
+        client_version => $zm_validator->client_version->compile,
+        config => joi->string->compile,
+        priority => $zm_validator->priority->compile,
+        queue => $zm_validator->queue->compile,
+
+    }
+};
 sub start_domain_test {
     my ( $self, $params ) = @_;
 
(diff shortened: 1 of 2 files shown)


blacksponge:
@mattias-p any thoughts on this implementation?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -652,9 +693,46 @@ sub validate_params {
...
+
+    # Customize error message from json validation
+    foreach my $err ( @json_validation_error ) {
+        print Data::Dumper::Dumper($err);
+
+        my $message = $err->message;
+        my @details = @{$err->details};
+
+        # Handle 'required' errors globally so it does not get overwritten
+        if (@details[1] eq 'required') {
```
matsduf, line 709: "Scalar value @details[1] better written as $details[1] at /usr/home/matsd/.cpanm/work/1630496242.3307/Zonemaster-Backend-6.2.0/blib/lib/Zonemaster/Backend/RPCAPI.pm line 709."

matsduf (changes requested):
```
t/test_validate_syntax.t .. 1/2 
#   Failed test 'no warnings'
#   at /usr/local/lib/perl5/5.32/Test/Builder.pm line 153.
# There were 1 warning(s)
# 	Previous test 0 ''
# 	Scalar value @details[1] better written as $details[1] at /usr/home/matsd/.cpanm/work/1630496463.3445/Zonemaster-Backend-6.2.0/blib/lib/Zonemaster/Backend/RPCAPI.pm line 709.
#  at /usr/home/matsd/.cpanm/work/1630496463.3445/Zonemaster-Backend-6.2.0/blib/lib/Zonemaster/Backend/RPCAPI.pm line 709.
# 	require Zonemaster/Backend/RPCAPI.pm called at t/test_validate_syntax.t line 14
# 	main::BEGIN() called at /usr/home/matsd/.cpanm/work/1630496463.3445/Zonemaster-Backend-6.2.0/blib/lib/Zonemaster/Backend/RPCAPI.pm line 743
# 	eval {...} called at /usr/home/matsd/.cpanm/work/1630496463.3445/Zonemaster-Backend-6.2.0/blib/lib/Zonemaster/Backend/RPCAPI.pm line 743
# 
# Looks like you failed 1 test of 2.
t/test_validate_syntax.t .. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/2 subtests 
t/validator.t .............         # No tests run!

    #   Failed test 'No tests run for subtest "ds_info"'
    #   at t/validator.t line 32.
    # Looks like you failed 1 test of 2.
t/validator.t ............. 1/2 
#   Failed test 'Everything but NoWarnings'
#   at t/validator.t line 206.
Can't call method "validate" on unblessed reference at t/validator.t line 30.
# Looks like your test exited with 255 just after 2.
t/validator.t ............. Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/2 subtests 
```

lib/Zonemaster/Backend/RPCAPI.pm
blacksponge: Fixed

blacksponge:
The tests should be fixed

matsduf (approved):
This looks good. Will it, in another PR, be possible to provide a translation mechanism?

blacksponge:
> This looks good. Will it, in another PR, be possible to provide a translation mechanism?

Yes, this has been the goal since the beginning, initially with #821 then with #823.

blacksponge:
@mattias-p as you proposed this idea I would like to have you feedback before merging.

mattias-p (commented):
I think this looks really good! It's more verbose but I'd say having the error messages together with the rules is worth the prices.

The `validate_params` procedure has grown a lot in complexity. Would you consider making some unit tests for it?

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -652,9 +693,46 @@ sub validate_params {
...
 
-    my @json_validation_error = $method_schema->validate( $params );
+    if (blessed $method_schema) {
+        $method_schema = $method_schema->compile;
+    }
+    my @json_validation_error = JSON::Validator::Schema::Draft7->new->coerce('booleans,numbers,strings')->data($method_schema)->validate( $params );
+
+    # Customize error message from json validation
+    foreach my $err ( @json_validation_error ) {
+        print Data::Dumper::Dumper($err);
```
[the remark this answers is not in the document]
blacksponge: oops, yes it should

blacksponge:
> The validate_params procedure has grown a lot in complexity. Would you consider making some unit tests for it?

Good idea, I'll make some tests

matsduf (commented):
It looks like the missing name server name is reported twice, first as missing and then as required. The invalid IP address is also reported twice.

```
      {
        "path": "/nameservers/0/ns",
        "message": "Missing property"
      },
      {
        "message": "Invalid IP address",
        "path": "/nameservers/0/ip"
      },
      {
        "message": "Domain name required",
        "path": "/nameservers/0/ns"
      },
      {
        "message": "Invalid IP address",
        "path": "/nameservers/0/ip"
      }
```

mattias-p:
The tests you put in look great! I believe the test file itself could be simpler if the `validate_params` method was refactored a bit. Have you considered making `validate_params` take the method schema and extra validator as arguments (instead of looking them up from global variables), and moving it to the Validator module?

blacksponge:
> It looks like the missing name server name is reported twice, first as missing and then as required. The invalid IP address is also reported twice.

I am aware of the issue, it is due to the double validation (json validator and extra validators), I am planning to remove the extra validator by using custom formats (see https://metacpan.org/pod/JSON::Validator#formats) but I would like to do that as a later step.

> Have you considered making validate_params take the method schema and extra validator as arguments (instead of looking them up from global variables), and moving it to the Validator module?

For now I would like to keep those with the RPCAPI so it is easier to associate the validation to the method. But I agree that having `validate_params` taking them as argument is better. As I said I am planning on removing the extra validators, then the replacement validation methods could totally go in the Validator module (for the schemas I am not so sure).

matsduf (approved):
I looks good.

mattias-p:
I tested this for v2021.2 and it looks good to me.

I found a very minor bug and reported it here: #919. We can fix that later.

PR #856, opened 2021-09-07
Use result entries table

blacksponge:
## Purpose

Move the `test_results.results` json array into a dedicated table.

## Context

Was briefly mentioned at last group meeting (2021-09-01)

## Changes

Adds a new table `result_entries`:

```
                                     Table "public.result_entries"
  Column   |          Type          | Collation | Nullable |                  Default                   
-----------+------------------------+-----------+----------+--------------------------------------------
 id        | integer                |           | not null | nextval('result_entries_id_seq'::regclass)
 hash_id   | character varying(16)  |           | not null | 
 level     | log_level              |           | not null | 
 module    | character varying(255) |           | not null | 
 testcase  | character varying(255) |           | not null | 
 tag       | character varying(255) |           | not null | 
 timestamp | real                   |           | not null | 
 args      | json                   |           | not null | 
Indexes:
    "result_entries_pkey" PRIMARY KEY, btree (id)
    "result_entries__hash_id" btree (hash_id)
    "result_entries__level" btree (level)



```

* The `get_test_result` and `get_test_history` methods are modified to use the new table
* `DB::test_result` is now only a getter, all write operations use the new database methods `add_result_entry` and `add_result_entries`

## How to test this PR

* Create a few tests
* Get the history / results
It should work the same way as it was

### DIFF
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -94,21 +94,19 @@ sub run {
         sub {
             my ( $entry ) = @_;
 
+            # TODO: Make minimum level configurable
             if ( $entry->numeric_level >= $numeric{INFO} ) {
-                $log->info($entry->string);
-                eval {
-                    $self->{_db}->add_result_entry( $test_id, {
-                        timestamp => $entry->timestamp,
-                        module    => $entry->module,
-                        testcase  => $entry->testcase,
-                        tag       => $entry->tag,
-                        level     => $entry->numeric_level,
-                        args      => $entry->args // {},
-                    });
-                };
-                if ($@) {
-                    $log->error($@);
-                }
+                $log->debug("Adding result entry in database: " . $entry->string);
+
+                $self->{_db}->add_result_entry( $test_id, {
+                    timestamp => $entry->timestamp,
+                    module    => $entry->module,
+                    testcase  => $entry->testcase,
+                    tag       => $entry->tag,
+                    level     => $entry->level,
+                    args      => $entry->args // {},
+                });
+
             }
 
             foreach my $trace ( reverse @{ $entry->trace } ) {
@@ -179,6 +177,7 @@ sub run {
     }
 
     # Actually run tests!
+    die Zonemaster::Backend::Error::Internal->new( reason => "called with results");
     eval { Zonemaster::Engine->test_zone( $domain ); };
     if ( $@ ) {
         my $err = $@;
@@ -190,9 +189,7 @@ sub run {
         }
     }
 
-    $self->{_db}->test_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
-
-    $progress = $self->{_db}->test_progress( $test_id );
+    $progress = $self->{_db}->test_progress( $test_id, 100 );
 
     return;
 } ## end sub run
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -177,7 +177,6 @@ sub run {
     }
 
     # Actually run tests!
-    die Zonemaster::Backend::Error::Internal->new( reason => "called with results");
     eval { Zonemaster::Engine->test_zone( $domain ); };
     if ( $@ ) {
         my $err = $@;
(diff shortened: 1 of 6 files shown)


matsduf (commented):
I think this looks like a good idea. I have some questions though:

1. How will this affect the performance of the Test Agent when writing the results?
2. How will this affect the performance of the RPCAPI reading the results and creating JSON for the response?
3. Does this have any affect on storage?
4. The `args` column in the new table is still a blob. Why not a separate table for that where each argument is one record?
5. This is a breaking change unless there is a migration script. Is that planned?

I will also do some tests.

blacksponge:
> How will this affect the performance of the Test Agent when writing the results?

I need make some test but I currently have some issue when running batches  on develop.

> How will this affect the performance of the RPCAPI reading the results and creating JSON for the response?

I should not affect the performance, the RPCAPI was already decoded JSON blobs, but I will test too.

> Does this have any affect on storage?

Again, I haven't done extensive testing yet.

> The args column in the new table is still a blob. Why not a separate table for that where each argument is one record?

I thought of totally splitting the args in a separate table `(arg_id, entry_id, arg_name, arg_value)` but I am not sure of the gain given the extra complexity it will add to both writing and reading the results. I can dig a bit more into that to see what can be done or not.

> This is a breaking change unless there is a migration script. Is that planned?

Yes, I wanted to have some early feedback before.

blacksponge:
> How will this affect the performance of the Test Agent when writing the results?

I changed the implementation to do only one query to the database instead of one for each log, this way the performance is mostly unaffected (Actually a fair share of the time taken to insert the results in the database is taken by the grep to filter the log entries to a given minimum log level. This overhead can be reduced by avoiding Moose in the Logger::Entry packages. I have a working POC [here](https://gitlab.rd.nic.fr/zonemaster/zonemaster-engine/-/blob/nightly-ronde/lib/Zonemaster/Engine/Logger/Entry.pm) that I am planning to integrate to the engine.)

matsduf:
I will review again when the conflicts are resolved.

matsduf:
> ```
>  module    | character varying(255) |           | not null | 
>  testcase  | character varying(255) |           | not null | 
>  tag       | character varying(255) |           | not null | 
> ```

Module name, testcase ID and message tag will never reach the length of 255 chacters. We could easily specify that they should be a maximum of 32, 32 and 64 characters. How much would we gain?

If I understand "character varying" I can handle Unicode characters. Today these three codes have names in ASCII. Could we gain some by using a column type for string of 8-bit length characters?

blacksponge:
> Module name, testcase ID and message tag will never reach the length of 255 chacters. We could easily specify that they should be a maximum of 32, 32 and 64 characters. How much would we gain?

varchar stores the data as length + data (without padding), there won't be any performance gain from restricting the length of string except for the length difference in the strings. What could be done is store them as enum / table + foreign key, but that could be done later as I want to keep the number of changes brought by this PR to a minimum.

> If I understand "character varying" I can handle Unicode characters. Today these three codes have names in ASCII. Could we gain some by using a column type for string of 8-bit length characters?

The database is encoded in utf8 so there is no overhead on ascii characters, they are still stored on 8-bit.

matsduf:
I think this looks fine. We should improve by having database engine independent documentation of the database structure to ensure that we get consistency between the engines.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -170,9 +173,10 @@ sub run {
         }
     }
 
-    $self->{_db}->test_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
+    my @entries = grep { $_->numeric_level >= $numeric{INFO} } @{ Zonemaster::Engine->logger->entries };
```
mattias-p, line 176: Did you consider filtering the entries before we even buffer them, as opposed to before we store them? I'm thinking we could save some memory that way.

mattias-p (commented):
This looks really good to me. Though Travis is unhappy. I only have a couple of questions/suggestions.

matsduf:
@blacksponge, I think this is interesting. If you e.g. test all domain names under a TLD you might want to extract all domain names getting a specific message tag or all message tags with a certain level for each domain name. Today you have to open the JSON blob for each domain name.

What are your plans?

pnax:
Replaced by #1092. The logic is kept, this is mainly a rebase to fix the conflicts.

PR #865, opened 2021-09-15
Missing index with SQLite

pnax:
## Purpose

This PR creates indexes for the `test_results` table with SQLite.

## Context

This is part 1/11 from #841

## Changes

Create indexes in SQLite.

## How to test this PR

A newly created SQLite database should have indexes `PRAGMA index_info(test_results)`.
Upgrading a database will be done in a later PR.

### DIFF
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -78,6 +78,23 @@ sub create_db {
         '
     ) or die "SQLite Fatal error: " . $self->dbh->errstr() . "\n";
 
+    $self->dbh->do(
+        'CREATE INDEX test_results__hash_id ON test_results (hash_id)'
+    );
+    $self->dbh->do(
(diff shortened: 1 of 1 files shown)


pnax:
Well the idea was to split #841 but Travis complains at my first split. I'm giving up on this idea.

matsduf:
The columns in the index must match the columns in the table definition.

pnax:
indeed, I can start again. Thanks for spotting the error.

matsduf:
> indeed, I can start again. Thanks for spotting the error.

Well, Travis told me. :-)

pnax:
Oops I may have been to fast on this one. This should be fixed.

matsduf (approved):
This works fine. I have not, however, measured if there is any gain.

pnax:
> I have not, however, measured if there is any gain.

This is something I plan to do in a later PR (we've spotted some performance issue in the Backend).

PR #866, opened 2021-09-16
Add a IF NOT EXISTS clause for all table/index/trigger creations

pnax:
## Purpose

Allow running the `share/create_db.pl` script several times without raising any error or warning. This is a step toward automatically creating the database on startup.

## Context

This is part 2/11 from #841

## Changes

Always check that a table, index or trigger does not exists before creation. For some cases, the clause `IF NOT EXISTS` is sufficient, for others, we need workaround because it is not present in all supported versions of the database engine.

## How to test this PR

Calling several times the `share/create_db.pl` script should not raise any issue.

### DIFF
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -137,21 +137,37 @@ sub create_db {
         '
     );
 
-    $dbh->do(
-        'CREATE INDEX test_results__hash_id ON test_results (hash_id)'
-    );
-    $dbh->do(
-        'CREATE INDEX test_results__params_deterministic_hash ON test_results (params_deterministic_hash)'
-    );
-    $dbh->do(
-        'CREATE INDEX test_results__batch_id_progress ON test_results (batch_id, progress)'
-    );
-    $dbh->do(
-        'CREATE INDEX test_results__progress ON test_results (progress)'
-    );
-    $dbh->do(
-        'CREATE INDEX test_results__domain_undelegated ON test_results (domain, undelegated)'
-    );
+    # Manually create the index if it does not exist
+    # the clause IF NOT EXISTS is not available for MySQL (used with FreeBSD)
+
+    # retrieve all indexes by key name
+    my $indexes = $dbh->selectall_hashref( 'SHOW INDEXES FROM test_results', 'Key_name' );
+
+    if ( not exists($indexes->{test_results__hash_id}) ) {
+        $dbh->do(
+            'CREATE INDEX test_results__hash_id ON test_results (hash_id)'
+        );
+    }
+    if ( not exists($indexes->{test_results__params_deterministic_hash}) ) {
+        $dbh->do(
+            'CREATE INDEX test_results__params_deterministic_hash ON test_results (params_deterministic_hash)'
+        );
+    }
+    if ( not exists($indexes->{test_results__batch_id_progress}) ) {
+        $dbh->do(
+            'CREATE INDEX test_results__batch_id_progress ON test_results (batch_id, progress)'
+        );
+    }
+    if ( not exists($indexes->{test_results__progress}) ) {
+        $dbh->do(
+            'CREATE INDEX test_results__progress ON test_results (progress)'
+        );
+    }
+    if ( not exists($indexes->{test_results__domain_undelegated}) ) {
+        $dbh->do(
+            'CREATE INDEX test_results__domain_undelegated ON test_results (domain, undelegated)'
+        );
+    }
 
 
     ####################################################################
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -124,41 +124,65 @@ sub create_db {
         '
     );
```
matsduf, line 125: Shouldn't there be a die() unless the table could be created?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -104,7 +104,7 @@ sub create_db {
     # TEST RESULTS
     ####################################################################
     $dbh->do(
-        'CREATE TABLE test_results (
+        'CREATE TABLE IF NOT EXISTS test_results (
```
matsduf, line 107: Couldn't it be enough to check if the table exists, and if not create all tables, triggers and indexes? Could that simplify the code?

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -100,7 +100,7 @@ sub create_db {
     # TEST RESULTS
     ####################################################################
     $dbh->do(
-        'CREATE TABLE test_results (
+        'CREATE TABLE IF NOT EXISTS test_results (
```
matsduf, line 103: Couldn't it be enough to check if the table exists, and if not create all tables, triggers and indexes?

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -119,28 +119,43 @@ sub create_db {
         '
     );
```
matsduf, line 120: Shouldn't there be a die() unless the table could be created?

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -79,27 +79,27 @@ sub create_db {
     ) or die "SQLite Fatal error: " . $self->dbh->errstr() . "\n";
```
matsduf, line 79: Here is a die, but not for the other tables or indexes. Why not everywhere?

pnax: Indeed this is something I forgot in the process. And this die has been added long time ago (https://github.com/zonemaster/zonemaster-backend/commit/cad366f845a71ae30efa9257b68b5b0766b4cb61).

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: I'd rather avoid dying if I can. I don't see what could prevent the table from being created at this point in the code. Especially since we explicitly check if it does not exist.

lib/Zonemaster/Backend/DB/MySQL.pm
pnax: I'm not sure. I find it better and proper to do all the checks. Then we are are sure we won't end with an improper behavior. Unless there is a better way to do all these checks, I'd like to keep them here.
And to add another argument, the goal I'd like to reach is to make it possible to let the agent call the `create_db` method on startup. So that we have a proper database on first run (without letting the user do this job for us). This would even allow to include the migration script within the code and run them on startup (but this will be done in a future work).

lib/Zonemaster/Backend/DB/MySQL.pm
matsduf: I do not understand the problem of die(). It would catch bugs that lets the code continue even if the table cannot be created.

pnax: I added a die after all table creation in case something happens.

pnax:
@matsduf I addressed your remark on the die usage, could you review again?

matsduf (changes requested):
It looks OK, and it works with SQLite, but not with MySQL, see #872.

mattias-p:
Looks good to me in principle, but I haven't tested.

pnax:
> it works with SQLite, but not with MySQL, see #872.

The error is not added with this PR. It seems to be already present in develop.

matsduf:
> > it works with SQLite, but not with MySQL, see #872.
> 
> The error is not added with this PR. It seems to be already present in develop.

Should I ignore MySQL? It will not work with MySQL.

matsduf:
Should #841 be resolved and merged first?

pnax:
> Should #841 be resolved and merged first?

Since this PR is a subset of #841, I don't see the point in merging both. Either I apply the changes made here in #841, or I merge this one and rebase #841 on top of it.

> Should I ignore MySQL? It will not work with MySQL.

As discussed, the error lies in develop, so either we fix develop and delay all PRs, or approve the PRs even if develop is broken.

matsduf:
> As discussed, the error lies in develop, so either we fix develop and delay all PRs, or approve the PRs even if develop is broken.

Since I cannot test this PR I would say fix develop first. It seems like #841 works better than develop.

pnax:
No relevant anymore since #841 has been merged and the changes made here were applied there.

PR #877, opened 2021-09-28
run make distcheck in ci

blacksponge:
## Purpose

This PR...

## Context

(e.g. Fixes #9999, Follow-up to #9999, etc.)

## Changes

...

## How to test this PR

...

### DIFF
--- a/MANIFEST.SKIP
+++ b/MANIFEST.SKIP
@@ -8,6 +8,7 @@
 ^\.github/pull_request_template\.md$
 ^\.travis\.yml$
 ^docs/internal-documentation/
+^lib/auto/share/dist/
 
 #!start included /usr/share/perl/5.20/ExtUtils/MANIFEST.SKIP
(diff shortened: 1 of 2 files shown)


matsduf:
See https://github.com/zonemaster/zonemaster-engine/pull/903 for the same goal for another repository. It is probably good if we have the same solutions in all perl based repositories, when possible.

MANIFEST.SKIP
```diff
@@ -8,6 +8,7 @@
 ^\.github/pull_request_template\.md$
 ^\.travis\.yml$
 ^docs/internal-documentation/
+^lib/auto/share/dist/
```
mattias-p, line 11: Shouldn't we be including this?

matsduf: There is another solution in Engine. I suggest that we have the same solution in all repositories where CPAN packages are created. Unless there is a good reason against that.

PR #855, opened 2021-09-02
Update documentation to use SQLite by default

pnax:
## Purpose

Try to remove some overhead in the installation process by moving the parts on installing MariaDB/MySQL and PostgreSQL engines to the end of the document.

## Context

N/A

## Changes

* Remove the Ubuntu section, merge it with the Debian section

* Update the "doc/Installation.md" with a new "Appendices" section.

* Move all references to MariaDB/MySQL or PostgreSQL to the appendices.

* Move the cleaning instructions to an appendix.

* Update the `backend_config.ini` file to use SQLite database engine by default

## How to test this PR

N/A (documentation only).

Issue #566, opened 2019-11-08
More streamlined database initialization


mattias-p:
It should be possible to initialize the database without running a finicky Perl one-liner to figure out where the SQL file was installed. This is currently a rough edge with regard to administration and installation testing.

A conceivable solution to this is to put a new executable in the path that locates the SQL file and pumps it to the appropriate SQL client. A command line parameter could be used for selecting DB engine.

pnax:
> It should be possible to initialize the database without running a finicky Perl one-liner to figure out where the SQL file was installed. 

I think that this part of this issue has been addressed with #834.

> A conceivable solution to this is to put a new executable in the path that locates the SQL file and pumps it to the appropriate SQL client. A command line parameter could be used for selecting DB engine. Such a tool could also be hooked up to the clean-up SQL files.

This second part is asking for an administrative tool that can be used. This can be a good idea. However I think there are already a lot of things on the roadmap. Could we postpone this issue to a later release (or delay it until further notice) ?

matsduf:
I do not think that running the script that creates the database is much of a problem. It is easy to create the instruction and it easy to run. I think we should put our efforts on something else.

PR #880, opened 2021-09-29
Update installation document

matsduf:
## Purpose

This PR adds a unified structure of the sections and headings, and it makes it easier to install for MariaDB and PostgreSQL by adding initialization scripts.

## Context

* Unifies the headline structure
* Simplifies the instructions by using initialization scripts
* Corrects instructions for FreeBSD not to use sudo

## Changes

Updates:
* docs/Installation.md
* MANIFEST

Adds:

* share/initial-mysql.sql
* share/initial-postgres.sql

## How to test this PR

* Follow the instructions for all OSs and all database backend and make sure that testing works.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -49,6 +49,8 @@ share/cleanup-mysql.sql
 share/cleanup-postgres.sql
 share/create_db.pl
 share/freebsd-pwd.conf
+share/initial-mysql.sql
+share/initial-postgres.sql
 share/patch_db_README.txt
 share/patch_mysql_db_zonemaster_backend_ver_1.0.3.pl
 share/patch_mysql_db_zonemaster_backend_ver_5.0.0.pl
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -1,6 +1,6 @@
 # Installation
 
-**Table of contents**
+## Table of contents
 
 * [1. Overview](#1-overview)
 * [2. Prerequisites](#2-prerequisites)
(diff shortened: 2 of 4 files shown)


pnax:
I don't get it. Are you just reverting changes made in #855, #841?

To me this is hiding essential steps in the database configuration within an obscure file, making it hard to easily configure the Backend. To make my point clear, I am against putting the `CREATE DATABASE` and `CREATE USER` instructions within a file. This is not fine in production environment because it is hiding how to define your own database/user. This might be fine to have such scripts in development environment. But then I don't think we should have them in master since it will require to keep them up to date, and we have already enough redundant things to maintain in the Backend.

I am also adding a pointer to issue #566 which is asking to avoid initializing the database with "finicky Perl one-liner". To my eyes this PR is definitely not solving this issue.

matsduf:
> I don't get it. Are you just reverting changes made in #855, #841?

Not at all. None of the files changed in #841 are touched in this PR. It does not revert #855, but it takes a step further to make it easier to install Backend. I have for PR testing installed the databases several time and I can see that it can be done easier.

> To me this is hiding essential steps in the database configuration within an obscure file, making it hard to easily configure the Backend. To make my point clear, I am against putting the `CREATE DATABASE` and `CREATE USER` instructions within a file. This is not fine in production environment because it is hiding how to define your own database/user. This might be fine to have such scripts in development environment. But then I don't think we should have them in master since it will require to keep them up to date, and we have already enough redundant things to maintain in the Backend.

I disagree. Nothing is hidden. The file is there to be updated if you want to have another name of database, user or another password. We have had the same database name, username and password for at least 5 years, probably since the creation of Zonemster-Backend. Database name etc have to be the same in the installation instructions as in the configuration file. It is not harder to keep the sql files updated than the installation instruction file. By having the sql files there is higher chance that installation is done consistently.

> I am also adding a pointer to issue #566 which is asking to avoid initializing the database with "finicky Perl one-liner". To my eyes this PR is definitely not solving this issue.

#846 removes the need of running a script to create the database tables. Instead that is magically done by the daemon. If that PR is merged then the instruction in this PR needs minor adjustment.

pnax:
> None of the files changed in #841 are touched in this PR

My bad, it was #834. In that PR I removed the "share/initial-mysql.sql" and "share/initial-postgresql.sql" files. These files are coming back here. Another reason why I removed them is because we cannot use them for Travis. So we had to maintain 2 different ways of creating the database. Now if they come back, the situation is back again. I really want to have less duplication within Backend.
 
>  We have had the same database name, username and password for at least 5 years, probably since the creation of Zonemster-Backend. Database name etc have to be the same in the installation instructions as in the configuration file. [...] By having the sql files there is higher chance that installation is done consistently.

Even though we agree on the end result, we don't on the path to reach it. I would like to make it possible to easily let the user adapt the database and user values. And I really think that using an SQL script is not an easy solution. Since I feel we'll continue to disagree on this point, I suggest we discuss about it during our next meeting and see if we can reach a consensus.

matsduf:
Replaced by #882.

PR #881, opened 2021-09-30
Remove the retry feature

matsduf:
## Purpose

The retry feature that depends on the `maximal_number_of_retries` key is removed by this PR. In the normal case a removal of a feature should start with deprecation. In this case the feature is documented as "experimental". From `Configuration.md`:

> This option is experimental and all edge cases are not fully tested. Do not use it (keep the default value "0"), or use it with care.

From default `backend_config.ini`:

```
# WARNING: The following option is experimental and all edge cases are not fully tested.
# Do not use it (keep the default value "0"), or use it with care.
#
#maximal_number_of_retries=3
```

There are no clear use cases for the feature. Instead for setting retries to 1 (permit running twice) the `max_zonemaster_execution_time` could be doubled with similar result.

The purpose here is to remove complexity not needed.

## Changes

The following files have been updated:

* docs/Configuration.md
* docs/internal-documentation/maintenance/Garbage-Collection-Testing.md
* lib/Zonemaster/Backend/Config.pm
* lib/Zonemaster/Backend/DB.pm
* lib/Zonemaster/Backend/DB/MySQL.pm
* lib/Zonemaster/Backend/DB/PostgreSQL.pm
* lib/Zonemaster/Backend/DB/SQLite.pm
* script/zonemaster_backend_testagent
* share/backend_config.ini
* t/config.t

The databases has a column `nb_retries` that is not removed by this PR. It should be taken in a second step.

## How to test this PR

1. Set `maximal_number_of_retries = 1` in the configuration file, and the daemon should not start.
2. Run various test and nothing should break unless expected to brea,
3. Create timeout by following `Garbage-Collection-Testing.md` and it should work as expected

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -32,7 +32,6 @@
 * [PUBLIC PROFILES and PRIVATE PROFILES sections](#PUBLIC-PROFILES-and-PRIVATE-PROFILES-sections)
 * [ZONEMASTER section](#ZONEMASTER-section)
   * [max_zonemaster_execution_time](#max_zonemaster_execution_time)
-  * [maximal_number_of_retries](#maximal_number_of_retries)
   * [number_of_processes_for_frontend_testing](#number_of_processes_for_frontend_testing)
   * [number_of_processes_for_batch_testing](#number_of_processes_for_batch_testing)
   * [lock_on_queue](#lock_on_queue)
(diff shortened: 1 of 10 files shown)


blacksponge:
The database creation script should also be updated. Should we provide a database upgrade script that remove the column from table?

blacksponge (approved):
Works as expected, it does indeed remove a bit of complexity.

matsduf:
@mattias-p, can you review?

PR #889, opened 2021-10-12
Display an error when trying to add an user with a non-localhost IP

ghost:
## Purpose

When trying to call `add_api_user` from a non-localhost IP, no error message is sent to the client, nor logged. This emits an error message using the custom error handling from #847.

## Context

Emit an error when calling `add_api_user` from a remote IP.

## Changes

Update the `add_api_user` method and update the API documentation with an output example.

## How to test this PR

Calling the `add_api_user` from a remote IP should send back an error message. For instance `curl -s "Content-Type: application/json" --data '{"method":"add_api_user","params:{"username":"toto","api_key":"toto"},"id":1,"jsonrpc":"2.0"}' http://10.0.0.1:5000/  | jq` should return something like
```
{
  "id": 1,
  "error": {
    "message": "Unauthorized to call this method from a remote IP",
    "data": {
      "remote_ip": "10.0.0.2"
    },
    "code": -32603
  },
  "jsonrpc": "2.0"
}
```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1215,7 +1215,7 @@ Trying to add a user over non-localhost:
     "data": {
       "remote_ip": "10.0.0.1"
     },
-    "message": "Unauthorized to call this method from a remote IP"
+    "message": "Call to \"add_api_user\" method not permitted from a remote IP"
   }
 }
 ```
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -577,7 +577,7 @@ sub add_api_user {
         }
         else {
             die Zonemaster::Backend::Error::PermissionDenied->new(
-                message => 'Unauthorized to call this method from a remote IP',
+                message => 'Call to "add_api_user" method not permitted from a remote IP',
                 data => { remote_ip => $remote_ip }
             );
         }


lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -575,6 +575,12 @@ sub add_api_user {
         if ( $allow ) {
             $result = 1 if ( $self->{db}->add_api_user( $params->{username}, $params->{api_key} ) eq '1' );
         }
+        else {
+            die Zonemaster::Backend::Error::PermissionDenied->new(
+                message => 'Unauthorized to call this method from a remote IP',
```
matsduf, line 580: I think that "forbidden" or "not permitted" is better than "unauthorized" in this context. There is no authorization mechanism for this. Suggestions:
```
'Call to this method from a remote IP is not permitted'
```

docs/API.md
```diff
@@ -1208,9 +1208,15 @@ Omitting params:
...
-  "result": 0,
   "id": 1,
-  "jsonrpc": "2.0"
+  "jsonrpc": "2.0",
+  "error": {
+    "code": -32603,
+    "data": {
+      "remote_ip": "10.0.0.1"
+    },
+    "message": "Unauthorized to call this method from a remote IP"
```
matsduf, line 1218: See comment in RPCAPI.pm.

lib/Zonemaster/Backend/RPCAPI.pm
pnax: Changed to:
```
Call to "add_api_user" method not permitted from a remote IP
```

matsduf:
## Testing v2021.2
LGTM.

PR #890, opened 2021-10-12
[optimization] New index to speed up the query to retrieve next test

pnax:
## Purpose

Currently the following SQL query (`get_test_request` in DB.pm) can take several seconds on huge tables (> 1 million entries). This makes it hard to use the batch mode. Creating an index helps reducing the query time to milliseconds.

https://github.com/zonemaster/zonemaster-backend/blob/7d37f9469653e40295625827c11ee57f3f9080ce/lib/Zonemaster/Backend/DB.pm#L240-L242

### Limits

This improvement has been tested with PostgreSQL. It is assumed that it will work with SQLite (even though using huge table is not recommended with SQLite).
For MariaDB/MySQL, this new index won't work (at least not with MySQL < 8). This is something weird, the index can properly be created, but the descending order is just ignored (see [this issue][1] on MaraiDB's Jira). However MySQL fixes that and supports descending indexes (see [release notes][2]).
There is a workaround to this. It would require to use PERSISTENT column storing the opposite value of the `priority` so that this new column can be used as ascending index (see [this blog post][3]). Having tested this workaround, this properly works. But I am a little bit reluctant to add it to the code since it requires a specific table for MariaDB.

### Discussion

I think that this index is needed for PostgreSQL. For MySQL we can add it (but alert that this is only working with MySQL >= 8). I feel unease to let people using MariaDB in the dark. Do you have any suggestions on this?

### Workaround

For MariaDB, the following workaround works:
```
ALTER TABLE test_results ADD COLUMN rev_priority INTEGER AS (0 - priority) PERSISTENT;
CREATE INDEX test_results__progress_priority_id ON test_results (progress, rev_priority, id);
```

## Context

Speed up query to retrieve next test to process.

## Changes

Create a new index (PostgreSQL, SQLite, MariaBD/MySQL).
Disclaimer: won't work with MariaDB and MySQL < 8 (unless using the workaround).

## How to test this PR

Populate the table `test_results` with millions of domains. Run the following SQL query:
```
SELECT * FROM test_results WHERE progress = 0 ORDER BY priority DESC, id ASC LIMIT 1;
```

Then create the index (TODO: add upgrade instructions):
* PostgreSQL and SQLite:
   ```
   CREATE INDEX test_results__progress_priority_id_progress_0 ON public.test_results USING btree (progress, priority DESC, id) WHERE (progress = 0);
   ```
* MariaDB and MySQL (the index creation works, for MariaDB and MySQL < 8 the descending order is ignored):
   ```
   CREATE INDEX test_results__progress_priority_id_progress_0 ON public.test_results USING btree (progress, priority DESC, id);
   ```

Run the `SELECT` query again and see the drop in query time.

## TODO

* add upgrade instructions
* add `queue` to the index since the query might use it

[1]: https://jira.mariadb.org/browse/MDEV-13756
[2]: https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html
[3]: https://falseisnotnull.wordpress.com/2014/04/25/descending-indexes-in-mariadb/


pnax:
It has been decided to keep the index for PostgreSQL (remove it from SQLite and MariaDB/MySQL). Add a comment somewhere to say that batches is optimized for PostgreSQL. This should lead to the creation of a new document about further usage / intense usage of Zonemaster.

matsduf (approved):

blacksponge:
# Tests release 2021.2
I mark this as tested because it does not break anything but I did not managed to reproduce the performance issue on my machine.

PR #892, opened 2021-10-20
Drop support for PostgreSQL ≤ 9

ghost:
## Purpose

With https://github.com/zonemaster/zonemaster/pull/991, we are supporting PostgreSQL >= 10. It is then safe to remove workaround added in commit e0ccd60bf22618df5fc2cc0ba9db729c0352ed82 (#841) checking for the existence of an index before creating it.
This allows to let PostgreSQL deal with potential race conditions.

## Context

Use clause `IF NOT EXISTS` on index creation for PostgreSQL.

## Changes

Update SQL query when creating indexes.

## How to test this PR

1. create a new database (with PostgreSQL >= 10)
2. run the `share/create_db.pl` script on this database
3. check that the indexes are properly created
4. re-run the script `share/create_db.pl`, no error should be emitted (notice messages are fine *).

\* _Do not throw an error if a relation with the same name already exists. A notice is issued in this case._ [PostgreSQL documentation]

[PostgreSQL documentation]: https://www.postgresql.org/docs/10/sql-createindex.html

### DIFF
--- a/lib/Zonemaster/Backend/DB/PostgreSQL.pm
+++ b/lib/Zonemaster/Backend/DB/PostgreSQL.pm
@@ -81,43 +81,26 @@ sub create_db {
         '
     ) or die Zonemaster::Backend::Error::Internal->new( reason => "PostgreSQL error, could not create 'test_results' table", data => $dbh->errstr() );
 
-    # Manually create the index if it does not exist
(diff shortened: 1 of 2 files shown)


pnax:
Fixed a conflict due to the renaming of the patch script. Please re-review.

PR #899, opened 2021-11-16
Adds PO file for translation to Swedish

matsduf:
## Purpose

This PR provides translation into Swedish.

## How to test this PR

Inspect output from RPC API when error message has been putputted.


share/sv.po
```diff
@@ -0,0 +1,76 @@
...
+
+msgid "Domain name required"
+msgstr "Domännamn är obligatoriskt"
+
+msgid ""
+"The domain name is not a valid IDNA string and cannot be converted to an A-"
+"label"
+msgstr ""
+"Domännamnet är inte ett giltigt IDN-namn och kan inte konverteras till A-"
+"labelformat"
```
mattias-p, line 22: Is it okay to have a compound word that switches languages half way through? Wouldn't it be better with "A-label-format"?

share/sv.po
```diff
@@ -0,0 +1,76 @@
...
+
+msgid "The domain name contains non-ascii characters and IDNA is not installed"
+msgstr ""
+"Domännamnet innehåller icke-ASCII-tecken, men stöd för IDN är inte "
+"installerat"
+
+msgid "The domain name character(s) are not supported"
+msgstr "Domännamnets teckeninnehåll stöds inte"
```
mattias-p, line 30: How about this?

> Domännamnet innehåller tecken som inte stöds

share/sv.po
```diff
@@ -0,0 +1,76 @@
...
+
+msgid ""
+"Warning: Zonemaster::LDNS not compiled with libidn, cannot handle non-ASCII "
+"names correctly."
+msgstr ""
+"Varning: Zonemaster::LDNS har inte kompilerats med libidn vilket gör att "
+"icke-ASCII-strängar inte kan hanteras korrekt"
+
+msgid "The domain name contains a character or characters not supported"
+msgstr "Domännamnet innehåller en eller flera tecken som inte stöds"
```
mattias-p, line 61: `en eller flera` -> `ett eller fler`

share/sv.po
```diff
@@ -0,0 +1,76 @@
...
+
+msgid "Invalid digest format"
+msgstr "Ogiltigt format på DS-digest"
+
+msgid "Algorithm must be a positive integer"
+msgstr "Algoritm ska vara ett positivt heltal"
```
mattias-p, line 67: `ska` -> `måste`

There are more like this below.

share/sv.po
```diff
@@ -0,0 +1,76 @@
...
+
+msgid "Digest type must be a positive integer"
+msgstr "Digest-typ ska var ett positivt heltal"
```
mattias-p, line 70: `ska var` -> `måste vara`

share/sv.po
matsduf: Yes, but it depends on how well-known the expression is. In this case an extra hyphen can be motivated. I'll update.

share/sv.po
matsduf: That could work, but that is a freer translation.

share/sv.po
matsduf: Fixed.

share/sv.po
matsduf: It is also possible to use "måste" but that is not a superior alternative.

https://sv.wiktionary.org/wiki/skola#Verb

```
(modalt hjälpverb) vara tvungen att

    Man ska inte kasta skräp i naturen!
    Synonymer: måste (om satsen inte innehåller negation)
```

share/sv.po
matsduf: See above.

mattias-p (approved):

PR #900, opened 2021-11-16
Adds Spanish language (es_ES) to various places

matsduf:
## Purpose

This PR adds Spanish

## How to test this PR

Nothing should break. Engine must be updated before any translation will be seen in messages.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -240,6 +240,7 @@ Language | Preferred language tag | Deprecated language tag
 ---------|------------------------|------------------
 Danish   | da                     | da_DK
 English  | en                     | en_US
+Spanish  | es                     | es_ES
 Finnish  | fi                     | fi_FI
 French   | fr                     | fr_FR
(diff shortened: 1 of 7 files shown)


t/test_validate_syntax.t
```diff
@@ -23,7 +23,7 @@ engine = SQLite
 database_file = $tempdir/zonemaster.sqlite
 
 [LANGUAGE]
-locale = en_US fr_FR da_DK fi_FI nb_NO sv_SE
+locale = da_DK en_US es_ES fi_FI fr_FR nb_NO sv_SE
```
mattias-p, line 26: `fr_FR` and `nb_NO` should be enough here.

matsduf: If it is not necessary to have all locale here it is out of scope to update it. I will just remove it from the PR with a force push. It could be updated in another PR concerning the unit test.

mattias-p (approved):

PR #904, opened 2021-11-17
remove unsupported perl version from travis

blacksponge:
## Purpose

The oldest perl version that we support is 5.26, don't run tests on unsupported perl versions.

## Changes

Remove version 5.24, 5.22 and 5.16 from tests

## How to test this PR

...


matsduf (approved):
Good point. We should check the other repositories too.

mattias-p (approved):

pnax (approved):

mattias-p (approved):

matsduf (approved):
I approve.

PR #916, opened 2021-11-22
Maximize the reach of eval+handle_exception

### LINKED ISSUE #914: Business logic in RPCAPI not guarded by handle_exception
There are many lines of business logic in the RPCAPI procedure methods that aren't covered by an eval+handle_exception construct. At the time of writing I'm not aware of any inputs that would cause actual exceptions thrown from these lines, but the point of the eval+handle_exception construct is to safeguard against information leakage and to ensure nice log messages and nice error responses. And right now we're not getting the coverage we should be getting.

mattias-p:
## Purpose

This PR means to ensure that all exceptions thrown inside the RPCAPI handler methods get handled by handle_exception(). Previously there were some spots where exceptions would not be properly caught. Instead they would be returned to the RPCAPI client and not show up in the log file.

## Context

Fixes #914.

## Changes

All of the business logic in the RPCAPI handler methods is moved in under the covered of eval+handle_response. Before only most of it was.

Some of the moved lines could not possibly throw exceptions, but I moved them into the eval blocks anyway because it's just simpler to have everything inside them without making exceptions.

In one case I replaced a call to handle_exception() with a call to die(). This is because you must not have a handle_exception() inside the eval block of another handle_exception().

## How to test this PR

1. Insert a `die "Oopsie!";` statement into the `around 'locale'` clause in `Zonemaster::Engine::Translator`.
2. Restart zm-rpcapi
3. Run `zmtest a.b.c`
4. Verify that you get an "Internal server error" response without source code references.
5. Verify that the Oopsie gets written into the log file.

### DIFF
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -134,15 +134,20 @@ $json_schemas{get_language_tags} = joi->object->strict;
 sub get_language_tags {
     my ( $self ) = @_;
 
-    my %locales = $self->{config}->LANGUAGE_locale;
-
     my @lang_tags;
-    for my $lang ( sort keys %locales ) {
-        my @locale_tags = sort keys %{ $locales{$lang} };
-        if ( scalar @locale_tags == 1 ) {
-            push @lang_tags, $lang;
+    eval {
+        my %locales = $self->{config}->LANGUAGE_locale;
+
+        for my $lang ( sort keys %locales ) {
+            my @locale_tags = sort keys %{ $locales{$lang} };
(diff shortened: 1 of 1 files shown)


matsduf:
>     1. Insert a `die "Oopsie!";` statement into the `around locale` clause in `Zonemaster::Engine::Translator`.

Look for `around 'locale'`

matsduf (approved):
I do not see any issue with this. If it is to be included in v2021.2 it should be merged now.

matsduf:
@blacksponge and @pnax, please review this.

blacksponge (approved):
nice improvement, works as expected

PR #924, opened 2021-12-03
Preparation for release v2021.2 (Backend)

matsduf:
## Purpose

* Bumps version in perl module
* Updates Changes file
* Updates requirement in Makefile.PL to latest version of Engine

## Changes

Updated files:

* Changes
* lib/Zonemaster/CLI.pm
* Makefile.PL

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,43 @@
 Release history for Zonemaster component Zonemaster-Backend
 
+v8.0.0 2021-12-03 (public release version)
+
+ [Features]
+ - Adds Finnish translation of error messages (#921, #911)
+ - Replaces CentOS with Rocky Linux (#908, #895)
+ - Adds upgrade documentation (#907, #879, #618)
+ - Adds Swedish translation of error messages (#899)
+ - Adds translation of error messages (#891)
+ - Drops support for PostgreSQL version 9 and before (#892)
+ - Adds support for collecting metrics (#844)
+ - Optimization of batches when PostgreSQL is used (#890)
+ - Improved result history response in API (#837, #830)
+ - Deprecates country code (#796, #794)
+ - Makes database port configurable (#755, #496)
+
+ [Fixes]
+ - Document clean-up and improvement (#913, #907, #882, #873, #871,
+   #822, #814, #816, #797)
+ - Improves error catching (#916, #914)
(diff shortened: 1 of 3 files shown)


Changes
```diff
@@ -1,5 +1,43 @@
...
+ - Replaces CentOS with Rocky Linux (#908, #895)
+ - Adds upgrade documentation (#907, #879, #618)
+ - Adds Swedish translation of error messages (#899)
+ - Adds translation of error messages (#891)
+ - Drops support for PostgreSQL version 9 and before (#892)
+ - Adds support for collecting metrics (#844)
+ - Optimization of batches when PostgreSQL is used (#890)
+ - Improved result history response in API (#837, #830)
+ - Deprecates country code (#796, #794)
+ - Makes database port configurable (#755, #496)
```
mattias-p, line 16: Dropping support for something isn't a feature - it's a breaking change. E.g. PostgreSQL. Or do we have reason to believe that PostgreSQL will still work, but we just haven't ensured that in release testing? In that case we should probably find a different wording.

Maybe group these so that similar items appear next to each other? E.g. updates related to translations.

matsduf: At some time we decided to change the major version, which means that we think it is a breaking change, but I do not think that is because of the PostgreSQL. Do you remember?

matsduf: I have created a section for breaking changes, and done some grouping. The new code will not work with old PostgreSQL.

mattias-p: Yeah, #892 says we've removed a workaround for PostgreSQL < 10.

Changes
```diff
@@ -1,5 +1,45 @@
 Release history for Zonemaster component Zonemaster-Backend
 
+v8.0.0 2021-12-03 (public release version)
+
+ [Breaking changes]
+ - Drops support for PostgreSQL version 9 and before (#892)
+ - Deprecates country code in RPC API calls (#796, #794)
```
mattias-p, line 7: Deprecations aren't breaking changes. They're just an early announcements about upcoming breaking changes. The breaking change happens when the feature is actually removed. We should probably have a special section for deprecations.

matsduf: Updated.

mattias-p (approved):
Looking good!

PR #928, opened 2021-12-06
Fix inconsistency in SQLite test's age

pnax:
## Purpose

Fix inconsistency in SQLite test's age.

## Context

#841 did not update SQLite to use `creation_time` instead of `test_start_time` when looking for test to reuse, which is inconsistent with MySQL and PostgreSQL.
In fact #841 made the change (a8a4cee) and removed it in a later commit (5df15c0).

Addresses #840.

## Changes

Update method `recent_test_hash_id` in `DB/SQLite.pm`.

## How to test this PR

With SQLite configured, start two tests for the same domain. Before, this fix, the test ID is different. With this fix, the test ID should be the same (until `age_reuse_previous_test` seconds are passed).
```
curl -sH "Content-Type: application/json" --data '{"method":"start_domain_test","params":{"domain":"domain.example"},"id":1,"jsonrpc":"2.0"}' localhost:5000/api | jq
```

### DIFF
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -128,7 +128,7 @@ sub recent_test_hash_id {
 
     my $dbh = $self->dbh;
     my ( $recent_hash_id ) = $dbh->selectrow_array(
-        "SELECT hash_id FROM test_results WHERE fingerprint = ? AND test_start_time > DATETIME('now', ?)",
+        "SELECT hash_id FROM test_results WHERE fingerprint = ? AND creation_time > DATETIME('now', ?)",
         undef, $fingerprint, "-$age_reuse_previous_test seconds"
     );
 


blacksponge (approved):

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -128,7 +128,7 @@ sub recent_test_hash_id {
 
     my $dbh = $self->dbh;
     my ( $recent_hash_id ) = $dbh->selectrow_array(
-        "SELECT hash_id FROM test_results WHERE fingerprint = ? AND test_start_time > DATETIME('now', ?)",
+        "SELECT hash_id FROM test_results WHERE fingerprint = ? AND creation_time > DATETIME('now', ?)",
```
matsduf, line 131: It is consistent with the other database modules to use "creation_time", but it is not logical. It is not the time of creation that makes sense, but the time of test execution. Are you sure that you do not want to change the other databases to use "test_start_time" instead?

matsduf (changes requested):

matsduf:
Is this inconsistency bad enough to be included in the patch release?

lib/Zonemaster/Backend/DB/SQLite.pm
pnax: This is exactly the purpose of #840 to discuss which one should be used. However when working on #841, I chose to make SQLite consistent with MySQL and PostgreSQL so that the behavior is the same with all databases.

> Is this inconsistency bad enough to be included in the patch release?

Since this should have been part of v2021.2 release, I think this should be included in v2021.2.1.

matsduf: If we can be safe that there are no bad side effects then it could be included even though it probably has very low impact since very few batches are run.

pnax: Since we reuse the same logic as the one with PostgreSQL and MySQL which have been around since the beginning, see commit ab540299f23 and 8b57fe576, I don't think this would have any bad side effects.

matsduf (approved):
I approve this, but does it qualify to be included in the patch release?

I still question the logic of using creation time, but in practice it has low impact.

pnax:
> I approve this, but does it qualify to be included in the patch release?

This should have been included in the latest release and I realized some code refactoring made it disappear. Therefore I do think  this should be included in the patch release in order to be aligned with what we thought should have been shipped with v2021.2.

> I still question the logic of using creation time, but in practice it has low impact.

I understand. Again this is not about answering this question but rather make all engines behave in a similar fashion. I'll merge this into develop and I suggest we update the logic in a future work (based on the discussion in #840).

PR #929, opened 2021-12-07
improve code quality

blacksponge:
## Purpose

Fix issue accessing hash

## Context

See mailing list

### DIFF
--- a/lib/Zonemaster/Backend/Metrics.pm
+++ b/lib/Zonemaster/Backend/Metrics.pm
@@ -1,5 +1,8 @@
 package Zonemaster::Backend::Metrics;
 
+use strict;
+use warnings;
+
 use Log::Any qw($log);
 
 eval("use Net::Statsd");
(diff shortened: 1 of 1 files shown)


matsduf:
Can you describe the issue? I find no email about it. What is affected by the bug? What milestone?

matsduf (approved):

blacksponge:
It was on the zonemaster-users mailing list 

> Error while loading /opt/zonemaster/bin/zonemaster_backend_rpcapi.psgi:
syntax error at /opt/zonemaster/lib/perl5/Zonemaster/Backend/Metrics.pm
line 35, near "%CODE_STATUS_HASH{"
syntax error at /opt/zonemaster/lib/perl5/Zonemaster/Backend/Metrics.pm
line 36, near "} else"
Compilation failed in require at
/opt/zonemaster/bin/zonemaster_backend_rpcapi.psgi line 29.
BEGIN failed--compilation aborted at
/opt/zonemaster/bin/zonemaster_backend_rpcapi.psgi line 29.

It is not on an os / perl version we support but it should be corrected. It is mostly about the hash being accessed with `%CODE_STATUS_HASH{$code};` instead of `$CODE_STATUS_HASH{$code};` and a few missing semi colons.

blacksponge:
> What milestone?

Ideally this should go with the upcoming fix release.

mattias-p (approved):

matsduf:
> It is not on an os / perl version we support but it should be corrected. It is mostly about the hash being accessed with `%CODE_STATUS_HASH{$code};` instead of `$CODE_STATUS_HASH{$code};` and a few missing semi colons.

The missing semicolons are optional in all versions of Perl, as far as I know, but it is better to include them. If it works with `%` or not seems to be version dependent. I guess newer versions accepts them, which I find surprising.

matsduf:
> > What milestone?
> 
> Ideally this should go with the upcoming fix release.

It should be safe to include it, I think. But we must test.

pnax (approved):

PR #930, opened 2021-12-07
Add danish translation

### LINKED ISSUE #909: Translate error messages into Danish (Backend)
As a new feature of Zonemaster-Backend, in the upcoming release,  error messages are better returned in the GUI, such as incorrect IP address. Those messages can also be translated. After the release, GUI will show the error message in the same language as has been selected.

The error messages comes from Backend via RPC-API. The translation is done in Backend with the use of PO file. For Danish there is [share/da.po]. Now it is empty, but if the draft "[Instructions for translators]" is followed it will be filled with *msgids* and translations can be added (*msgstrs*).

There will be 19 *msgids* to be translated, most of them very short.

[Instructions for translators]: https://github.com/zonemaster/zonemaster/blob/e2a8c4eb74e45d7e4405ba478938a8ee374bd6ec/docs/internal-documentation/maintenance/Instructions-for-translators.md

[share/da.po]: https://github.com/zonemaster/zonemaster-backend/blob/develop/share/da.po

haarbo:
## Purpose

This PR...

## Context

Fixes #909 

## Changes

Add danish translation

## How to test this PR

...

### DIFF
--- a/share/da.po
+++ b/share/da.po
@@ -2,12 +2,71 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-10-14 11:25+0200\n"
+"POT-Creation-Date: 2021-12-07 15:34+0000\n"
 "PO-Revision-Date: 2021-10-14 11:25+0200\n"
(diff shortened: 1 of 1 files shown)


matsduf (approved):
It looks fine to me. I approve.

mattias-p (approved):

PR #932, opened 2021-12-08
Fix internationalized testing failure with error: "malformed UTF-8 character in JSON string"

pnax:
## Purpose

The different database drivers do not retrieve the data with the same Perl's internal representation. Therefore it is required to check this representation before calling "decode_json" that "expects an UTF-8 (binary) string" (https://metacpan.org/pod/JSON::PP#decode_json).

## Context

Fixes #931
Also see #570

## Changes

Fix the `get_test_params` method in `DB.pm`.

## How to test this PR

Check that internationalized and non-internationalized domains works properly for all database engines.

PR #934, opened 2021-12-08
Wrapper around "decode_json" to properly encode the string

pnax:
## Purpose

Create a wrapper around `decode_json` in order to properly encode the data in the expected format.

> expects an UTF-8 (binary) string
https://metacpan.org/pod/JSON::PP#decode_json

## Context

Refactor code added in #932 and #571.

## Changes

New wrapper used in different places.

## How to test this PR

This is mostly a refactoring. So nothing should break.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -316,7 +316,7 @@ sub force_end_test {
     my ( $self, $hash_id, $results, $timestamp ) = @_;
     my $result;
     if ( defined $results && $results =~ /^\[/ ) {
-        $result = decode_json( $results );
+        $result = _decode_json_sanitize( $results );
     }
     else {
         $result = [];
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -260,11 +260,7 @@ sub get_test_params {
 
     my $result;
     eval {
-        if ( utf8::is_utf8( $params_json ) ) {
-            $result = decode_json( encode_utf8( $params_json ) );
-        } else {
-            $result = decode_json( $params_json );
-        }
+        $result = _decode_json_sanitize( $params_json );
     };
 
     die Zonemaster::Backend::Error::JsonError->new( reason => "$@", data => { test_id => $test_id } )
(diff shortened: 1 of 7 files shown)


matsduf:
Why do you create two PRs, this and #932?

pnax:
> Why do you create two PRs, this and #932?

I started to provide a fix for the breaking part. Then I thought it would be nice to do some refactoring. And because I wasn't sure how we wanted to deal with this refactoring (release it in the next fix release, or in v2022.1), I chose to make 2 PRs.
If it's ok to include this PR into v2021.2.1 this it fine for me.

mattias-p:
Forgive me for not being quite up to speed on this issue. I haven't thought about it in a long time.

But it looks to me that we've been allowing some flexibility in how we're storing JSON blobs in PostgreSQL. And this PR looks like it extends that flexibility to the other databases. Is that correct?

If that is the case I think this is a step in the wrong direction. A better way would be to normalize the contents of the database.

pnax:
> But it looks to me that we've been allowing some flexibility in how we're storing JSON blobs in PostgreSQL. And this PR looks like it extends that flexibility to the other databases. Is that correct?

Not exactly. This does not affect how we store the data in the database. But how we retrieve the data. The drivers do not return the data in the same format. Hence the suggestion to add a wrapper that we use for all database engines in order to work with the same string format after that point. This extends the work perform in #932 to all databases.

mattias-p:
> > But it looks to me that we've been allowing some flexibility in how we're storing JSON blobs in PostgreSQL. And this PR looks like it extends that flexibility to the other databases. Is that correct?
> 
> Not exactly. This does not affect how we store the data in the database. But how we retrieve the data. The drivers do not return the data in the same format. Hence the suggestion to add a wrapper that we use for all database engines in order to work with the same string format after that point. This extends the work perform in #932 to all databases.

So here is my current understanding. Do you agree with it?
* To our best knowledge, we're always storing JSON blobs in the database in a consistent format.
* Sometimes the DBI driver returns JSON blobs in inconsistent formats.
* To our best knowledge, this inconsistency only affects the PostgreSQL adapter.

If the above statements are true, wouldn't it be better to only have the selective `encode_utf8` call in the PostgreSQL adapter?

pnax:
> Do you agree with it?

Yes.

Looking into the DBD::Pg documentation I came across [`pg_enable_utf8`](https://metacpan.org/pod/DBD::Pg#pg_enable_utf8-(integer)): "_If this attribute is set to 0, then DBD::Pg will never UTF8-decode returned data, regardless of the current client_encoding_"

I might go with this attribute instead.

pnax:
After taking some time to better understand how Perl holds strings as either text string or binary data, I opened #963 that tries to better handle strings representation across the code and sets `pg_enable_utf8 = 0` to always pass binary data to the DBI drivers.
This PR is replaced by #963.

PR #936, opened 2021-12-09
First translation to spanish language

### LINKED ISSUE #910: Translate error messages into Spanish (Backend)
As a new feature of Zonemaster-Backend, in the upcoming release,  error messages are better returned in the GUI, such as incorrect IP address. Those messages can also be translated. After the release, GUI will show the error message in the same language as has been selected.

The error messages comes from Backend via RPC-API. The translation is done in Backend with the use of PO file. For Spanish there is [share/es.po]. Now it is empty, but if "[Instructions for translators]" is followed it will be filled with *msgids* and translations can be added (*msgstrs*).

There will be 19 *msgids* to be translated, most of them very short.

[Instructions for translators]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/Instructions-for-translators.md

[share/es.po]: https://github.com/zonemaster/zonemaster-backend/blob/develop/share/es.po

huguei:
## Purpose

This PR adds a first version of the spanish translation.

## Context

Fixes #910 

## Changes

Just share/es.po. I tried to add this file to MANIFEST but it only contains the .mo files.

## How to test this PR

New language addition.

### DIFF
--- a/share/es.po
+++ b/share/es.po
@@ -2,12 +2,75 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-10-14 11:25+0200\n"
-"PO-Revision-Date: 2021-10-14 11:25+0200\n"
+"POT-Creation-Date: 2021-12-09 15:49-0300\n"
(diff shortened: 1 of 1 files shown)


matsduf (approved):
This looks good as far as I can see. I approve.

PR #937, opened 2021-12-13
Clean up determination of config file default path

mattias-p:
## Purpose

This is mainly a code quality improvement. It also avoids locating the default backend_config.ini if it isn't needed.

## Context

N/A

## Changes

* It avoids executing some code at load time.
* It replaces a global variable (`$Zonemaster::Backend::Config::path`) with a procedure (`Zonemaster::Backend::Config::get_default_path()`).
* It adds code documentation about where the default backend_config.ini file is sought.
* It simplifies the code to determine the config file path.

## How to test this PR

This should not change the behavior of Backend in production context.

A unit test has been added to make sure the Zonemaster::Backend::Config module can be loaded without it being able to locate a backend_config.ini file at load time.

### DIFF
--- a/lib/Zonemaster/Backend/Config.pm
+++ b/lib/Zonemaster/Backend/Config.pm
@@ -5,7 +5,7 @@ use 5.14.2;
 
 our $VERSION = '1.1.0';
 
-use Carp qw( confess );
+use Carp qw( confess croak );
 use Config::IniFiles;
 use Config;
 use File::ShareDir qw[dist_file];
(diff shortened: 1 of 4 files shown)


pnax (approved):

matsduf:
Is there really any good reason to have a default path the configuration file? Backend is always started by start script and in that the environment variable can be set, unless already set. By not having any default file the risk that backend is started by an unintended configuration file lowers and it is also less complex.

mattias-p:
It might make sense to do something in that direction. But the situation is more complex than just starting the daemons from the service scripts. There's also the database creation and migration scripts. And the unit tests. And maybe something else I'm forgetting. I suggest you create a new issue for it.

matsduf:
> It might make sense to do something in that direction. But the situation is more complex than just starting the daemons from the service scripts. There's also the database creation and migration scripts. And the unit tests. And maybe something else I'm forgetting. I suggest you create a new issue for it.

I think database creation and migration script underlines the problem with the built in default configuration files, escpecially "DIST_DIR/backend_config.ini" since it is not really supposed to be used by an installed Backend.

pnax:
The new test file needs to be added to the Manifest

mattias-p:
> I think database creation and migration script underlines the problem with the built in default configuration files, escpecially "DIST_DIR/backend_config.ini" since it is not really supposed to be used by an installed Backend.

Agreed. But that's out of scope for this PR.

I added the unit test to the manifest and reran Travis.

Please re-review.

pnax (approved):

lib/Zonemaster/Backend/Config.pm
```diff
@@ -15,27 +15,52 @@ use Readonly;
...
-}
+=head2 get_default_path
 
-Readonly my @SIG_NAME => split ' ', $Config{sig_name};
+Determine the path for the default backend_config.ini file.
+A list of values and locations are checked and the first match is returned.
+If all places are checked and no file is found, an exception is thrown.
+
+This procedure is idempotent - i.e. if you call this procedure multiple times
+the same value is returned no matter if envirnmental circumstances have changed.
```
matsduf: I do not understand this. Please explain.

Say that the environment variable is unset, but /etc/zonemaster/backend_config.ini exists, then get_default_path() would return "/etc/zonemaster/backend_config.ini".

If then "/etc/zonemaster/backend_config.ini" is removed, would get_default_path() still return "/etc/zonemaster/backend_config.ini".

Please re-write in simplified language.

matsduf (changes requested):

lib/Zonemaster/Backend/Config.pm
mattias-p: Is it better now?

lib/Zonemaster/Backend/Config.pm
```diff
@@ -15,27 +15,53 @@ use Readonly;
...
+
+If a file exists at this path, it is returned.
+DIST_DIR is wherever File::ShareDir installs the Zonemaster-Backend dist.
+
+=back
+
+=cut
+
+sub get_default_path {
+    state $path =
```
matsduf, line 58: I cannot see that `use feature state` is set. Isn't that necessary?

lib/Zonemaster/Backend/Config.pm
matsduf: Yes, thanks.

lib/Zonemaster/Backend/Config.pm
mattias-p: It's enabled by default since Perl 5.10 and we're declaring a minimum version of 5.14.

matsduf (approved):

blacksponge (approved):

PR #942, opened 2022-01-04
Improve get_test_history

### LINKED ISSUE #888: RPCAPI method "get_test_history" includes tests not yet run
If tests are queued to be run, but not yet started, they will still be included in `get_test_history`. The documentation says

> Returns a list of completed tests for a domain.

Furthermore, the not started tests have overall_result "ok":

When the test was created but before it was run:
```json
    {
      "undelegated": false,
      "creation_time": "2021-10-08 14:10:16",
      "overall_result": "ok",
      "id": "19e491e93e2994d5"
    },
```

After it was run:
```
    {
      "creation_time": "2021-10-08 14:10:16",
      "undelegated": false,
      "overall_result": "critical",
      "id": "19e491e93e2994d5"
    },
```

How to replicate this:

1. Create API user and API key
2. Run the follwing command
```
zmb get_test_history --domain xi | jq .
```
which will probably give a JSON without old tests
```
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": []
}
```
2. Run a batch command where domain xi is repeated after some other domains are entered. E.g.:
```
zmb --verbose add_batch_job  --username USER  --api-key APIKEY --domain arpa --domain iis.se --domain afnic.fr --domain errors.com --domain icann.com --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi --domain xi
```
3. Immediately run the following command and check the overall status. All tests are listed with status "ok".
```
zmb get_test_history --domain xi | jq .
```
4. When the batch has been completed (`zmb get_batch_job_result --batch-id BATCH-ID`) run the previous `get_test_history` command again. Now the overall_status is "critical" (since that TLD does not exist).

pnax:
## Purpose

Harmonize the code for `get_test_history` across database engines and fixes #888.

## Context

Fixes #888 

## Changes

* refactoring
* use SQL bind parameters
* remove unused `params` value
* keep finished tests only 

## How to test this PR

see #888 description and example

### DIFF
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -5,7 +5,7 @@ our $VERSION = '1.1.0';
 use Moose;
 use 5.14.2;
 
-use DBI qw(:utils);
+use DBI qw(:utils :sql_types);
 use Digest::MD5 qw(md5_hex);
 use JSON::PP;
 
@@ -209,11 +209,11 @@ sub get_test_history {
 
     my $dbh = $self->dbh;
 
-    my $undelegated = "";
+    my $undelegated = undef;
     if ($p->{filter} eq "undelegated") {
-        $undelegated = "AND undelegated = 1";
+        $undelegated = 1;
     } elsif ($p->{filter} eq "delegated") {
-        $undelegated = "AND undelegated = 0";
+        $undelegated = 0;
     }
 
     my @results;
@@ -225,13 +225,20 @@ sub get_test_history {
             undelegated,
             results
         FROM test_results
-        WHERE progress = 100 AND domain = ? ] . $undelegated . q[
+        WHERE progress = 100 AND domain = ? AND ( ? IS NULL OR undelegated = ? )
         ORDER BY id DESC
         LIMIT ?
         OFFSET ?];
 
     my $sth = $dbh->prepare( $query );
-    $sth->execute( $p->{frontend_params}{domain}, $p->{limit}, $p->{offset} );
+
+    $sth->bind_param( 1, $p->{frontend_params}{domain} );
+    $sth->bind_param( 2, $undelegated, SQL_INTEGER );
+    $sth->bind_param( 3, $undelegated, SQL_INTEGER );
+    $sth->bind_param( 4, $p->{limit} );
+    $sth->bind_param( 5, $p->{offset} );
+
+    $sth->execute();
 
     while ( my $h = $sth->fetchrow_hashref ) {
         $h->{results} = decode_json($h->{results}) if $h->{results};
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -225,7 +225,7 @@ sub get_test_history {
             undelegated,
             results
         FROM test_results
-        WHERE domain = ? ] . $undelegated . q[
+        WHERE progress = 100 AND domain = ? ] . $undelegated . q[
         ORDER BY id DESC
         LIMIT ?
         OFFSET ?];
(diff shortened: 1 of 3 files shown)


lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -196,23 +196,23 @@ sub get_test_history {
...
+    my @results;
+    my $query = q[
+        SELECT
+            id,
+            hash_id,
+            creation_time,
+            undelegated,
+            results
+        FROM test_results
+        WHERE progress = 100 AND domain = ? ] . $undelegated . q[
```
mattias-p, line 208: The `IS` operator could be used here to avoid concatenation of SQL syntax.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -225,11 +225,13 @@ sub get_test_history {
             undelegated,
             creation_time at time zone current_setting('TIMEZONE') at time zone 'UTC' as creation_time
         FROM test_results
-        WHERE domain=" . $dbh->quote( $p->{frontend_params}->{domain} ) . " $undelegated
-        ORDER BY id DESC
-        OFFSET $p->{offset} LIMIT $p->{limit}";
+        WHERE progress = 100 AND domain = ? ] . $undelegated . q[
```
mattias-p, line 228: The `IS NOT DISTINCT FROM` operator could be used here to avoid concatenation of SQL syntax.

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -207,56 +207,34 @@ sub test_results {
...
+    my @results;
+    my $query = q[
+        SELECT
+            id,
+            hash_id,
+            CONVERT_TZ(`creation_time`, @@session.time_zone, '+00:00') AS creation_time,
+            undelegated,
+            results
+        FROM test_results
+        WHERE progress = 100 AND domain = ? ] . $undelegated . q[
```
mattias-p, line 228: The `<=>` operator could be used here to avoid concatenation of SQL syntax.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -225,11 +225,13 @@ sub get_test_history {
...
+        LIMIT ?
```
mattias-p, line 229: Shouldn't we keep the `ORDER BY` directive here?

mattias-p (commented):
This looks good to me.

lib/Zonemaster/Backend/DB/PostgreSQL.pm
pnax: Good catch. I put it back.

lib/Zonemaster/Backend/DB/SQLite.pm
mattias-p: What I had in mind was pretty much what you did here but using SQLite specific syntax for the comparison (i.e `undelegated IS ?`). But I like what you did here better.

mattias-p (dismissed):
Looking good!

mattias-p:
@pnax What does it mean to "keep finished tests only"? Is that the part that actually fixes #888?

Also did you consider possibility of adding an automatic test for this?

pnax:
> What does it mean to "keep finished tests only"? Is that the part that actually fixes #888?

Yes exactly. Would you say that the wording is not very understandable?

> Also did you consider possibility of adding an automatic test for this?

Nope and I think it's good that you asked. I'll provide something.

pnax:
@mattias-p, I've worked on testing and opened #947. Since #947 does a lot of changes, I suggest either:
* adding the test here once #947 is merged
* or merge this PR and add the test later, either in #947 or in a new PR.

mattias-p:
Either way sounds great to me!

mattias-p (approved):

pnax:
@mattias-p: I've done it in #947

PR #943, opened 2022-01-04
Fix unstoppable Testagent daemon

### LINKED ISSUE #905: TestAgent daemon cannot be properly stopped and/or restarted
Currently the agent cannot be properly stopped if the config file is erroneous or the database cannot be reached.

Step to reproduce:
1. prepare a clean install of the Backend
2. start the agent `systemctl start zm-testagent`
3. insert an error in the `backend_conf.ini` file (duplicating a key for instance)
4. stop the agent `systemctl stop zm-testagent`
5. check that the agent is still running `ps aux | grep testagent`

It would be nice to be able to stop the agent even if there are errors in the config.

pnax:
## Purpose

The pre-flight checks were always run when calling `zonemaster_backend_testagent` daemon. If they fail the daemon stops without performing its task. Therefore the daemon cannot be properly stopped if the checks fail. This runs the checks only on daemon `start`.

## Context

Fixes #905

## Changes

Add a check before running pre-flight checks to run them only when calling the agent with the `start` parameter and bypass them for any other parameters.

## How to test this PR

see #905 description

### DIFF
--- a/script/zonemaster_backend_testagent
+++ b/script/zonemaster_backend_testagent
@@ -191,6 +191,23 @@ sub main {
     return;
 }
 
+sub preflight_checks {
+    # Make sure we can load the configuration file
+    $log->debug("Starting pre-flight check");
+    my $initial_config = Zonemaster::Backend::Config->load_config();
+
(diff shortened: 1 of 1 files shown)


mattias-p (approved):
LGTM.

Issue #274, opened 2017-04-01
Feature request: avoid "progress" update for batch tests


aabdnn:
Batch tests are not normally submitted by interactive clients (such as the GUI), and as such, have no need for a progress meter. In fact, when you submit domains in a batch, there is no way, as far as I can see, to get the progress of an individual test of the batch.

Can the wbdaemon be modified so that it doesn't bother updating the "progress" column of the database while running? It should only change progress from 0 to 1 at the start of the test, and then to 100 at completion. This avoids extra unnecessary writes to the database.

pnax:
Addressed by #944.
However note the following side effect. When running batches and single tests on the same instance, single tests won't have their progress updated if they are reusing tests created in a batch, see #979.

PR #944, opened 2022-01-04
Bypass update of "progress" for batch tests

pnax:
## Purpose

When running batches, updating the `progress` while the test is running will perform a lot of unnecessary database writes which may lead to performance issues. This remove the `progress` database update for a test with a `batch_id`.

## Context

Addresses #274

## Changes

Extract the `batch_id` of a test when retrieving the next test so that we can check if the test belongs to a batch or not and bypass the `progress` update or not.

## How to test this PR

Create a batch job and check that the `progress` of each test goes from 0 to 1 and then from 1 to 100 without any steps.
Create a test and check that the `progress` of the test goes from 0 to 1 and then gradually until 100.

### DIFF
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -65,8 +65,6 @@ sub run {
 
     my $params;
 
-    my $progress = $self->{_db}->test_progress( $test_id, 1 );
-
     $params = $self->{_db}->get_test_params( $test_id );
 
     my %methods = Zonemaster::Engine->all_methods;
@@ -172,7 +170,7 @@ sub run {
 
     $self->{_db}->test_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
 
-    $progress = $self->{_db}->test_progress( $test_id );
+    $self->{_db}->test_progress( $test_id );
 
     return;
 } ## end sub run
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -84,47 +84,49 @@ sub run {
     # used for progress indicator
     my ( $previous_module, $previous_method ) = ( '', '' );
 
-    # Callback defined here so it closes over the setup above.
-    Zonemaster::Engine->logger->callback(
-        sub {
-            my ( $entry ) = @_;
+    if ( $show_progress ) {
+        # Callback defined here so it closes over the setup above.
+        Zonemaster::Engine->logger->callback(
+            sub {
+                my ( $entry ) = @_;
 
-            foreach my $trace ( reverse @{ $entry->trace } ) {
-                foreach my $module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
-                    if ( index( $trace->[1], $module_method ) > -1 ) {
-                        my $percent_progress = 0;
-                        my ( $module ) = ( $module_method =~ /(.+::)[^:]+/ );
-                        if ( $previous_module eq $module ) {
-                            $counter_for_progress_indicator{executed}{$module_method}++;
-                        }
-                        elsif ( $previous_module ) {
-                            foreach my $planned_module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
-                                $counter_for_progress_indicator{executed}{$planned_module_method}++
-                                  if ( $counter_for_progress_indicator{planned}{$planned_module_method} eq
-                                    $previous_module );
+                foreach my $trace ( reverse @{ $entry->trace } ) {
+                    foreach my $module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
+                        if ( index( $trace->[1], $module_method ) > -1 ) {
+                            my $percent_progress = 0;
+                            my ( $module ) = ( $module_method =~ /(.+::)[^:]+/ );
+                            if ( $previous_module eq $module ) {
+                                $counter_for_progress_indicator{executed}{$module_method}++;
                             }
-                        }
-                        $previous_module = $module;
+                            elsif ( $previous_module ) {
+                                foreach my $planned_module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
+                                    $counter_for_progress_indicator{executed}{$planned_module_method}++
+                                      if ( $counter_for_progress_indicator{planned}{$planned_module_method} eq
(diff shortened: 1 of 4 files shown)


lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -126,7 +124,7 @@ sub run {
 
             $counter{ uc $entry->level } += 1;
         }
-    );
+    ) if ( $show_progress );
```
mattias-p, line 127: Would you consider using the prefix syntax for the `if` statement? For readability.

script/zonemaster_backend_testagent
```diff
@@ -149,30 +149,32 @@ sub main {
...
         Zonemaster::Backend::Metrics::timing("zonemaster.testagent.fetchtests_duration_seconds", tv_interval($fetch_test_timer) * 1000);
 
-        if ( $id ) {
-            $log->info( "Test found: $id" );
-            if ( $self->pm->start( $id ) == 0 ) {    # Forks off child process
-                $log->info( "Test starting: $id" );
+        my $show_progress = defined $batch_id ? 0 : 1;
+
+        if ( $test_id ) {
+            $log->info( "Test found: $test_id" );
```
mattias-p, line 163: Maybe replace `info()` with `infof()` while we're at it?

matsduf:
The consequence of this change would be that if you start a manual test with the same parameters as a just started batch test then there would be no progress on that manual test. A solution -- or work-around -- for this would be to prevent tests started as batch tests to be reused for a manual test.

lib/Zonemaster/Backend/TestAgent.pm
pnax: done

script/zonemaster_backend_testagent
pnax: I replaced the next occurrences as well.

matsduf:
I think we should first do a change so that batch tests are never reused for ordinary tests. All tests should of course come up in test history.

pnax:
> I think we should first do a change so that batch tests are never reused for ordinary tests. All tests should of course come up in test history.

Yes, there is indeed an issue here. I think we started discuss this in another thread, and I wanted to try to come with something that would "properly" solve this. I'll try to be back with something.

pnax:
Regarding the fact that tests started as batches, and then called as single test won't have any progress bar anymore, I haven't found any neat solution to this problem yet.
However I'm wondering if this is a strong blocking issue to solve before going further with this PR. The current behavior when it comes to test reuse between batches and single test is not properly defined and handled. I suggest we move on with this suggestion and try to find a good way to address the case when test is reused in different context.
Furthermore I do not think this use case happens very often, the impact to merge this PR would be limited I guess (this should not hide the fact that there is still an issue to address here).

@matsduf would it be fine to merge this PR and let us come with a proper solution to handle tests reuse in a near future?

matsduf:
> Regarding the fact that tests started as batches, and then called as single test won't have any progress bar anymore, I haven't found any neat solution to this problem yet.

My suggestion is that single tests never reuses tests in a batch, unless then test is completed. This would make it possible to handle tests in batch differently from single tests.

matsduf:
Will the proposal set progress to 1 when the test starts? Else I think that should be changed. It should be possible to see the status of a test in the database. See #960.

pnax:
> What is the problem with my proposal?

It contradicts what we have discussed and concluded [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-901731760), [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-901740711) and [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-919807767).

> Will the proposal set progress to 1 when the test starts?

This won't change the current behavior. When an agent retrieves a test, it sets the `progress` field to 1 and the `started_at` field to the current time (in UTC) (see `test_progress()`).

matsduf:
> > What is the problem with my proposal?
> 
> It contradicts what we have discussed and concluded [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-901731760), [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-901740711) and [here](https://github.com/zonemaster/zonemaster-backend/issues/840#issuecomment-919807767).

Now we have a new context. The assumption then was that tests started in a batch behaved the same way. My suggestion for this PR is that a test started in a batch is not reused unless it has progress==100.

matsduf (dismissed):
I approve. When this PR is merged, please create an issue tagged BUG on the progress bar issue for single tests.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -86,47 +84,49 @@ sub run {
...
-                        my ( $module ) = ( $module_method =~ /(.+::)[^:]+/ );
-                        if ( $previous_module eq $module ) {
-                            $counter_for_progress_indicator{executed}{$module_method}++;
-                        }
-                        elsif ( $previous_module ) {
-                            foreach my $planned_module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
-                                $counter_for_progress_indicator{executed}{$planned_module_method}++
-                                  if ( $counter_for_progress_indicator{planned}{$planned_module_method} eq
-                                    $previous_module );
+    if ( defined $show_progress and $show_progress ) {
```
mattias-p, line 87: This would be enough:

```suggestion
    if ( $show_progress ) {
```

mattias-p (dismissed):
LGTM

mattias-p (dismissed):

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -172,7 +172,7 @@ sub run {
 
     $self->{_db}->test_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
 
-    $progress = $self->{_db}->test_progress( $test_id );
+    $self->{_db}->test_progress( $test_id );
```
blacksponge, line 175: This line is not necessary, the only thing it does is returning the progress, you should be able to remove it safely.

blacksponge (commented):
Looks good to me, I still have a comment though.

lib/Zonemaster/Backend/TestAgent.pm
pnax: good point, I've removed it.

blacksponge (approved):
lgtm

PR #945, opened 2022-01-13
Fix travis-perl error

pnax:
An update in travis-perl added comment to the init file. This makes the `eval` call fail because the script become a one-liner in this case with a comment in the middle.

This PR is about removing commented lines from the script before running it.

https://github.com/travis-perl/travis-perl.github.io/commit/34d18cc150497a00dc7719fea2472c183e6607ad
    
Credits to @blacksponge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,9 +79,9 @@ before_install:
   - export PERL5LIB=/usr/share/perl5
 
     # Provide cpanm helper
-    # the eval call combines all lines into one, therefore comments should be
-    # removed before executing the script
-  - eval $(curl https://travis-perl.github.io/init | grep -v '#') --auto
+    # quoting preserves newlines in the script and then avoid error when
+    # the script contains comments
+  - eval "$(curl https://travis-perl.github.io/init)" --auto
 
     # Zonemaster LDNS needs a newer version of Module::Install
   - cpan-install Module::Install Module::Install::XSUtil
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,8 @@ before_install:
   - export PERL5LIB=/usr/share/perl5
 
     # Provide cpanm helper
+    # the eval call combines all lines into one, therefore comments should be
+    # removed before executing the script
   - eval $(curl https://travis-perl.github.io/init | grep -v '#') --auto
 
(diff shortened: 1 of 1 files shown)


matsduf:
What a hack! :-)

matsduf:
If you update the PR description I guess this is an acceptable work-around.

.travis.yml
```diff
@@ -79,7 +79,7 @@ before_install:
   - export PERL5LIB=/usr/share/perl5
 
     # Provide cpanm helper
-  - eval $(curl https://travis-perl.github.io/init) --auto
+  - eval $(curl https://travis-perl.github.io/init | grep -v '#') --auto
```
blacksponge, line 82: maybe you could update the comment above the line in so when we'll see that in the future we know what we are doing

blacksponge (commented):
one comment otherwise looks good to me :)

.travis.yml
pnax: Good idea.
I was also wondering if we should do a better comment removal. I mean what if the comment is at the end of a line?
```
$( curl ... | sed 's/#.+//' )
```

matsduf: I guess there is good chance that the comments are completely removed, but else it would be safer to just remove the comment, and not the entire line. On the other hand, just removing by "#" is never completely safe. It could be part of a command.

matsduf (approved):
I think this is good enough. We might to modify it later anyway.

.travis.yml
```diff
@@ -79,7 +79,9 @@ before_install:
   - export PERL5LIB=/usr/share/perl5
 
     # Provide cpanm helper
-  - eval $(curl https://travis-perl.github.io/init) --auto
+    # the eval call combines all lines into one, therefore comments should be
+    # removed before executing the script
+  - eval $(curl https://travis-perl.github.io/init | grep -v '#') --auto
```
mattias-p, line 84: Couldn't we just quote it?

```suggestion
  - eval "$(curl https://travis-perl.github.io/init)" --auto
```

pnax: you just killed the game ;)

matsduf (approved):

PR #946, opened 2022-01-18
Do not expose database primary key

pnax:
## Purpose

There is no use of the database `id` since we have implemented the `hash_id`. Therefore, I think it is safe and good that we remove the `id` from the results in the `get_test_results` method.

## Context

Follows deprecation done in #949 (v2022.1).

## Changes

Update the API `get_test_results` method.

## How to test this PR

A call `get_test_results` should not give the database `id` of the test, but only its `hash_id`.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -932,7 +932,6 @@ Example response:
   "result": {
     "creation_time": "2016-11-15 11:53:13.965982",
     "created_at": "2016-11-15T11:53:13Z",
-    "id": 25,
     "hash_id": "c45a3f8256c4a155",
     "params": {
       "ds_info": [],
@@ -991,7 +990,6 @@ An object with the following properties:
   The time in UTC at which the *test* was created.
 * `"created_at"`: A [*timestamp*][Timestamp]. The time in UTC at which the *test*
   was created.
-* `"id"`: **Deprecated** (planned removal: v2022.2). An integer.
 * `"hash_id"`: A [*test id*][Test id]. The *test* in question.
 * `"params"`: See below.
   `start_domain_test` when the *test* was started.
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -269,7 +269,6 @@ sub select_test_results {
     my ( $hrefs ) = $self->dbh->selectall_hashref(
         q[
             SELECT
-                id,
                 hash_id,
                 created_at AS creation_time,
                 params,
--- a/t/test01.t
+++ b/t/test01.t
@@ -118,7 +118,7 @@ subtest 'API calls' => sub {
 
     subtest 'get_test_results' => sub {
         my $res = $rpcapi->get_test_results( { id => $hash_id, language => 'en_US' } );
-        is( $res->{id}, $test_id, 'Retrieve the correct "id"' );
+        ok( ! defined $res->{id}, 'Do not expose primary key' );
         is( $res->{hash_id}, $hash_id, 'Retrieve the correct "hash_id"' );
         ok( defined $res->{params}, 'Value "params" properly defined' );
         ok( defined $res->{creation_time}, 'Value "creation_time" properly defined' );


matsduf:
I am not against the change as such, but this is a change in the API. Maybe it should be deprecated in next release and removed in a future release.

pnax:
Fair point. We can move this PR to milestone v2022.2. And I can open a new one deprecating the value. I guess here we only need to update the API documentation with a _Deprecated_ message next to the `id` line.

matsduf:
> I guess here we only need to update the API documentation with a _Deprecated_ message next to the `id` line.

Yes, plus remember to point that out in the release notes so a PR of its own with a clear headline helps that.

Does v2022.2 give enough time for those with customized GUI and other interfaces to RPC-API?

mattias-p:
> Does v2022.2 give enough time for those with customized GUI and other interfaces to RPC-API?

Good question. I think the answer is yes. We don't have any methods that accept the id as an input, so it seems a bit far fetched that someone should be using it.

pnax:
The deprecation is done in #949

matsduf:
> > Does v2022.2 give enough time for those with customized GUI and other interfaces to RPC-API?
> 
> Good question. I think the answer is yes. We don't have any methods that accept the id as an input, so it seems a bit far fetched that someone should be using it.

My question is not if they are used or not, but they can still be required by an RPC-API client that checks for the existence since it is in the specification.

mattias-p:
> > > Does v2022.2 give enough time for those with customized GUI and other interfaces to RPC-API?
> > 
> > 
> > Good question. I think the answer is yes. We don't have any methods that accept the id as an input, so it seems a bit far fetched that someone should be using it.
> 
> My question is not if they are used or not, but they can still be required by an RPC-API client that checks for the existence since it is in the specification.

Why would you write an RPC-API client like that?

matsduf:
Since the keys are useless, i.e. cannot be used from the outside, unless the same hashing algorithm could be applied on them, I cannot see how they can be misused. 

To be sure `id` could be set to `1` in `get_test_results`. It is only promised to be an integer, and it cannot be used anywhere. This could be done now.

pnax:
> Does v2022.2 give enough time for those with customized GUI and other interfaces to RPC-API?

I haven't properly answered to this question. I do think this is fine and gives enough time. There are 6 months between 2 releases. This means people have 6 months to get up to date with the changelog.
If I understand correctly you wonder if this might not break custom Zonemaster installations. Maybe, maybe not. Since they are custom I expect people running them going through the Changelog before any upgrade. Therefore pointing deprecated parts in a release and removing them in the next release sounds perfectly fine to me.
To my eyes Zonemaster is a product with slow updates. This has pros and cons. I would then argue that we should not make it even slower by waiting too long to remove deprecated code (and again 6 months gives, I think, some time to prepare for the release removing the feature). 

> To be sure id could be set to 1 in get_test_results. It is only promised to be an integer, and it cannot be used anywhere. This could be done now.

Not sure what would be the gain here. If we think people are using custom Zonemaster using this `id`, this would probably break it (only for installation using this `id` field, which I guess is none). So we can go even one step further and remove the field directly and miss the deprecation step. But if we want to follow a deprecation-removal process, I think we should do it in 2 releases.

blacksponge:
> Why would you write an RPC-API client like that?

It could be argued that this could be the default behaviour when deserialising the response in a strongly typed languages (example in rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b0a3444d078ce1f17fc8cc727559fcb3). Even if the field is not used afterwards. But I don't know how many clients (if there are any) would be in this case.

mattias-p:
@blacksponge Yeah, I guess there are probably applications out there that are implemented that way. Though best practice ought to be to only include the fields in your struct that you actually use. That way fields in the input that you don't use are simply ignored, and if the server suddenly stops including them that doesn't won't hurt you. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6976bf1312cb4640aa948af918fca9c
In the case of rust/serde that also fixes the compiler warnings you get for unused fields.

matsduf:
If we see or suspect that there could be a security issue with exposing the primary key we should try to remove it now. Since it is not used in any key, the client can only expect to get the field with an integer. Therefore we could replace the real key with some other integer.

If we cannot see that it could be a security issue, we should still remove it to avoid having meaningless fields, but then we could wait until v2023.1 to give all users more time to update their clients.

There could be clients that assume that the field is there even if the client does not use it for something useful.

mattias-p:
> There could be clients that assume that the field is there even if the client does not use it for something useful.

Any such clients are clearly making unreasonable assumptions. It doesn't seem reasonable to me for us to spend our resources trying to keep such clients from crashing.

matsduf (approved):

Issue #806, opened 2021-06-29
Cleaning up the database used in unit tests


mattias-p:
Today we have unit tests that use a special database as part of their infrastructure. This works well in CI because we get a fresh machine every time. However when running tests locally you have to clean up this database between each run or the tests will fail because test data from one run collides with test data from the next.

The unit tests should perform the clean up as part of their own setup. To ensure that the business logic can run without privileges to clean up the database, a separate database user should be used for the clean up.

mattias-p:
No, we still have this problem.

pnax:
Done in #947 with commits 6a6b5cddbb20cb56fc004e169dc66d9f6dae6d62 and 8d768803a5f00595405f554c0c4d63f8dc7738d5

Issue #747, opened 2021-04-19
Running tests without littering in the file system


mattias-p:
There should be an easy way to run unit tests with SQLite backed by an in-memory database or alternatively an auto-cleaned up temp file. This is [supported by DBD::SQLite](https://metacpan.org/pod/DBD::SQLite#Database-Name-Is-A-File-Name) as long as we don't need to fork.

This may be complicated by the fact that this feature doesn't really make sense in production mode since the database needs to be shared across two different daemons. If support for this is added to the config file the documentation should probably clarify that this is only intended for testing.

pnax:
Closing the issue since I think #947 fixes it. Feel free to re-open it if you think this is not the case.

PR #947, opened 2022-01-18
Improve testing

pnax:
## Purpose

Offer better automated testing coverage of the Backend code.

## Context

Addresses #806, #747

## Changes

Most of the changes happened in `t/test01.t` with some refactoring and adding new tests.
A new `t/test_profile.json` file has been created to allow using a specific profile while testing (for instance to bypass tests leaking through the network).
`t/test01.t` performs the tables creation and cleanup (see #806). Therefore `.travis.yml` has been updated to remove the call to `create_db()`.

Add new tests to check #942 

## How to test this PR

All tests should pass. Travis (once back on its feet) should not complain.


t/test01.t
```diff
@@ -136,15 +137,25 @@ subtest 'add a first test' => sub {
     is( $hash_id_db, $hash_id , 'Correct hash_id in database' );
 
     # test test_progress API
-    is( $backend->test_progress( { test_id => $hash_id } ), 0 , 'API test_progress -> OK');
+    $progress = $backend->test_progress( { test_id => $hash_id } );
```
mattias-p, line 152: Maybe declare the variable anew in each subtest where it's used? The value is not supposed to survive to the next subtest, so why should the variable?

t/test01.t
```diff
@@ -383,6 +387,11 @@ if ( $ENV{ZONEMASTER_RECORD} ) {
 
 done_testing();
 
+# cleaning up
+$dbh->do( "DROP table test_results" );
+$dbh->do( "DROP table users" );
+$dbh->do( "DROP table batch_jobs" );
```
mattias-p, line 420: Why not drop the tables (IF EXISTS) right before they're created? That way you can study the state of the database after the test is run, in case you need to. And when the test dies in the middle of a run you don't have to drop the dirty tables manually before your next run.

mattias-p (approved):
This looks really good to me!

t/test01.t
pnax: indeed there is no reason to declare `progress` only once, I applied your suggestion.

mattias-p (approved):
Looking good!

PR #949, opened 2022-01-24
Deprecate the result 'id' field in get_test_results

pnax:
## Purpose

The "id" field in the "result" object is leaking the PRIMARY KEY from the database. It should then be removed from the results. Furthermore there is no use of it.

In order to properly remove this field from the results, it needs to be deprecated first.

## Context

As discussed in #946, the field should be first documented as _Deprecated_.

## Changes

Deprecate the "id" field from the "result" object in `get_test_results`.

## How to test this PR

Documentation only

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -982,7 +982,7 @@ An object with the following properties:
 #### `"result"`
 
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
-* `"id"`: An integer.
+* `"id"`: **Deprecated**. An integer.
 * `"hash_id"`: A *test id*. The *test* in question.
 * `"params"`: See below.
   `start_domain_test` when the *test* was started.
--- a/docs/API.md
+++ b/docs/API.md
@@ -982,7 +982,7 @@ An object with the following properties:
 #### `"result"`
 
 * `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
-* `"id"`: **Deprecated**. An integer.
+* `"id"`: **Deprecated** (will be removed in v2022.2). An integer.
 * `"hash_id"`: A *test id*. The *test* in question.
 * `"params"`: See below.
   `start_domain_test` when the *test* was started.


mattias-p (dismissed):
LGTM.

matsduf (changes requested):
1. I think we should state here that the key will be removed from the RPC-API output by some specific release.
2. I think we should set that release to be v2023.1.

mattias-p:
> * I think we should state here that the key will be removed from the RPC-API output by some specific release.

I agree with this.

> * I think we should set that release to be v2023.1.

That seems a bit excessive. There is no use case for this feature. Nobody is going to miss it. But then having it around for a while longer doesn't seem like much of a burden either.

pnax:
> I think we should state here that the key will be removed from the RPC-API output by some specific release.

As soon as we agree on which release, I'll provide an update.

> I think we should set that release to be v2023.1.

I don't see the gain in waiting so long. From a developing point of view, this would only let our codebase being obstructed with things to remove in 1 year time. And I would argue for a shorter cleaning process. From a user point of view, there is a 6 months delay between releases and a detailed Changelog at each release. So I think this is perfectly fine to keep me up to date with latest changes and possible removal since they are documented 6 months ahead.

matsduf:
> I don't see the gain in waiting so long. From a developing point of view, this would only let our codebase being obstructed with things to remove in 1 year time. And I would argue for a shorter cleaning process. From a user point of view, there is a 6 months delay between releases and a detailed Changelog at each release. So I think this is perfectly fine to keep me up to date with latest changes and possible removal since they are documented 6 months ahead.

I cannot see the problem of keeping it around for a year. It gives longer time for users to do the update. What would the problem be?

pnax:
@matsduf is it ok if I remove the field in v2022.2?

matsduf:
> @matsduf is it ok if I remove the field in v2022.2?

Yes, it is OK (but I still do not see why this hurry). Can you update the API.md with the removal version?

pnax:
* I've rebased on top of develop and resolved the conflict within the API.md document
* I added a mention about the removal of the field in v2022.2

matsduf (approved):

PR #958, opened 2022-03-02
Allow DB adapter consturction without actually connecting

mattias-p:
## Purpose

Remove double connection to the database on startup.

## Context

This came up as an alternative solution to a drive-by fix in #955.

## Changes

This PR decouples the opening of database connections from the construction of Zonemaster::Backend::RPCAPI and Zonemaster::Backend::TestAgent instances. This allow us greater flexibility in how we can express initialization code for our daemons and it also makes the initialization code more understandable.

## How to test this PR

Run the [smoke test](https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/Installation.md#61-smoke-test).

### DIFF
--- a/script/zonemaster_backend_testagent
+++ b/script/zonemaster_backend_testagent
@@ -133,8 +133,6 @@ sub main {
     local $SIG{TERM} = $catch_sigterm;
 
     my $agent = Zonemaster::Backend::TestAgent->new( { config => $self->config } );
-    # Disconnect from database in parent to avoid sharing the connection between children
-    $agent->{_db}->dbh->disconnect;
 
     while ( !$caught_sigterm ) {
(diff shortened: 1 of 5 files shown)


script/zonemaster_backend_testagent
```diff
@@ -133,8 +133,6 @@ sub main {
     local $SIG{TERM} = $catch_sigterm;
 
     my $agent = Zonemaster::Backend::TestAgent->new( { config => $self->config } );
-    # Disconnect from database in parent to avoid sharing the connection between children
-    $agent->{_db}->dbh->disconnect;
```
blacksponge, line 137: Nice thing that this hack is not needed any more :)

blacksponge (approved):
Looks good to me

pnax (approved):

PR #962, opened 2022-03-09
Use Perl time on test creation

pnax:
## Purpose

From #769 

> N.B. The database schemas still use references to NOW() in default value definitions for certain fields, but all insert queries are updated to make sure that those default values are never used.

It seems we missed some insert queries. This PR fixes that.

## Context

Follow-up on #769

## Changes

Set the `creation_time` with Perl time.

## How to test this PR

Check that all `creation_time` in database are UTC.


matsduf (approved):
I approve.

Issue #860, opened 2021-09-09
"get_batch_job_result" should create an error if batch ID is not defined


matsduf:
Batch 18 has not been created. One should expect `get_batch_job_result` to produce an error when an undefined batch ID is used.

```
zmb get_batch_job_result --batch-id 18 | jq .
{
  "result": {
    "nb_finished": 0,
    "nb_running": 0
  },
  "jsonrpc": "2.0",
  "id": 1
}
```

pnax:
Solved with #965

PR #965, opened 2022-03-23
Produce an error on unknown batch

pnax:
## Purpose

When calling "get_batch_job_result", if the batch is undefined then produce an error `Zonemaster::Backend::Error::ResourceNotFound`.

## Context

#860 

## Changes

* make the domain list in `add_batch_job` non empty
* update `get_batch_job_result()` in DB.pm
* document the error message
* small cleanup: remove the `user_exists` abstraction

## How to test this PR

Can be manually tested using #860.
A new test has also been added.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1433,9 +1433,21 @@ An object with the following properties:
 
 #### `"error"`
 
->
-> TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
->
+If the `batch_id` is undefined the following error is returned:
+
+```json
+{
+  "id": 1,
+  "error": {
+    "data": {
+      "batch_id": "10"
+    },
+    "message": "Unknown batch",
+    "code": -32603
+  },
+  "jsonrpc": "2.0"
+}
+```
 
 ## API method: `get_test_params`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -1287,7 +1287,7 @@ An object with the following properties:
 
 * `"username"`: A *username*, required. The name of the account of an authorized user.
 * `"api_key"`: An *api key*, required. The api_key associated with the username.
-* `"domains"`: A list of *domain names*, required. The domains to be tested.
+* `"domains"`: A non-empty list of *domain names*, required. The domains to be tested.
 * `"test_params"`: As described below, optional. (default: `{}`)
 
 The value of `"test_params"` is an object with the following properties:
@@ -1346,6 +1346,25 @@ Trying to add a batch when wrong *username* or *api key* is used:
   },
   "id": 1,
   "jsonrpc": "2.0"
+}
+```
+
+Trying to add a batch with an empty list of domains:
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "data": [
+      {
+        "message": "Not enough items: 0/1.",
+        "path": "/domains"
+      }
+    ],
+    "message": "Invalid method parameter(s).",
+    "code": "-32602"
+  }
 }
 ```
 
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -320,6 +320,19 @@ sub user_authorized {
     return $id;
 }
 
+sub batch_exists_in_db {
+    my ( $self, $batch_id ) = @_;
+
+    my $dbh = $self->dbh;
+    my ( $id ) = $dbh->selectrow_array(
+        q[ SELECT id FROM batch_jobs WHERE id = ? ],
+        undef,
+        $batch_id
+    );
+
+    return $id;
+}
+
 # Standard SQL, can be here
 sub get_test_request {
     my ( $self, $queue_label ) = @_;
@@ -367,6 +380,9 @@ sub get_test_params {
 sub get_batch_job_result {
     my ( $self, $batch_id ) = @_;
 
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Unknown batch", data => { batch_id => $batch_id } )
+        unless defined $self->batch_exists_in_db( $batch_id );
+
     my $dbh = $self->dbh;
 
     my %result;
@@ -389,9 +405,6 @@ sub get_batch_job_result {
             $result{nb_running}++;
         }
     }
-
-    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Unknown batch", data => { batch_id => $batch_id } )
-        if ( $result{nb_running} == 0 and $result{nb_finished} == 0 );
 
     return \%result;
 }
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -105,15 +105,6 @@ sub dbh {
     return $self->dbhandle;
 }
 
-sub user_exists {
-    my ( $self, $user ) = @_;
-
-    die Zonemaster::Backend::Error::Internal->new( reason => "username not provided to the method user_exists")
-        unless ( $user );
-
-    return $self->user_exists_in_db( $user );
-}
-
 sub add_api_user {
     my ( $self, $username, $api_key ) = @_;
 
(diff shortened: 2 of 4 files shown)


docs/API.md
```diff
@@ -1414,9 +1414,22 @@ An object with the following properties:
 
 #### `"error"`
 
->
-> TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
->
+If the `batch_id` is undefined or returned no results, the followin error is
+returned:
```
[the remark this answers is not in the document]
pnax: TBH I haven't checked the code. And the logic I implemented is rather permissive (it's not looking in database if the batch_id exists in the `batch_jobs` table but looks at the number of results).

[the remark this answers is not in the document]
pnax: > Is it possible to create an empty batch? The specification for add_batch_job in API.md does not say that parameter "domains" must be non-empty, which I think it should.

Looks like we don't have any restriction in the code either. I'll provide an update.

pnax:
* I've updated the code and documentation to make the domain list non-empty when calling `add_batch_job`.
* I adapted the PR accordingly

docs/API.md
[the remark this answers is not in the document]
matsduf: Why does not the code look for the existence of the batch_id instead? Such code would be easier to understand in the future.

docs/API.md
```diff
@@ -1349,6 +1349,25 @@ Trying to add a batch when wrong *username* or *api key* is used:
...
+Trying to add a batch with an empty list of domains:
+
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "error": {
+    "data": [
+      {
+        "message": "Not enough items: 0/1.",
```
matsduf, line 1361: How should I interpret "0/1"?

lib/Zonemaster/Backend/DB.pm
```diff
@@ -390,6 +390,9 @@ sub get_batch_job_result {
         }
     }
 
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Unknown batch", data => { batch_id => $batch_id } )
+        if ( $result{nb_running} == 0 and $result{nb_finished} == 0 );
+
```
matsduf, line 395: Why does not the code look for undefined batch ID instead? That would be easier to understand in the future.

docs/API.md
pnax: This means that there are 0 item out of 1 expected (this is defined [in the module](https://metacpan.org/release/JHTHORSEN/JSON-Validator-5.08/source/lib/JSON/Validator/Error.pm#L14)).

lib/Zonemaster/Backend/DB.pm
pnax: I was miser on database connection. I'll add a look into the DB.

pnax: I updated the code.

docs/API.md
matsduf: A list of at list one item is expected. Is the list the missing item? For me it looks like we expect exactly 1 domain name.

Why not [contains] in the module?

[contains]: https://metacpan.org/release/JHTHORSEN/JSON-Validator-5.08/source/lib/JSON/Validator/Error.pm#L16

lib/Zonemaster/Backend/DB.pm
matsduf: Nice.

docs/API.md
matsduf: Is the purpose of "Not enough items: 0/1." that it should go to a human reader? Or is "0/1" interpreted by a machine?

Again, should I interpret "1" as one list or one item?

matsduf, line 1361: I suggest a nicer message to the human:
```
"message": "List of domains must not be empty.",
```

matsduf (changes requested):

lib/Zonemaster/Backend/DB.pm
```diff
@@ -367,6 +364,9 @@ sub get_test_params {
 sub get_batch_job_result {
     my ( $self, $batch_id ) = @_;
 
+    die Zonemaster::Backend::Error::ResourceNotFound->new( message => "Unknown batch", data => { batch_id => $batch_id } )
+        unless defined $self->batch_exists_in_db( $batch_id );
```
mattias-p, line 368: Couldn't we just look at the number of keys in %result to make this determination without and extra SQL query?

lib/Zonemaster/Backend/DB.pm
```diff
@@ -105,23 +105,14 @@ sub dbh {
...
-
 sub add_api_user {
     my ( $self, $username, $api_key ) = @_;
 
     die Zonemaster::Backend::Error::Internal->new( reason => "username or api_key not provided to the method add_api_user")
         unless ( $username && $api_key );
 
     die Zonemaster::Backend::Error::Conflict->new( message => 'User already exists', data => { username => $username } )
-        if ( $self->user_exists( $username ) );
+        if ( $self->user_exists_in_db( $username ) );
```
mattias-p, line 115: Have you considered taking this cleanup one step further? It looks to me like add_api_user only ever throws "add_api_user_to_db not successful" if there is a race between user_exists_in_db() and add_api_user_to_db(). If this is true we could simply remove the user_exists_in_db() check and have the $result check throw 'User already exists' instead. WDYT?

mattias-p (approved):
LGTM. I have a questions though.

lib/Zonemaster/Backend/DB.pm
pnax: I did this at first, then after the [discussion here](https://github.com/zonemaster/zonemaster-backend/pull/965#discussion_r835040107) I added the commit https://github.com/zonemaster/zonemaster-backend/pull/965/commits/bc96642dca8a34b2a6a24ebc6be6b7147969bc60. I can easily remove it if we think it is better to avoid an extra call to the database (which was my first thought when starting to work on this PR).

lib/Zonemaster/Backend/DB.pm
mattias-p: I just took for granted that it was already unique.. I'm fine with a separate PR.

lib/Zonemaster/Backend/DB.pm
matsduf: The number of lookups based on batch ID will always be small.

mattias-p: Yeah, most likely. What's at stake here is an occasional extra SQL query and about 10 extra lines of code (ignoring any documentation and unit tests). I don't think that's huge and I won't quarrel over it.

matsduf: If the check could be done more efficiently some other way it could also be fine if the magic in that solution is documented in the code.

docs/API.md
pnax: This would mean overwrite the module output for this message. So far I haven't found a nice way to do it. The solution I currently have results in some tweaking that heavily relies on the module internals. And I don't feel confident in using such solution because of possible failures due to a module internal update.

matsduf: Then, would the following work?
```
"message": "List of domains must not be empty: 0/1",
```

pnax: The approach I took is based on the [EXPERIMENTAL hack](https://metacpan.org/pod/JSON::Validator::Error#message) from the documentation:

```diff
diff --git a/lib/Zonemaster/Backend/RPCAPI.pm b/lib/Zonemaster/Backend/RPCAPI.pm
index ea6cf46..89f82b3 100644
--- a/lib/Zonemaster/Backend/RPCAPI.pm
+++ b/lib/Zonemaster/Backend/RPCAPI.pm
@@ -40,6 +40,11 @@ our %json_schemas;
 our %extra_validators;
 my $recursor = Zonemaster::Engine::Recursor->new;
 
+# overwrite the 'minItems' error message
+$JSON::Validator::Error::MESSAGES = {
+    array => {minItems => N__ 'At least %4 items expected but received %3'},
+};
+
 sub joi {
     return JSON::Validator::Joi->new;
 }
```
If you have a better approach that is not documented as experimental we can try to go with it. Otherwise I'm afraid I don't have a solution to properly set the error message for now.

matsduf (dismissed):
I accept as is. The load on this function is low.

pnax:
I chose to keep the SQL query for now (sorry for the double force-push).

PR #967, opened 2022-03-31
Use ISO 8601 to represent datetimes

pnax:
## Purpose

Standardize the way datetimes are represented in results from the RCPAPI.

## Context

This work has been first conducted in #964, but due to a possible breaking change, this has been moved to another PR.

## Changes

* New `created_at` field in `get_test_results` and `get_test_history` with the time represented using ISO 8601 `2022-03-28T12:07:40Z`
* Deprecate the `creation_time` field in `get_test_results` and `get_test_history`
* Update API documentation

## How to test this PR

Check that there is a new `created_at` field in the output from the `get_test_result` and `get_test_history` methods is in ISO 8601 format.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -986,11 +986,11 @@ An object with the following properties:
 
 #### `"result"`
 
-* `"creation_time"`: **Deprecated**. A [*deprecated timestamp*][Timestamp deprecated].
+* `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
   The time at which the *test* was created (in UTC).
 * `"created_at"`: A [*timestamp*][Timestamp]. The time at which the *test* was created
   (in UTC).
-* `"id"`: **Deprecated** (will be removed in v2022.2). An integer.
+* `"id"`: **Deprecated** (planned removal: v2022.2). An integer.
 * `"hash_id"`: A [*test id*][Test id]. The *test* in question.
 * `"params"`: See below.
   `start_domain_test` when the *test* was started.
@@ -1098,7 +1098,7 @@ The value of "frontend_params" is an object with the following properties:
 An object with the following properties:
 
 * `"id"` A *test id*.
-* `"creation_time"`: **Deprecated**. A [*deprecated timestamp*][Timestamp deprecated].
+* `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
   The time at which the *test* was created (in UTC).
 * `"created_at"`: A [*timestamp*][Timestamp]. The time at which the *test* was created
   (in UTC).
--- a/docs/API.md
+++ b/docs/API.md
@@ -987,9 +987,9 @@ An object with the following properties:
 #### `"result"`
 
 * `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
-  The time at which the *test* was created (in UTC).
-* `"created_at"`: A [*timestamp*][Timestamp]. The time at which the *test* was created
-  (in UTC).
+  The time in UTC at which the *test* was created.
+* `"created_at"`: A [*timestamp*][Timestamp]. The time in UTC at which the *test*
+  was created.
 * `"id"`: **Deprecated** (planned removal: v2022.2). An integer.
 * `"hash_id"`: A [*test id*][Test id]. The *test* in question.
 * `"params"`: See below.
@@ -1099,9 +1099,9 @@ An object with the following properties:
 
 * `"id"` A *test id*.
 * `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
-  The time at which the *test* was created (in UTC).
-* `"created_at"`: A [*timestamp*][Timestamp]. The time at which the *test* was created
-  (in UTC).
+  The time in UTC at which the *test* was created.
+* `"created_at"`: A [*timestamp*][Timestamp]. The time in UTC at which the *test*
+  was created.
 * `"overall_result"`: A string. It reflects the most severe problem level among
   the test results for the test. It has one of the following values:
   * `"ok"`, if there are only messages with [*severity level*][Severity level] `"INFO"` or
--- a/docs/API.md
+++ b/docs/API.md
@@ -397,7 +397,7 @@ This key is added when the module name is `"NAMESERVER"`.
 
 Basic data type: string
 
-**Deprecated representation**
+**Deprecated representation** (planned removal: v2023.1).
 Default database timestamp format: "Y-M-D H:M:S.ms".
 Example: "2017-12-18 07:56:17.156939"
 
@@ -986,7 +986,7 @@ An object with the following properties:
 
 #### `"result"`
 
-* `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
+* `"creation_time"`: **Deprecated** (planned removal: v2023.1). A [*deprecated timestamp*][Timestamp deprecated].
   The time in UTC at which the *test* was created.
 * `"created_at"`: A [*timestamp*][Timestamp]. The time in UTC at which the *test*
   was created.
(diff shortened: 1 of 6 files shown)


matsduf:
> (...) will not display the test creation time because of [this line](https://github.com/zonemaster/zoneult/result.component.ts#L132) (...)

The link is broken.

matsduf:
> If we agree on the value of such a change, in which release should this be done (because of a potential breaking change)? (I've set it to v2022.1 for now.)

I suggest v2023.1. We should give some time, and this update is not urgent, as far as I can see.

matsduf (commented):
It looks fine.

mattias-p:
I've been toying with the idea of renaming `creation_time` to `created_at`. This seems to be a widely used naming convention. If we deprecate `creation_time` and leave it with the old format and introduce `created_at` with the new format, we could have it included in 2022.1 without it being a breaking change.

In case we want to stick with the `creation_time` name we could update the documentation for 2022.1 to state that the current format is deprecated and it will be replaced with the new format in 2022.2 or 2023.1 and that clients are adviced to support both formats until they can fully commit to the newer versions of Zonemaster.

matsduf:
> I've been toying with the idea of renaming `creation_time` to `created_at`. This seems to be a widely used naming convention. If we deprecate `creation_time` and leave it with the old format and introduce `created_at` with the new format, we could have it included in 2022.1 without it being a breaking change.

I think that could be fine. Then other clients can have a controlled migration.

> In case we want to stick with the `creation_time` name we could update the documentation for 2022.1 to state that the current format is deprecated and it will be replaced with the new format in 2022.2 or 2023.1 and that clients are adviced to support both formats until they can fully commit to the newer versions of Zonemaster.

Why should clients support both formats? Isn't enough for clients to shift from "creation_time" to "creation_at" with the new datetime format when that is available?

mattias-p:
> Why should clients support both formats? Isn't enough for clients to shift from "creation_time" to "creation_at" with the new datetime format when that is available?

It helps them decouple their release cycle from ours. This is critical if they don't control the deployment of the RPCAPI server they're using. E.g. if they're using zonemaster.net.

matsduf:
> > Why should clients support both formats? Isn't enough for clients to shift from "creation_time" to "creation_at" with the new datetime format when that is available?
> 
> It helps them decouple their release cycle from ours. This is critical if they don't control the deployment of the RPCAPI server they're using. E.g. if they're using zonemaster.net.

I still do not understand. If the developer of the client waits for the new field to be in place before doing the update, why should she add support for both instead of just switching? The client must of course accept the presence or absence of the old field during the transition, but not use it.

pnax:
> The link is broken.

Fixed.

> I've been toying with the idea of renaming creation_time to created_at. This seems to be a widely used naming convention. If we deprecate creation_time and leave it with the old format and introduce created_at with the new format, we could have it included in 2022.1 without it being a breaking change.

I like the new name (and I'm even wondering about updating the database columns accordingly: `created_at`, `started_at`, `ended_at`).
I'll provide an update in this PR with the new `created_at` field in the result and deprecate the `creation_time` field.

matsduf:
> I like the new name (and I'm even wondering about updating the database columns accordingly: `created_at`, `started_at`, `ended_at`).

I have no objection to that. On the contrary, it makes it clearer if column name and field in API are the same, when possible.

pnax:
While rebasing the PR, I've made the following updates:
* I've updated the PR to keep the way we represent times in the `creation_time` field.
* As suggested I've added a new field `created_at` that uses the ISO 8601 time and date representation.
* I set the `creation_time` field as deprecated in the documentation.

docs/API.md
```diff
@@ -395,9 +395,19 @@ This key is added when the module name is `"NAMESERVER"`.
...
 Basic data type: string
 
+#### Deprecated representation
+
+**Deprecated**
 Default database timestamp format: "Y-M-D H:M:S.ms".
 Example: "2017-12-18 07:56:17.156939"
 
+
+#### New representation
```
matsduf, line 405: I think "Prescribed representation" or "Recommended representation" would be better. Another possibility:

```
### Deprecated Timestamp

Basic data type: string

**Deprecated**
Default database timestamp format: "Y-M-D H:M:S.ms".
Example: "2017-12-18 07:56:17.156939"

### Timestamp

Basic data type: string

A string representing a date and time using the following ISO 8601 format:
"YYYY-MM-DDThh:mm:ssZ".
Example: "2017-12-18T07:56:17Z"
```

docs/API.md
```diff
@@ -981,7 +992,8 @@ An object with the following properties:
 
 #### `"result"`
 
-* `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
+* `"creation_time"`: **Deprecated**. A *timestamp*. The time at which the *test* was enqueued (in UTC).
+* `"created_at"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
```
matsduf, line 996: The key says "created" but the documentation says "enqueued". I think it is unnecessary to have that confusion. Both could work, but "created" has been selected in most places.

If both has *timestamp*, is it just a matter of key name change? No, there is more to it.
```
* `"creation_time"`: **Deprecated**. A *timestamp* in deprecated representation. The time in UTC at which the *test* was created.
* `"created_at"`: A *timestamp* in prescribed representation. The time at in UTC which the *test* was created.
```

Or with my alternative when "timestamp" is defined:

```
* `"creation_time"`: **Deprecated**. A *deprecated timestamp*. The time in UTC at which the *test* was created.
* `"created_at"`: A *timestamp*. The time at in UTC which the *test* was created.
```

docs/API.md
```diff
@@ -1086,7 +1100,8 @@ The value of "frontend_params" is an object with the following properties:
 An object with the following properties:
 
 * `"id"` A *test id*.
-* `"creation_time"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
+* `"creation_time"`: **Deprecated**. A *timestamp*. The time at which the *test* was enqueued (in UTC).
+* `"created_at"`: A *timestamp*. The time at which the *test* was enqueued (in UTC).
```
matsduf, line 1104: See above.

matsduf (changes requested):
I think this looks fine, but I have some commends on the API documentation.

docs/API.md
pnax: I went with your second suggestion.

docs/API.md
pnax: > No, there is more to it.

Good point. I applied your suggested changes.

pnax:
* rebased to fix conflict
* set a removal milestone for deprecated `creation_time`

docs/API.md
```diff
@@ -974,9 +986,11 @@ An object with the following properties:
 
 #### `"result"`
 
-* `"creation_time"`: A [*timestamp*][Timestamp]. The time at which the *test*
-   was enqueued (in UTC).
-* `"id"`: **Deprecated** (will be removed in v2022.2). An integer.
+* `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
+  The time at which the *test* was created (in UTC).
+* `"created_at"`: A [*timestamp*][Timestamp]. The time at which the *test* was created
+  (in UTC).
```
matsduf, line 992: To have less parentheses and also to move UTC to its natural place "time" I suggest:
```
The time in UTC at which the *test* was created.
```
2 places above and 2 places below.

docs/API.md
```diff
@@ -974,9 +986,11 @@ An object with the following properties:
 
 #### `"result"`
 
-* `"creation_time"`: A [*timestamp*][Timestamp]. The time at which the *test*
-   was enqueued (in UTC).
-* `"id"`: **Deprecated** (will be removed in v2022.2). An integer.
+* `"creation_time"`: **Deprecated** (planned removal: v2022.2). A [*deprecated timestamp*][Timestamp deprecated].
```
matsduf, line 989: I see no good reason to remove the old field already at the v2022.2 release. My suggestion is still that we should wait for the v2023.1 release.

matsduf (changes requested):
This looks good. I have two comments, of which one is a major one. I think it will be too early to remove the old field in the v2022.2 release. I see no good reason not to wait for the v2023.1 release.

docs/API.md
```diff
@@ -392,13 +393,23 @@ Sometimes additional keys are present.
 This key is added when the module name is `"NAMESERVER"`.
 
 
-### Timestamp
+### Timestamp (deprecated)
 
 Basic data type: string
 
+**Deprecated representation** (planned removal: v2023.1).
```
mattias-p, line 400: Maybe add an empty line after this to make the rendering nicer?

mattias-p (dismissed):
LGTM

docs/API.md
matsduf: I agree.

matsduf (dismissed):
I approve.

docs/API.md
pnax: added

matsduf (approved):
I approve.

mattias-p:
I successfully tested this for v2022.1.

Issue #951, opened 2020-12-15
Zone names from classless delegations (RFC 2317) not accepted


stirnim:
The zonemaster.net UI does not accept zone names from classless delegation (RFC 2317).

Web UI error:
Internal server error

Zone name used for testing:
192/26.26.176.195.in-addr.arpa.

matsduf:
Resolved by #969 and is now available in the v2022.1 release.

vcunat:
We still see errors with the latest zonemaster versions, e.g.:
https://zonemaster.nic.cz/en/result/b5c9c145622d2d3a
https://zonemaster.nic.cz/en/result/8fa76fed8b144f09

Are we missing something?

matsduf:
The second link shows a test of a non-zone. That is why it returns an error (CRITICAL).

The second link shows complaint about "/" in a zone name. Before the change in #969 the zone was not tested at all if it contained a "/". Now it does. But [Syntax01] still complaints about "/". That test case needs an update, especially to produce better messages. I will create an issue on that.

[Syntax01]: https://doc.zonemaster.net/latest/specifications/tests/Syntax-TP/syntax01.html

PR #969, opened 2022-04-04
Makes solidus valid character in ASCII labels

matsduf:
## Purpose

Besides usual a-z, 0-9 and "-" also underscore "_" and solidus "/" should be permitted in ASCII labels. The two latter are used, but in limited cases and not as host names. There is no security issue to permit them. Also see [draft updated Basic00] and  [draft Basic00 implementation].

This PR:
* Adds "/" as permitted character in ASCII labels.
* Adds missing "_" in one regex expression for the domain name (there are two).

## Context

Resolves issue #951.

## How to test this PR

Enter 192/26.26.176.195.in-addr.arpa to RPC-API via GUI, `zmb` or `zmtest`.

[draft updated Basic00]: https://github.com/zonemaster/zonemaster/blob/9e043638b939022429a45be5adc2a2d7fcd73f91/docs/specifications/tests/Basic-TP/basic00.md#ascii-domain-name

[draft Basic00 implementation]: https://github.com/zonemaster/zonemaster-engine/pull/1040


lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -268,7 +268,7 @@ sub _check_domain {
         }
     }
 
-    if ( $domain !~ m/^[\-a-zA-Z0-9\.\_]+$/ ) {
+    if ( $domain !~ m{^[a-z0-9_./-]+$}i ) {
```
mattias-p, line 271: Have you considered calling untaint_ldh_domain() here instead?

matsduf: No, my approach was to make minimal update, and I did not even think of it before you mentioned it.

&untaint_ldh_domain() also tests the length of the domain namn, which I guess should not be tested here. That should rather be handled by test case code. Independent, that would go beyond the purpose of this PR.

blacksponge (approved):
Looks go to me.

pnax (approved):

PR #970, opened 2022-04-07
Database update: new UNIQUE constraints + rename some columns

ghost:
## Purpose

This adds a constraint at the database level to avoid duplicate items such as `hash_id` and `username`.

Note: I haven't address the case where a row with the same `hash_id` as one already in database is inserted. The constraint will raise an error. But there is currently no logic to catch this error. This is left for a future work. The idea is first to make sure we can't have duplicated `hash_id` in database. 

## Context

For the UNIQUE constraint, see https://github.com/zonemaster/zonemaster-backend/pull/965#discussion_r838285362
For the column renaming, see https://github.com/zonemaster/zonemaster-backend/pull/967#issuecomment-1090071803

## Changes

* Add UNIQUE constraints to `hash_id` column in "test_results" table and `username` column in "users" table
* Rename columns `creation_time`, `test_start_time`, `test_end_time` to `created_at`, `started_at` and `ended_at`
* Adapt the upgrade script
* Refactor the `add_api_user` method

## How to test this PR

* The upgrade script should work, and the database should be properly updated: new columns' name + UNIQUE constraint
* Trying to add twice the same user should result in an error

### DIFF
--- a/lib/Zonemaster/Backend/DB/SQLite.pm
+++ b/lib/Zonemaster/Backend/DB/SQLite.pm
@@ -9,6 +9,7 @@ use DBI qw(:utils :sql_types);
 use Digest::MD5 qw(md5_hex);
 use JSON::PP;
 
+
 use Zonemaster::Backend::Errors;
 
 with 'Zonemaster::Backend::DB';
@@ -46,7 +47,7 @@ sub DEMOLISH {
 }
 
 sub get_dbh_specific_attributes {
-    return {};
+    return { sqlite_extended_result_codes => 1 };
 }
 
 sub create_schema {
@@ -293,7 +294,7 @@ sub is_duplicate {
     my ( $self ) = @_;
 
     # for the list of codes see: https://sqlite.org/rescode.html
-    return ( $self->dbh->err == 19 );
+    return ( $self->dbh->err == 2067 );
 }
 
 no Moose;


lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -67,16 +67,18 @@ sub create_schema {
...
+            ended_at DATETIME DEFAULT NULL,
             priority integer DEFAULT 10,
             queue integer DEFAULT 0,
             progress integer DEFAULT 0,
             fingerprint character varying(32),
             params blob NOT NULL,
             results mediumblob DEFAULT NULL,
-            undelegated integer NOT NULL DEFAULT 0
+            undelegated integer NOT NULL DEFAULT 0,
+
```
matsduf, line 80: Any reason for an empty line?

matsduf (commented):
I think that the next version of Zonemaster-Backend should be v9.0.0, which affects the name of the patch file.

matsduf:
> Note: I haven't address the case where a row with the same `hash_id` as one already in database is inserted. The constraint will raise an error. But there is currently no logic to catch this error. This is left for a future work. The idea is first to make sure we can't have duplicated `hash_id` in database.

If it is worth to add the constraint, why not add some logic that will catch such an error? What would the error be with the code in the PR?

pnax:
> If it is worth to add the constraint, why not add some logic that will catch such an error?

I think it is worth to add the constraint because so far there was no guardrail to prevent `hash_id` collision in database anymore (I think such guardrail was present before for some databases but was removed recently). So we need such protection back.

About the logic to catch the error, as said in the description I chose to let if for a future work because I'm not sure how to deal with such case. Should the insertion be tried again with another `hash_id` value? Or should it only fail? Or anything else?

> What would the error be with the code in the PR?

The insertion will fail, and you would receive the following message:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32603,
    "data": null,
  },
  "id": 2
}
```

With the following line in the log (for SQLite):

```
2022-04-11T09:16:03Z [968597] ERROR - Caught Zonemaster::Backend::Error::Internal in the `Zonemaster::Backend::RPCAPI::start_domain_test` method: DBD::SQLite::db do failed: UNIQUE constraint failed: test_results.hash_id at /zonemaster/zonemaster-backend/lib/Zonemaster/Backend/DB.pm line 182.
```

matsduf:
> (I think such guardrail was present before for some databases but was removed recently).

I do not think there ever was such check. When I updated the support for SQLite (to make it work) some year or two back in time, there was not such check for any database engine. In my first draft I added a check (less sophisticated than your implementation) but I removed it again before it was merged.

I think that the idea was that collision was so unlikely that we could live with the risk.

I do not think that an "internal error" is a nice response, but if we can still assume that the risk of collision is very small then I think we can live with that.

Besides the complexity in the code and maybe the extra load on the database engines I see no harm with this PR.

matsduf (dismissed):

pnax:
Rebased to fix the conflicts.

> I do not think there ever was such check.

I looked where I think it was but couldn't find anything. There might never be such check indeed.

> I do not think that an "internal error" is a nice response

Agreed, I'd like to provide something better but couldn't come with something nice at this point.

> but if we can still assume that the risk of collision is very small then I think we can live with that.

Nothing has changed, so the risk is still the same. But now there is a guardrail if a collision occurs.

mattias-p:
I found a problem during release testing and I created #1030 to fix it.

mattias-p:
After the fix in #1030 I successfully tested this for v2022.1.

PR #971, opened 2022-04-12
Fix crash when calling "get_test_results"

pnax:
## Purpose

The way the hash is passed in the Translator and how the Logger is implemented will alter the data in the origin hash resulting in an error when calling `get_test_results`:

```
$ zmb get_test_results --test-id ccaaf2e5df7a26f6 --language  | jq
{
  "error": {
    "data": null,
    "code": -32603,
  },
  "id": 1,
  "jsonrpc": "2.0"
}
```

## Context

n/a

## Changes

Fix Translator.pm

## How to test this PR

1. create a test, get its id
2. call `get_test_results`: `zmb get_test_results --test-id ccaaf2e5df7a26f6 --language  | jq`
3. there should be results and not an "Internal server error" message

### DIFF
--- a/lib/Zonemaster/Backend/Translator.pm
+++ b/lib/Zonemaster/Backend/Translator.pm
@@ -18,7 +18,7 @@ extends 'Zonemaster::Engine::Translator';
 sub translate_tag {
     my ( $self, $hashref ) = @_;
 
-    my $entry = Zonemaster::Engine::Logger::Entry->new( %{ $hashref } );
+    my $entry = Zonemaster::Engine::Logger::Entry->new( { %{ $hashref } } );
     my $octets = Zonemaster::Engine::Translator::translate_tag( $self, $entry );
 
     return decode_utf8( $octets );


pnax:
> Is the PR description correct?

Indeed there is an error, it should be `--lang` with an argument as stated by the error message you pasted in your comment. I updated the description.

blacksponge (approved):
Working as expected.

matsduf:
Again, when do you see the error? Everytime? Sometimes? Specific situation?

```
zmb get_test_results --test-id 6e48a3d1d3836cb9 --lang en | jq . 
{
  "result": {
    "results": [
      {
        "level": "INFO",
        "module": "SYSTEM",
        "message": "Using version v4.4.0 of the Zonemaster engine.\n"
      },
      {
        "message": "Parent domain 'se' was found for the tested domain.\n",
        "level": "INFO",
        "module": "BASIC"
      },
      {
        "module": "BASIC",
        "level": "WARNING",
        "message": "Nameserver nsa.dnsnode.net/2a01:3f1:46::53 does not respond over neither UDP nor TCP.\n"
      },
(...)
}
```

blacksponge:
The error is present on develop

matsduf (approved):

lib/Zonemaster/Backend/Translator.pm
```diff
@@ -18,7 +18,7 @@ extends 'Zonemaster::Engine::Translator';
 sub translate_tag {
     my ( $self, $hashref ) = @_;
 
-    my $entry = Zonemaster::Engine::Logger::Entry->new( %{ $hashref } );
+    my $entry = Zonemaster::Engine::Logger::Entry->new( { %{ $hashref } } );
```
mattias-p, line 21: Wouldn't this work too?

```suggestion
    my $entry = Zonemaster::Engine::Logger::Entry->new( $hashref );
```

mattias-p (approved):

lib/Zonemaster/Backend/Translator.pm
pnax: From what I recall, this won't work because of a deletion of keys in Zonemaster::Engine::Logger::Entry that would affect this hashref, so we need to pass a copy.

https://github.com/zonemaster/zonemaster-engine/blob/342a37aff272a7813643efe270a7c4fa94ed76d4/lib/Zonemaster/Engine/Logger/Entry.pm#L56-L58

mattias-p: Oh. But in that case maybe the fix should be to change the constructor rather than the call site?

PR #972, opened 2022-04-14
Normalize profile name in DB

blacksponge:
## Purpose

* Make profile case insensitive when computing the fingerprint.
* Normalize profile name before storing it.

## Context

\-

## Changes

*Lower case the profile name when projecting the params.

## How to test this PR

The following commands should return the same test id when executed
* `./script/zmb start_domain_test --domain 'zonemaster.net' --lang en --profile DEFAULT`
* `./script/zmb start_domain_test --domain 'zonemaster.net' --lang en --profile default`

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -533,10 +533,10 @@ sub _project_params {
 
     my %projection = ();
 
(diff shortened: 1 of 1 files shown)


pnax (approved):
Good catch. LGTM.

matsduf (approved):
Profile names are case insensitive, so this is good. I approve.

Issue #737, opened 2021-04-12
Documentation of environment variables


mattias-p:
Various programs in Backend are configurable using various environment variables. Currently these aren't well documented. For each command there should be a complete list of the environment variables it uses. The set of all unit tests should probably be considered a single command in this respect. The list should specify:
* The name of the variable
* The set of valid values (syntax)
* The meaning of each valid value
* The meaning of not setting the variable (e.g. the default value)
* Any interactions with other configuration parameters

matsduf:
Good suggestion.

Could we extend the scope of [Configuration.md] and include it there? That document is not too large, and I feel it is better to keep all configuration there. Setting an environment variable is a sort of configuration. And `ZONEMASTER_BACKEND_CONFIG_FILE` is one such variable that points out the configuration file.

[Configuration.md]: https://github.com/zonemaster/zonemaster-backend/blob/master/docs/Configuration.md

mattias-p:
Another option to consider is to have separate documentations for each command. We do have a couple of them and they aren't all configurable using the same environment variables.

matsduf:
Resolved by #974 and #993

PR #974, opened 2022-04-19
Improve env doc

blacksponge:
## Purpose

Add missing environment variable in documentation.

## Context
https://github.com/zonemaster/zonemaster-backend/pull/966#issuecomment-1102217575
https://github.com/zonemaster/zonemaster-backend/issues/737

## Changes
Add documentation for `ZONEMASTER_BACKEND_CONFIG_FILE`

## How to test this PR
\-

### DIFF
--- a/docs/EnvironmentVariables.md
+++ b/docs/EnvironmentVariables.md
@@ -1,7 +1,10 @@
 # Environment variables
 
-The PSGI server (`zonemaster_backend_rpcapi.psgi`) support the following
-environment variables.
+* `ZONEMASTER_BACKEND_CONFIG_FILE`: Set a custom path for the backend
+  configuration file.
+
+The following environement variables are only supported by the PSGI server
+(`zonemaster_backend_rpcapi.psgi`).
 
 * `ZM_BACKEND_RPCAPI_LOGLEVEL`: Configure the log level, `trace` by default.
   Accepted values are:
--- a/docs/EnvironmentVariables.md
+++ b/docs/EnvironmentVariables.md
@@ -1,10 +1,11 @@
 # Environment variables
 
+## Variable used by both RPCAPI daemon and Test Agent daemon
+
 * `ZONEMASTER_BACKEND_CONFIG_FILE`: Set a custom path for the backend
   configuration file.
 
-The following environement variables are only supported by the PSGI server
-(`zonemaster_backend_rpcapi.psgi`).
+## Variables used by RPCAPI daemon only
 
 * `ZM_BACKEND_RPCAPI_LOGLEVEL`: Configure the log level, `trace` by default.
   Accepted values are:


docs/EnvironmentVariables.md
```diff
@@ -1,7 +1,10 @@
 # Environment variables
 
-The PSGI server (`zonemaster_backend_rpcapi.psgi`) support the following
-environment variables.
+* `ZONEMASTER_BACKEND_CONFIG_FILE`: Set a custom path for the backend
+  configuration file.
+
+The following environement variables are only supported by the PSGI server
```
mattias-p, line 6: In other places we're referring to "the RPCAPI daemon" or just "the RPCAPI". Maybe use that here as well?

matsduf: I agree, but instead I suggest:

```
## Variables used by RPCAPI daemon only
```

docs/EnvironmentVariables.md
```diff
@@ -1,7 +1,10 @@
 # Environment variables
```
matsduf, line 2: I suggest that the following is inserted:

```
## Variable used by both RPCAPI daemon and Test Agent daemon
```

(We should be consistent if it should be "Test agent", "TestAgent" or "Test Agent". It differs in different documents.)

docs/EnvironmentVariables.md
pnax, line 6: typo: "environement" -> "environment"

pnax (approved):
Nice to have such documentation.

This also addresses #737

docs/EnvironmentVariables.md
blacksponge: done

docs/EnvironmentVariables.md
blacksponge: done

pnax (approved):

matsduf (approved):
I approve this, but I am thinking of the title of the document. These environment variables are for Backend only, and I think it might be good to let the title of the document reflect that. I let you decide.

* Backend Environment Variables
* Zonemaster-Backend Environment Variables

Else this is a small but great addition, and now we should make sure to add references to it.

blacksponge:
> I approve this, but I am thinking of the title of the document. These environment variables are for Backend only, and I think it might be good to let the title of the document reflect that. I let you decide.

I used "Environment variables" as the configuration title is "Configuration", with no reference that it is backend only, the same goes for the other documents. I guess the fact that it lives in the backend repository already make that clear. I would be in favor to keep it to "Environment variables" to be coherent with the rest of the documentation.

matsduf:
> I used "Environment variables" as the configuration title is "Configuration", with no reference that it is backend only, the same goes for the other documents. I guess the fact that it lives in the backend repository already make that clear. I would be in favor to keep it to "Environment variables" to be coherent with the rest of the documentation.

I am fine with that. There are pros and cons with different names.

matsduf:
@blacksponge, please merge.

PR #977, opened 2022-04-19
Store index using BIGINT in database

pnax:
## Purpose

Currently we store the index using an INTEGER in database. When using batches there is a risk we reach the upper limit.

## Context

n/a

## Changes

* update database schema for PostgreSQL and MySQL
* update the database upgrade script

## How to test this PR

Run the upgrade script and check that the `id` field has the right data type

### DIFF
--- a/lib/Zonemaster/Backend/DB/MySQL.pm
+++ b/lib/Zonemaster/Backend/DB/MySQL.pm
@@ -63,7 +63,7 @@ sub create_schema {
     ####################################################################
     $dbh->do(
         'CREATE TABLE IF NOT EXISTS test_results (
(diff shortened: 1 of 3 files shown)


mattias-p (approved):
LGTM

blacksponge (approved):

matsduf:
@pnax, what is the reason that SQLite was not updated?

pnax:
Because SQLite handles it automatically based on the need.

> INTEGER. The value is a signed integer, stored in 0, 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value.
https://www.sqlite.org/datatype3.html

mattias-p:
I successfully tested this for v2022.1.

PR #980, opened 2022-04-26
Set batch creation time using Perl time

ghost:
## Purpose

Time is not set by default by the database engine on insertion. It should be passed manually on insertion.
As with #962, it seems we missed to update the batch job creation as well.

## Context

Follow-up on https://github.com/zonemaster/zonemaster-backend/pull/769

## Changes

* Update "create_new_batch_job" in DB.pm.
* Fix an error in the upgrade script for MySQL database, the "created_at" column should not be nullable.

## How to test this PR

Create a batch job and verify that the "created_at" contains the current date.
Check that after running the upgrade script using MySQL, the "created_at" column from "batch_jobs" table is not null.


matsduf (approved):
I approve.

PR #981, opened 2022-04-26
Deduplicate database call

pnax:
## Purpose

* Avoid code duplication.
* Avoid extra call to database

## Context

n/a

## Changes

Update "DB.pm" file

## How to test this PR

Internal change only, nothing should break.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -381,12 +381,12 @@ sub get_test_request {
     my $result_id;
     my $dbh = $self->dbh;
 
-    my ( $id, $hash_id, $batch_id );
+    my ( $hash_id, $batch_id );
     if ( defined $queue_label ) {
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/Backend/DB.pm
```diff
@@ -242,6 +242,8 @@ sub test_progress {
                 $test_id,
             );
         }
+
+        return $progress;
```
matsduf, line 246: This is a new return value. What call will use that? Will calling code not be affected?

pnax: Nope it's not, the function already returns a value (3 lines down). Here we return earlier to avoid an extra call to the database that fetch the current progress value.

matsduf: Thanks for clarification.

matsduf (approved):
I approve.

Issue #952, opened 2022-02-15
Improve code coverage in unit tests


ghost:
The Backend lacks proper automated tests covering its code.

Some issues relate to this lack of unit tests (or asking for feature regarding unit tests):

* #376 
* #546 
* #747 
* #758 
* #761 
* #806 
* #827 

And currently open PRs (as time of creating this issue):

* #769 
* #947

PR #982, opened 2022-04-27
Testing batch feature

pnax:
## Purpose

This is a first step towards better test coverage for batch feature.

## Context

#952 

## Changes

new `t/batches.t` file

## How to test this PR

tests should pass

### DIFF
--- a/t/batches.t
+++ b/t/batches.t
@@ -12,17 +12,12 @@ use Test::Exception;
 use Test::More;    # see done_testing()
 
 use Zonemaster::Engine;
+use Zonemaster::Backend::Config;
 
 # Use the TARGET environment variable to set the database to use
 # default to SQLite
-my $db_backend = $ENV{TARGET};
-if ( not $db_backend ) {
-    $db_backend = 'SQLite';
-} elsif ( $db_backend !~ /^(?:SQLite|MySQL|PostgreSQL)$/ ) {
-    BAIL_OUT( "Unsupported database backend: $db_backend" );
-}
-
-diag "database: $db_backend";
+my $db_backend = Zonemaster::Backend::Config->check_db( $ENV{TARGET} || 'SQLite' );
+note "database: $db_backend";
 
 my $tempdir = tempdir( CLEANUP => 1 );
 
--- a/t/batches.t
+++ b/t/batches.t
@@ -1,0 +1,323 @@
+use strict;
+use warnings;
+use 5.14.2;
+
+use Cwd;
+use Data::Dumper;
+use File::Temp qw[tempdir];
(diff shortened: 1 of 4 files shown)


t/batches.t
```diff
@@ -0,0 +1,335 @@
...
+use File::Temp qw[tempdir];
+use JSON::PP;
+use POSIX qw( strftime );
+use Time::Local qw( timelocal_modern );
+use Test::Exception;
+use Test::More;    # see done_testing()
+
+use Zonemaster::Engine;
+
+# Use the TARGET environment variable to set the database to use
```
matsduf, line 16: The TARGET environment variable is used for Travis testing. Maybe should be noted here.

t/batches.t
```diff
@@ -0,0 +1,335 @@
...
+    username => 'user',
+    api_key => 'key'
+};
+
+# define the default properties for the tests
+my $params = {
+    client_id      => 'Unit Test',
+    client_version => '1.0',
+    ipv4           => JSON::PP::true,
+    ipv6           => JSON::PP::true,
```
matsduf, line 82: I might be of value to be able to run with at least IPv6 turned off with an environment variable.

matsduf (commented):
It is great with unit tests.

t/batches.t
pnax: Good idea (and same for IPv4). However this is outside the scope of this PR.

t/batches.t
pnax: This variable is not only used by Travis. For instance I use it when running the tests locally.

t/batches.t
matsduf: It depends on how this test will work in a non-IPv6 environment. It will always be run when updated Zonemaster-Backend is installed, won't it? What happens then in a non-IPv6 environment?

t/batches.t
matsduf: Still, it is assumed by Travis, and that should be noted here.

t/batches.t
pnax: It won't behave any different than currently with [t/test01.t](https://github.com/zonemaster/zonemaster-backend/blob/develop/t/test01.t#L120) or [t/idn.t](https://github.com/zonemaster/zonemaster-backend/blob/develop/t/idn.t#L82).

t/batches.t
pnax: I chose to document this variable in the new EnvironmentVariable.md document, see #993.

matsduf: That is a good choice.

matsduf (approved):

PR #984, opened 2022-04-28
Use libidn2 instead of libidn

pnax:
## Purpose

Replace libidn with libidn2.

## Context

https://github.com/zonemaster/zonemaster-ldns/pull/133

## Changes

* Update Travis dependency
* Replace references to "libidn" with "libidn2"

## How to test this PR

Using Zonemaster::LDNS (with libidn2 support), Zonemaster should work the same as before.

### DIFF
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -232,7 +232,7 @@ sub to_idn {
         return Zonemaster::LDNS::to_idn( $str );
     }
     else {
-        warn __( "Warning: Zonemaster::LDNS not compiled with libidn, cannot handle non-ASCII names correctly." );
+        warn __( "Warning: Zonemaster::LDNS not compiled with libidn2, cannot handle non-ASCII names correctly." );
         return $str;
     }
 }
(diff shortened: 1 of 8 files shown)


blacksponge (approved):

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -232,7 +232,7 @@ sub to_idn {
         return Zonemaster::LDNS::to_idn( $str );
     }
     else {
-        warn __( "Warning: Zonemaster::LDNS not compiled with libidn, cannot handle non-ASCII names correctly." );
+        warn __( "Warning: Zonemaster::LDNS not compiled with libidn2, cannot handle non-ASCII names correctly." );
```
matsduf, line 235: I think the following would be better and not bound to details not relevant here.
```
Warning: Zonemaster::LDNS not compiled with IDN support, cannot handle names with non-ASCII characters correctly.
```
But I accept as-is.

Compare with https://github.com/zonemaster/zonemaster-cli/pull/254#discussion_r860800455

matsduf (approved):

lib/Zonemaster/Backend/TestAgent.pm
pnax: Good point. I'll do the change in another PR.

PR #988, opened 2022-04-28
Simplify test progress' computation

pnax:
## Purpose

The progress computation relies on parsing the logger trace which is quite heavy work. It appears that the searched values are available in the entry directly when receiving "TEST_CASE_END" message (at least).

## Context

n/a

## Changes

* Fix the total number of planned methods (check that each method is part of the current profile)
* Update TestAgent.pm and the progress computation
* Provide some small refactoring (see the commits' history to follow the changes)

## How to test this PR

Tests should pass.

## Note

I've set this PR to next release, I think this is a nice improvement that removes some heavy code and computation power. However this is totally fine to move it to v2022.2 if it appears to be too short to have a proper review process.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -209,6 +209,16 @@ sub recent_test_hash_id {
     return $recent_hash_id;
 }
 
+=head2 test_progress($test_id, $progress)
+
+Takes a test_id and returns the current progress value for the associated test.
+If progress is set, update the progress value of the test.
+If the progress value is 1, set the C<started_at> field to the current time in UTC.
+
+The progress value is an integer in the range 0-100.
+
+=cut
+
 sub test_progress {
     my ( $self, $test_id, $progress ) = @_;
 
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -215,7 +215,7 @@ Takes a test_id and returns the current progress value for the associated test.
 If progress is set, update the progress value of the test.
 If the progress value is 1, set the C<started_at> field to the current time in UTC.
 
-The progress value is an integer in the range 0-100.
+The C<$progress> value is an integer in the range 1-99.
 
 =cut
 
@@ -224,6 +224,8 @@ sub test_progress {
 
     my $dbh = $self->dbh;
     if ( $progress ) {
+        $progress = $progress < 1 ? 1 :
+                    $progress > 99 ? 99 : $progress;
         if ( $progress == 1 ) {
             $dbh->do(
                 q[
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -122,7 +122,7 @@ sub run {
                 if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
                     $nbr_testcases_finished++;
                     # limit to max 99%, 100% is reached when data is stored in database
-                    my $progress_percent = 99 * $nbr_testcases_finished /  $nbr_testcases_planned;
+                    my $progress_percent = int( 99 * $nbr_testcases_finished /  $nbr_testcases_planned );
                     $self->{_db}->test_progress( $test_id, $progress_percent );
                 }
             }
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -101,22 +101,21 @@ sub run {
         Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, ( $params->{ipv6} ) ? ( 1 ) : ( 0 ) );
     }
 
-    my %methods = Zonemaster::Engine->all_methods;
+    if ( $show_progress ) {
+        my %methods = Zonemaster::Engine->all_methods;
 
-    # BASIC methods are always run: Basic0{0..4}
-    my $nbr_testcases_planned = 5;
-    my $nbr_testcases_finished = 0;
+        # BASIC methods are always run: Basic0{0..4}
+        my $nbr_testcases_planned = 5;
+        my $nbr_testcases_finished = 0;
 
-    foreach my $module ( keys %methods ) {
-        foreach my $method ( @{ $methods{$module} } ) {
-            if ( Zonemaster::Engine::Util::should_run_test( $method ) ) {
-                $nbr_testcases_planned++;
+        foreach my $module ( keys %methods ) {
+            foreach my $method ( @{ $methods{$module} } ) {
+                if ( Zonemaster::Engine::Util::should_run_test( $method ) ) {
+                    $nbr_testcases_planned++;
+                }
             }
         }
-    }
 
-
-    if ( $show_progress ) {
         Zonemaster::Engine->logger->callback(
             sub {
                 my ( $entry ) = @_;
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -113,6 +113,7 @@ sub run {
         }
     }
 
+    my @planned_methods = sort keys %{ $counter_for_progress_indicator{planned} };
 
     # used for progress indicator
     my ( $previous_module, $previous_method ) = ( '', '' );
@@ -124,7 +125,7 @@ sub run {
                 my ( $entry ) = @_;
 
                 foreach my $trace ( reverse @{ $entry->trace } ) {
-                    foreach my $module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
+                    foreach my $module_method ( @planned_methods ) {
                         if ( index( $trace->[1], $module_method ) > -1 ) {
                             my $percent_progress = 0;
                             my ( $module ) = ( $module_method =~ /(.+::)[^:]+/ );
@@ -132,7 +133,7 @@ sub run {
                                 $counter_for_progress_indicator{executed}{$module_method}++;
                             }
                             elsif ( $previous_module ) {
-                                foreach my $planned_module_method ( keys %{ $counter_for_progress_indicator{planned} } ) {
+                                foreach my $planned_module_method ( @planned_methods ) {
                                     if ( $counter_for_progress_indicator{planned}{$planned_module_method} eq $previous_module ) {
                                         $counter_for_progress_indicator{executed}{$planned_module_method}++;
                                     }
@@ -145,7 +146,7 @@ sub run {
                                     "%.0f",
                                     99 * (
                                         scalar( keys %{ $counter_for_progress_indicator{executed} } ) /
-                                          scalar( keys %{ $counter_for_progress_indicator{planned} } )
+                                          scalar( @planned_methods )
                                     )
                                 );
                                 $self->{_db}->test_progress( $test_id, $percent_progress );
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -61,7 +61,6 @@ sub new {
 sub run {
     my ( $self, $test_id, $show_progress ) = @_;
     my @accumulator;
-    my %counter_for_progress_indicator;
 
     my $params;
 
@@ -104,46 +103,29 @@ sub run {
 
     my %methods = Zonemaster::Engine->all_methods;
 
-    # BASIC methods are always run
-    $counter_for_progress_indicator{planned} = {
-        'Basic::basic00' => 1,
-        'Basic::basic01' => 1,
-        'Basic::basic02' => 1,
-        'Basic::basic03' => 1,
-        'Basic::basic04' => 1,
-    };
+    # BASIC methods are always run: Basic0{0..4}
+    my $nbr_testcases_planned = 5;
+    my $nbr_testcases_finished = 0;
+
     foreach my $module ( keys %methods ) {
         foreach my $method ( @{ $methods{$module} } ) {
             if ( Zonemaster::Engine::Util::should_run_test( $method ) ) {
-                $counter_for_progress_indicator{planned}{ $module . '::' . $method } = 1;
+                $nbr_testcases_planned++;
             }
         }
     }
 
-    my @planned_methods = sort keys %{ $counter_for_progress_indicator{planned} };
-
-    # used for progress indicator
-    my ( $previous_module, $previous_method ) = ( '', '' );
 
     if ( $show_progress ) {
         # Callback defined here so it closes over the setup above.
         Zonemaster::Engine->logger->callback(
             sub {
                 my ( $entry ) = @_;
-
                 if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
-                    my $current_module = ucfirst( lc $entry->{_module} );
-                    my $current_testcase = $entry->{args}->{testcase};
-                    my $current_module_method = $current_module . "::" . $current_testcase;
-
-                    $counter_for_progress_indicator{executed}{$current_module_method}++;
-
+                    $nbr_testcases_finished++;
                     my $progress_percent = sprintf(
                         "%.0f",
-                        99 * (
-                            scalar( keys %{ $counter_for_progress_indicator{executed} } ) /
-                              scalar( @planned_methods )
-                        )
+                        99 * $nbr_testcases_finished /  $nbr_testcases_planned
                     );
 
                     $self->{_db}->test_progress( $test_id, $progress_percent );
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -123,11 +123,7 @@ sub run {
                 my ( $entry ) = @_;
                 if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
                     $nbr_testcases_finished++;
-                    my $progress_percent = sprintf(
-                        "%.0f",
-                        99 * $nbr_testcases_finished /  $nbr_testcases_planned
-                    );
-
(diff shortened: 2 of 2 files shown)


lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +102,56 @@ sub run {
...
 
+    my %methods = Zonemaster::Engine->all_methods;
+
+    # BASIC methods are always run
+    $counter_for_progress_indicator{planned} = {
+        'Basic::basic00' => 1,
+        'Basic::basic01' => 1,
+        'Basic::basic02' => 1,
+        'Basic::basic03' => 1,
+        'Basic::basic04' => 1,
```
matsduf, line 113: Is it good and necessary to refer to specific test cases here? Which test cases that are run even not included in profile can change.

Basic03 is normally not run. There is no "TEST_CASE_END basic03" message and HAS_NAMESERVER_NO_WWW_A_TEST does not come from test case basic03.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +102,56 @@ sub run {
...
+    # used for progress indicator
+    my ( $previous_module, $previous_method ) = ( '', '' );
+
+    if ( $show_progress ) {
+        # Callback defined here so it closes over the setup above.
+        Zonemaster::Engine->logger->callback(
+            sub {
+                my ( $entry ) = @_;
+
+                if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
```
matsduf, line 134: Then it must be ensured that TEST_CASE_END does not change.

matsduf (commented):
There is no reason to let the progress bar consume a lot of resources. It would probably be enough if it increased each time a module (Bascic, Consistency etc) is completed.

pnax:
> There is no reason to let the progress bar consume a lot of resources.

This is exactly what I had in mind when working on this PR. Lift any unnecessary burden when computing test progress.

> It would probably be enough if it increased each time a module (Bascic, Consistency etc) is completed.

The thing is that a whole module is not always run, the profile allows to specify which test case to run. I think this is a good granularity to use. Then `progress = (number of finished test cases) / (number of planned test cases)`.

lib/Zonemaster/Backend/TestAgent.pm
pnax: > Is it good and necessary to refer to specific test cases here?

Since progress is defined as `progress = (nbr finished test cases ) / (nbr planned test cases)`, I need to compute the number of planned test. Getting them from the profile is not enough as you noted.

> Which test cases that are run even not included in profile can change.

True, this is something to keep in mind. Maybe Engine should be updated in a way so that it makes it easier to get the list of planned test cases. But this would be part of a future work.

> Basic03 is normally not run.

True that Basic03 is normally not run. I thought it won't alter the progress computation too much to keep it anyways in the number of planned test. I think it's better that the following inequality always holds: `(nbr planned test cases) >= (nbr finished test cases)`.

> There is no "TEST_CASE_END basic03" message ...

This is not what I found while working on this. "TEST_CASE_END" is properly defined ([master](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/Basic.pm#L126) and [develop](https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Basic.pm#L125)) and used ([master](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/Basic.pm#L523) and [develop](https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Basic.pm#L522)).

> ...  and HAS_NAMESERVER_NO_WWW_A_TEST does not come from test case basic03.

Sorry I don't understand what you mean here. Could you elaborate please?

lib/Zonemaster/Backend/TestAgent.pm
pnax: TEST_CASE_END was requested by https://github.com/zonemaster/zonemaster-engine/issues/665 and added with https://github.com/zonemaster/zonemaster-engine/pull/764.
As with everything from any other components, there might be some change in the future. And we should pay attention to such evolution to not break things in other components.
It would be great if the "TEST_CASE_END" message does not change to much indeed. And so far I don't see why it would. Are you expecting it to change soon? If so could you point to any related issue or PR or discussion about such future change, or explain how you would expect it to change?

lib/Zonemaster/Backend/TestAgent.pm
matsduf: > > There is no "TEST_CASE_END basic03" message ...
> 
> This is not what I found while working on this. "TEST_CASE_END" is properly defined ([master](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/Basic.pm#L126) and [develop](https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Basic.pm#L125)) and used ([master](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine/Test/Basic.pm#L523) and [develop](https://github.com/zonemaster/zonemaster-engine/blob/develop/lib/Zonemaster/Engine/Test/Basic.pm#L522)).

What I meant was that since basic03 is not run in most cases, the "TEST_CASE_END basic03" message is not outputted in those cases. Of course it is outputted if it is run.

> > ...  and HAS_NAMESERVER_NO_WWW_A_TEST does not come from test case basic03.
> 
> Sorry I don't understand what you mean here. Could you elaborate please?

The HAS_NAMESERVER_NO_WWW_A_TEST tag is outputted outside the basic03 code.

lib/Zonemaster/Backend/TestAgent.pm
matsduf: It was just a comment. So far it has only been meant for human reading.

matsduf:
> > There is no reason to let the progress bar consume a lot of resources.
> 
> This is exactly what I had in mind when working on this PR. Lift any unnecessary burden when computing test progress.

I was unclear. I think it is good that you do something about it. It is important to have a progress bar, but it should not eat too much reasources and it does not have to be exact.

> > It would probably be enough if it increased each time a module (Bascic, Consistency etc) is completed.
> 
> The thing is that a whole module is not always run, the profile allows to specify which test case to run. I think this is a good granularity to use. Then `progress = (number of finished test cases) / (number of planned test cases)`.

There are 9 modules, and if each of them adds 10 (10%) to progress it would be good enough. When the test is completed it will jump to 100, which is fine from a user perspective.

It would probably be good enough without regarding the basic test cases always run as long as progress does not go beyond 100, and that could be fixed by not letting it go beyond 100.

lib/Zonemaster/Backend/TestAgent.pm
pnax: > The HAS_NAMESERVER_NO_WWW_A_TEST tag is outputted outside the basic03 code.

The progress computation does not rely on this message. Only on "TEST_CASE_END".

> What I meant was that since basic03 is not run in most cases, the "TEST_CASE_END basic03" message is not outputted in those cases. Of course it is outputted if it is run.

Indeed, having a proper way to list all planned test would be great but outside the scope of this PR. TBH even though BASIC03 is not always run I think this is fine to keep it listed as part of always planned test cases to avoid cases where `(nbr of finished tests) > (nbr of planned tests)` (and a progress > 100).

matsduf:
The title of this test case says "simplify" so by not caring about some special handling of Basic today (which might be different tomorrow) it will be simpler to understand and maintain. What happens if Basic00 is removed?

pnax:
> There are 9 modules, and if each of them adds 10 (10%) to progress it would be good enough. When the test is completed it will jump to 100, which is fine from a user perspective.

It would require to know which test is currently running (to know in which module we are and when we leave it). Is there any other way to get this info without using the logger callback? Otherwise if we need to rely on the callback, I don't see any issue with the current proposal.

pnax:
> What happens if Basic00 is removed?

Not much. The idea is to compute `N_p` (the number of planned test cases) and `N_f` (number of finished test cases) so that at any given time `N_p >= N_f`. If Basic00 is removed, `N_p` will always be greater than `N_f` (if we forget to update the Backend).

matsduf:
> > There are 9 modules, and if each of them adds 10 (10%) to progress it would be good enough. When the test is completed it will jump to 100, which is fine from a user perspective.
> 
> It would require to know which test is currently running (to know in which module we are and when we leave it). Is there any other way to get this info without using the logger callback? Otherwise if we need to rely on the callback, I don't see any issue with the current proposal.

There is a MODULE_VERSION for the start of the module and MODULE_END when the module ens.

```
$ zonemaster-cli --show-testcase --level DEBUG iis.se  --raw | grep MODULE
   0.00 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Basic; version=v1.0.18
   0.94 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Basic
   0.94 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Address; version=v1.0.8
   2.93 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Address
   2.93 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Connectivity; version=v1.0.17
   4.64 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Connectivity
   4.65 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Consistency; version=v1.1.16
   7.33 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Consistency
   7.33 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Delegation; version=v1.0.20
   7.52 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Delegation
   7.52 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::DNSSEC; version=v1.1.57
   8.31 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::DNSSEC
   8.31 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Nameserver; version=v1.0.27
   8.89 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Nameserver
   8.89 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Syntax; version=v1.0.10
   9.15 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Syntax
   9.15 DEBUG     UNSPECIFIED    MODULE_VERSION   module=Zonemaster::Engine::Test::Zone; version=v1.0.14
   9.52 DEBUG     UNSPECIFIED    MODULE_END   module=Zonemaster::Engine::Test::Zone
```

pnax:
Ok I think I misunderstood what you were saying. I thought you were saying that using the logger callback was to heavy on resources. Now I understand that you are just asking to use the same logic but look at the module end point instead of the test case end point.

The current proposal result in roughly the same number of update of the progress value as before. An advantage is that the progress value evolves consistently based on the configured profile. A disadvantage is that when there is one database write per finished test case.
With your proposal, we'll have a coarse progress value (disadvantage) with fewer database updates (advantage).

In both proposal there is a need to stay aligned with how the engine deals with the logged messages (disadvantage).

Now choosing one solution or another, since I provided this PR, I think I can't easily pick one (or I'll chose mine for its preciseness).

matsduf:
> The current proposal result in roughly the same number of update of the progress value as before. An advantage is that the progress value evolves consistently based on the configured profile. A disadvantage is that when there is one database write per finished test case. With your proposal, we'll have a coarse progress value (disadvantage) with fewer database updates (advantage).

I can accept your proposal. I assume then that the load of that is not much higher than basing the progress on module level. I suggest that the special treatment of basic is removed to simplify the solution.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +102,56 @@ sub run {
...
+                    my $current_module = ucfirst( lc $entry->{_module} );
+                    my $current_testcase = $entry->{args}->{testcase};
+                    my $current_module_method = $current_module . "::" . $current_testcase;
+
+                    $counter_for_progress_indicator{executed}{$current_module_method}++;
```
blacksponge, line 139: Is there an advantage to use a hash to keep track of the number of planned / executed tests instead of simply a int?

pnax: Very good remark, I kept this part from the old code without too much thinking. There is no point in using a hash anymore. I'll provide an update.

blacksponge (approved):
This looks fine to me.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +101,34 @@ sub run {
...
+        }
+
+        Zonemaster::Engine->logger->callback(
+            sub {
+                my ( $entry ) = @_;
+                if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
+                    $nbr_testcases_finished++;
+                    # limit to max 99%, 100% is reached when data is stored in database
+                    my $progress_percent = 99 * $nbr_testcases_finished /  $nbr_testcases_planned;
+                    $self->{_db}->test_progress( $test_id, $progress_percent );
```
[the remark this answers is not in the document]
pnax: Good catch. Fixed.

matsduf (commented):
This should be fine.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +101,34 @@ sub run {
...
+                    $nbr_testcases_planned++;
+                }
+            }
+        }
+
+        Zonemaster::Engine->logger->callback(
+            sub {
+                my ( $entry ) = @_;
+                if ( $entry->{tag} and $entry->{tag} eq 'TEST_CASE_END' ) {
+                    $nbr_testcases_finished++;
```
mattias-p, line 123: Maybe check here that $entry->{testcase} is actually included in the profile? Otherwise we risk overshooting our 100%, like @matsduf says.

mattias-p (commented):
This is a very nice simplification.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -209,6 +209,17 @@ sub recent_test_hash_id {
     return $recent_hash_id;
 }
 
+=head2 test_progress($test_id, $progress)
+
+Takes a test_id and returns the current progress value for the associated test.
+If progress is set, update the progress value of the test.
+If the progress value is 1, set the C<started_at> field to the current time in UTC.
+If the progress value is 100, set the C<ended_at> field to the current time in UTC.
```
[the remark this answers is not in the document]
matsduf: What do you mean? You check the progress by `test_progress()` until progress has reached 100 and then you can use `get_test_results()` to get the test results.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -209,6 +209,17 @@ sub recent_test_hash_id {
...
+
+The progress value is an integer in the range 0-100.
```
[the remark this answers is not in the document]
matsduf: What do you mean? Progress is set to 0 when the test is created before Test Agent has picked it up.

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
matsduf: `test_results()` does not set `progress`, does it?

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
matsduf: Should it be required that the value is higher than the current value to be set so that progress does not go backwards?

matsduf:
The code says that "started_at" is set when progress is set to exactly 1. Is there a risk that this update will jump the value 1?

I think "started_at" should be set when a progress higher than 0 is set.
```
        if ( $progress == 1 ) {
            $dbh->do(
                q[
                    UPDATE test_results
                    SET progress = ?,
                        started_at = ?
                    WHERE hash_id = ?
                      AND progress <> 100
                ],
                undef,
                $progress,
                $self->format_time( time() ),
                $test_id,
            );
        }
```

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
pnax: The value `1` has a specific meaning, it means that the test has been picked up by an agent.

matsduf:
> There are few comments on the `test_progress()` method. Its logic is not changed here. I only added documentation (with a mistake I've fixed). And I agree that we could clarify how this method works and what are allowed values or not.

Today the first value after creation is 1, and if that has not changed then `test_progress()` can probably be as is.

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
pnax: In fact when $progress = 0, the `if ( $progress ) { ... }` condition will not pass. So there will be no update in database.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -158,6 +101,34 @@ sub run {
...
+                    # limit to max 99%, 100% is reached when data is stored in database
+                    my $progress_percent = int( 99 * $nbr_testcases_finished /  $nbr_testcases_planned );
```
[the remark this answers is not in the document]
pnax: I don't see how this could happen, $nbr_testcases_finished is always greater than 0 at this point.

matsduf (approved):

PR #991, opened 2022-05-02
Removes Locale::Msgfmt as required from Makefile.PL

### LINKED ISSUE #990: Locale::Msgfmt is not listed as dependency but required by Makefile.PL
[Installation] document does not list Locale::Msgfmt as a dependency for any OS, but it is required in Makefile.PL. The result is that it is fetched from CPAN.

In d341439101e15812115f80941ddda9eaaa57deb1 it was added to Makefile that is should be required, and that as dependency in the installation document.

In 82e24a01ae066fab88f39d83d711c77b66ce7166 the installation ducument was updated and the dependency was removed.

[Installation]: https://github.com/zonemaster/zonemaster-backend/blob/develop/docs/Installation.md

matsduf:
## Purpose

Removes Locale::Msgfmt from Makefile.PL as required since it is not used

## Context

Resolves #990

## Changes

Updates Makefile.PL

## How to test this PR

Backend should not complain of missing Locale::Msgfmt and Locale::Msgfmt should not be installed.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -9,8 +9,6 @@ bugtracker 'https://github.com/zonemaster/zonemaster-backend/issues';
 # "2.1.0" could be declared as "2.001" but not as "2.1"
 # (see Zonemaster::LDNS below)
 
-configure_requires 'Locale::Msgfmt' => 0.15;
-
 requires
   'Class::Method::Modifiers'    => 1.09,
   'Config::IniFiles'            => 0,


pnax (approved):
To me this is fine. @blacksponge could you also review, since this was introduced in one of your PR?

blacksponge (approved):
It probably was a copy/paste from engine when I added internationalization in backend. Looking at engine it seems that it has been removed in https://github.com/zonemaster/zonemaster-engine/commit/42391006d78f180a2b6ebeb092308196da1e0feb so I would say it is safe to do so here as well.

PR #993, opened 2022-05-02
Document testing environment variables

pnax:
## Purpose

Document testing environment variables

## Context

#737

## Changes

Documentation

## How to test this PR

Documentation only

### DIFF
--- a/docs/EnvironmentVariables.md
+++ b/docs/EnvironmentVariables.md
@@ -21,3 +21,15 @@
 * `ZM_BACKEND_RPCAPI_LOGJSON`: Setting it to any thruthy value (non-empty
   string or non-zero number) will configure the logger to log in JSON format,
   undefined by default.
+
+## Variables used by unit test scripts
+
+* `TARGET`: Set the database to use. If empty or undefined, "SQLite" will be
+  assumed. Accepted values are:
+  * `SQLite`
+  * `PostgreSQL`
+  * `MySQL`
+
+* `ZONEMASTER_RECORD`: Setting it to any thruthy value (non-empty string or
+  non-zero number) will record the data from the test to a file. Otherwise the
+  data is loaded from a file.


docs/EnvironmentVariables.md
```diff
@@ -21,3 +21,14 @@
 * `ZM_BACKEND_RPCAPI_LOGJSON`: Setting it to any thruthy value (non-empty
   string or non-zero number) will configure the logger to log in JSON format,
   undefined by default.
+
+## Variables specific to testing
```
matsduf, line 25: Everything is about testing. I suggest:
```
## Variables used by unit test scripts
```

docs/EnvironmentVariables.md
```diff
@@ -21,3 +21,14 @@
...
+
+* `TARGET`: Set the database to use. If undefined, will assume SQLite.
+  Accepted values are:
+  * `SQLite`
+  * `PostgreSQL`
+  * `MySQL`
+
+* `ZONEMASTER_RECORD`: If set, the data from the test is recorded to a file.
+  Otherwise the data is loaded from a file.
```
matsduf, line 34: Is "set" correct here? Shouldn't it be "if true"? If set to 0 or empty string will give the same affect as unset, won't it?

docs/EnvironmentVariables.md
```diff
@@ -21,3 +21,14 @@
 * `ZM_BACKEND_RPCAPI_LOGJSON`: Setting it to any thruthy value (non-empty
   string or non-zero number) will configure the logger to log in JSON format,
   undefined by default.
+
+## Variables specific to testing
+
+* `TARGET`: Set the database to use. If undefined, will assume SQLite.
```
matsduf, line 27: If think the following is more correct and better language:
```
If empty or undefind, "SQLite" will be assumed.
```

matsduf (changes requested):

docs/EnvironmentVariables.md
pnax: updated

docs/EnvironmentVariables.md
pnax: fixed

docs/EnvironmentVariables.md
pnax: updated

matsduf (approved):
It is great to have to documented.

PR #994, opened 2022-05-02
Update Travis

pnax:
## Purpose

* Fix not found versions
* Test against Perl 5.16 (Travis requires Ubuntu 18.04)
* Use generic Perl version when possible
* Check MANIFEST
* Use stages to check MANIFEST first (run tests on success only)

## Context

https://github.com/zonemaster/zonemaster-ldns/pull/145#pullrequestreview-958795810

## Changes

.travis.yml

## How to test this PR

Travis tests should succeed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ addons:
     - libtext-csv-perl
     - libtool
     - m4
+
       # From Zonemaster Backend installation instruction
     - libclass-method-modifiers-perl
     - libconfig-inifiles-perl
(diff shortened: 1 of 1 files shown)


pnax:
It looks like Mojolicious requires  Perl 5.16. Should we hence drop support for Perl 5.14?

matsduf:
> It looks like Mojolicious requires Perl 5.16. Should we hence drop support for Perl 5.14?

That would be fine. CentOS 7 is the one with lowest version and that is 5.16.

pnax:
Ok, then I'll see how to enable Perl 5.16 in Travis.

pnax:
* I've added a test with Perl 5.16.
* I also added a first check that runs `make distcheck`

matsduf (approved):

PR #1005, opened 2022-05-12
Normalize domain column

pnax:
## Purpose

Store the normalized domain value in the "domain" column in database.

## Context

Follow-up of #983 

## Changes

Database files.
Normalize passed "domain" in get_test_history() before retrieval (since domains are now normalized in database).
Update the upgrade script.

## How to test this PR

Unit tests are added, they should pass. Or start a test with a trailing dot to the domain (and/or with uppercase letters), the domain stored in database should be normalized.

Using the upgrade script should remove any trailing dot from the domain column, as well as down casing it.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -136,7 +136,7 @@ sub create_new_test {
 
     my $dbh = $self->dbh;
 
-    $test_params->{domain} = $domain;
+    $test_params->{domain} = _normalize_domain( $domain );
 
     my $fingerprint = $self->generate_fingerprint( $test_params );
     my $encoded_params = $self->encode_params( $test_params );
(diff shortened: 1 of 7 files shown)


matsduf (approved):

matsduf:
> I guess this requires a database update to normalize all stored domains. This could be done in a second stage, once this PR is approved.

It would be nice, including down-casing the domain name, it is however, not strictly necessary by this PR. The issue of domain names being stored with and without trailing dot is an old issue.

pnax:
I updated the upgrade script since this change would alter how `get_test_history()` works. With this change some tests might be never reachable if the database is not updated.

mattias-p (approved):
LGTM

mattias-p:
After the fix in #1030 I successfully tested this for v2022.1.

Issue #1011, opened 2022-05-17
t/batches.t: UNIQUE constraint failed


matsduf:
When running the tests, `batches.t` returns "PASS", but something fails in the Perl code:
```
$ PERL_DL_NONLAZY=1 "/usr/local/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(1, 'inc', 'blib/lib', 'blib/arch')" t/batches.t
(...)
# Subtest: duplicate user should fail
DBD::SQLite::db do failed: UNIQUE constraint failed: users.username at /usr/home/matsd/.cpanm/work/1652791322.55779/Zonemaster-Backend-8.1.0/blib/lib/Zonemaster/Backend/DB.pm line 117.
    ok 1
    ok 2 - correct error type
    ok 3 - correct error message
    ok 4 - correct data type returned
    1..4
ok 7 - duplicate user should fail
(...)
All tests successful.
Files=1, Tests=8,  1 wallclock secs ( 0.03 usr  0.01 sys +  0.77 cusr  0.12 csys =  0.92 CPU)
Result: PASS
```

FreeBSD 13.0
SQLite 3.36.0 in p5-DBD-SQLite-1.70

pnax:
This is the expected behavior of the test. This message is present for all databases. It is emitted by the driver and this is the expected behavior when you try to insert data that are already present (the UNIQUE constraint fails). The test properly catches the error.

matsduf:
Is it possible to make the script out that is what should be expected? Not to confuse the users like me.

matsduf:
Resolved by #1015

PR #1015, opened 2022-05-18
Capture error message on constraint violation

pnax:
## Purpose

Catch stderr output on passing tests. When testing for UNIQUE constraint violation an error message is displayed. This message can be misleading and let think that the test failed. Hence such message should not be outputted but captured.

## Context

#1011

## Changes

t/batches.t

## How to test this PR

Running `prove t/batches.t` should not output any error message.


matsduf:
It looks fine, but conflict must be resolved.

mattias-p:
Didn't we say this could happen in production too? Maybe I misremember.

t/batches.t
```diff
@@ -2,6 +2,7 @@ use strict;
 use warnings;
 use 5.14.2;
 
+use Capture::Tiny qw( capture );
```
mattias-p, line 5: This is a new dependency. Maybe add it as `test_requires` in Makefile.PL and the installation instruction too?

t/batches.t
```diff
@@ -310,8 +311,10 @@ subtest 'duplicate user should fail' => sub {
     my $config = Zonemaster::Backend::Config->parse( $config );
     my $backend = init_from_config( $db_backend, $config );
 
-    dies_ok {
-        $backend->add_api_user( { username => $user->{username}, api_key => "another api key" } );
+    capture {
```
mattias-p, line 314: As an alternative to this, have you tried using [PrintWarn](https://metacpan.org/pod/DBI#PrintWarn) or one of the other techniques suggested there?

pnax: Nope I didn't tried, but this looks way better. I've provided an update.

t/batches.t
pnax: Not applicable anymore since we now rely on a DBI attribute as you suggested.

pnax:
> Didn't we say this could happen in production too? Maybe I misremember.

Such message will be logged in production. I feel a little bit uncomfortable when it comes to suppress any error message coming up from the database in production.

However the goal here is to remove the error message while testing for the feature, which is something acceptable I guess.

matsduf (approved):
Looks fine now.

Is it expected that there should be no description on subtest 1?
```
# Subtest: duplicate user should fail
    ok 1
    ok 2 - correct error type
    ok 3 - correct error message
    ok 4 - correct data type returned
    1..4
ok 5 - duplicate user should fail
```

blacksponge (approved):

blacksponge (approved):

matsduf (approved):

matsduf:
Please merge.

PR #1016, opened 2022-05-22
Updates Swedish translation Backend

matsduf:
## Purpose

This PR updates the translation (PO file) for Swedish.

## How to test this PR

Change to Swedish and verify that translation comes out as expected for updated msgids and msgstrs.

### DIFF
--- a/share/sv.po
+++ b/share/sv.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-11-17 08:57+0100\n"
+"POT-Creation-Date: 2022-05-22 15:30+0000\n"
 "PO-Revision-Date: 2021-11-17 17:32+0100\n"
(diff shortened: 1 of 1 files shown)


share/sv.po
```diff
@@ -2,75 +2,76 @@ msgid ""
...
+"Warning: Zonemaster::LDNS not compiled with IDN support, cannot handle non-"
+"ASCII names correctly."
 msgstr ""
-"Domännamnet är inte ett giltigt IDN-namn och kan inte konverteras till A-"
-"label-format"
+"Varning: Zonemaster::LDNS är inte kompilerad med IDNA-stöd, så enbart ASCII-"
+"namn kan hanteras."
+
+msgid "Invalid digest format"
+msgstr "Ogiltigt format på DS-digest"
```
mattias-p, line 28: Have you considered removing the reference to DS from the msgstr? The msgid speaks of digests in general and does not mention the DS record type. This means that an identical msgid could be used in another place where the DS reference isn't applicable. That would render the Swedish translation in that place incorrect, and we would have no good way of noticing.

mattias-p (approved):
Looking good

PR #1021, opened 2022-05-23
Update installation doc

blacksponge:
## Purpose

* The current version of JSON Validator on Ubuntu is up to date, the note no longer applies.
* Some build dependencies are missing.

## Context

\-

## Changes

* Remove JSON Validator note in Ubuntu installation
* Add Test::Difference and Test::Exception to the list of test dependencies

## How to test this PR

Installation should work.

PR #1023, opened 2022-05-23
Installation on Rocky

mattias-p:
## Purpose

This PR fixes some dependency issues and a bug that I noticed for Rocky.

## Context

N/A

## Changes

* Installs OS package perl-Test-Mojo so the tests don't fail on installation.
* Fixes a bug in t/lifecycle.t that made us break an assumption made by Mojolicious.
* Installs OS packages for perl-Test-Differences and perl-Test-Exception, which we depend on as noted in #1021.

## How to test this PR

Follow the installation instruction using Rocky Linux.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -79,7 +79,7 @@ for Zonemaster::Backend, see the [declaration of prerequisites].
 Install dependencies available from binary packages:
 
 ```sh
-sudo dnf -y install jq perl-Class-Method-Modifiers perl-Config-IniFiles perl-DBD-SQLite perl-DBI perl-HTML-Parser perl-JSON-RPC perl-libwww-perl perl-Log-Dispatch perl-Mojolicious perl-Net-Server perl-Parallel-ForkManager perl-Plack perl-Plack-Middleware-ReverseProxy perl-Plack-Test perl-Role-Tiny perl-Router-Simple perl-String-ShellQuote perl-Test-Mojo perl-Test-NoWarnings perl-Test-Warn perl-Try-Tiny perl-libintl redhat-lsb-core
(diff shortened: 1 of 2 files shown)


matsduf (approved):

docs/Installation.md
```diff
@@ -89,7 +89,7 @@ sudo dnf -y install jq perl-Class-Method-Modifiers perl-Config-IniFiles perl-DBD
 Install dependencies not available from binary packages:
 
 ```sh
-sudo cpanm Daemon::Control JSON::Validator Log::Any Log::Any::Adapter::Dispatch Starman Plack::Middleware::ReverseProxy
```
pnax, line 92: There is no package `perl-Plack-Middleware-ReverseProxy` available for CentOS 7.

matsduf: > There is no package `perl-Plack-Middleware-ReverseProxy` available for CentOS 7.

But available from CPAN?

mattias-p: Okay, I'll revert that change.

matsduf: In Engine a separate section for CentOS 7 has been proposed in https://github.com/zonemaster/zonemaster-engine/pull/1080. I suggest that we do the same thing here to make it as good as possible for Rocky Linux.

pnax: The installation document for Backend is quite long and heavy, I would suggest not to add an extra section specific to CentOS 7 since the installation works with the instructions from the Rocky Linux section if you use `yum` instead of `dnf`.

pnax (approved):

docs/Installation.md
mattias-p: Installing five or six distributions from CPAN instead isn't a big difference.

matsduf: > The installation document for Backend is quite long and heavy, I would suggest not to add an extra section specific to CentOS 7 since the installation works with the instructions from the Rocky Linux section if you use `yum` instead of `dnf`.

Then, at least that information should be stated. And there should be a headline (entrance) for CentOS Linux 7. Furthermore, PostgreSQL is not support for CentOS Linux 7.

mattias-p: I'll leave it to @pnax to make the CentOS updates.

PR #986, opened 2022-04-28
Update IDN support warning message

pnax:
## Purpose

Avoid getting into to much details in the warning message.

## Context

https://github.com/zonemaster/zonemaster-backend/pull/984#discussion_r860804409

## Changes

* Update warning message
* Provide a French translation

## How to test this PR

The message should be changed and correctly translated (based on available translations).

## Other translations

Translations in other language for the update string can be provided as commit directly to this PR, or as comments that I'll add to the PR.

PR #1029, opened 2022-06-01
Add missing text library to TestAgent.pm

tgreenx:
## Purpose

This PR adds a missing library to TestAgent.pm, to be able to show warning messages. Currently it will crash:
`[ERROR] [main] Test died: 8b96afcc7a6ad5b8: Undefined subroutine &Zonemaster::Backend::TestAgent::__ called at /usr/local/share/perl/5.32.1/Zonemaster/Backend/TestAgent.pm line 206`

## Context

See [#986](https://github.com/zonemaster/zonemaster-backend/pull/986#issuecomment-1143388127) (comments)

## How to test this PR

Install Zonemaster::LDNS, without IDN support
Create a test with IDN characters (e.g., `café.fr`)
Start the testagent
Logs should show a warning message instead of crashing, like so:
`[WARNING] [main] Warning: Zonemaster::LDNS not compiled with IDN support, cannot handle non-ASCII names correctly. at /usr/local/share/perl/5.32.1/Zonemaster/Backend/TestAgent.pm line 208.`

### DIFF
--- a/lib/Zonemaster/Backend/TestAgent.pm
+++ b/lib/Zonemaster/Backend/TestAgent.pm
@@ -9,6 +9,7 @@ use DBI qw(:utils);
 use JSON::PP;
 use Scalar::Util qw( blessed );
 use File::Slurp;
+use Locale::TextDomain qw[Zonemaster-Backend];
 
 use Zonemaster::LDNS;
 


matsduf (approved):

tgreenx:
To translate messages, environment variable "LANG" is used in Backend. However, the Perl default uses "LANGUAGE". 
To circumvent that behavior, I added a change to give precedence to "LANG" instead, analogously to what is already done in other parts of Backend (see [here](https://github.com/zonemaster/zonemaster-backend/blob/master/script/zonemaster_backend_rpcapi.psgi#L22-L25)).

blacksponge (approved):
Tested on Ubuntu 22.04, working as expected.

mattias-p (approved):
This should be fine.

matsduf (approved):

PR #1030, opened 2022-06-01
Fix some problems when upgrading the database to the schema for 8.2.0

mattias-p:
## Purpose

This PR fixes problems found during release testing.

## Context

Fixes problems introduced in #1005 and #970.

## Changes

* MariaDB: Adds an instruction to back up the zonemaster database before running the migration script. This is needed because ALTER TABLE statements don't respect transactions on MariaDB.
* MariaDB: Refactors the migration code so the somewhat riskiest operation comes first. I.e. the one to add the unique constraint to users.username.
* PostgreSQL: Fixes the normalization of domain names.

### PostgreSQL
1. Install Backend 8.1.0.
2. Run `zmtest zonemaster.net` and `zmtest zonemaster.net.`.
3. Install this branch of Backend on top of it. (Or current develop to reproduce the bug.)
4. Run the database migration script.
5. Run `echo "select domain from test_results;" | sudo -u zonemaster psql zonemaster`.

Before this fix domain names with trailing dots get normalized to the empty string.
With this fix domain names are properly normalized.

### MariaDB
1. Pretend we have the previous version installed, that we've somehow managed to get two users with the same name into the database:
   1. Install Backend 8.1.0.
   2. Run `zmb add_api_user --username jdoe --api-key asdfasdf` to create a user.
   3. Run `echo "insert into users (username, api_key) values ('jdoe', 'qwerqwer');" | sudo mysql zonemaster`. To force the creation of a second user with the same name.
2. Upgrade to this version of Backend:
   1. Install this branch of Backend on top of it.
3. Follow the database upgrade instruction:
   1. Take a backup of the zonemaster database.
   2. Run the database migration script.
4. Trying to solve the problem we notice that the database was broken by the migration script:
   1. Notice the error message about the duplicate entry.
   2. Run `echo "delete from users where api_key = 'qwerqwer';" | sudo mysql zonemaster` so we can add the unique constraint.
   3. Run the database migration script again to see it fail. (The tables were already altered.)
5. Get the database back into a healthy state:
   1. Restore the database from the backup.
6. Fix the duplicate entry problem before running the upgrade script.
   1. Run `echo "delete from users where api_key = 'qwerqwer';" | sudo mysql zonemaster` so we can add the unique constraint.
   2. Run the database migration script again.
7. Make sure the migration was successful:
   1. Run `echo "select * from table_constraints where constraint_schema = 'zonemaster';" | sudo mysql information_schema` and make sure the unique constraint has been added.

### DIFF
--- a/share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
+++ b/share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
@@ -28,21 +28,21 @@ sub patch_db_mysql {
     my $db = Zonemaster::Backend::DB::MySQL->from_config( $config );
     my $dbh = $db->dbh;
 
+    # add table constraints
+    $dbh->do( 'ALTER TABLE users ADD CONSTRAINT UNIQUE (username)' );
+    $dbh->do( 'ALTER TABLE test_results ADD CONSTRAINT UNIQUE (hash_id)' );
+
+    # update columns names, data type and default value
+    $dbh->do( 'ALTER TABLE test_results MODIFY COLUMN id BIGINT AUTO_INCREMENT' );
+    $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN creation_time created_at DATETIME NOT NULL' );
+    $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN test_start_time started_at DATETIME DEFAULT NULL' );
+    $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN test_end_time ended_at DATETIME DEFAULT NULL' );
+
+    $dbh->do( 'ALTER TABLE batch_jobs CHANGE COLUMN creation_time created_at DATETIME NOT NULL' );
+
     $dbh->{AutoCommit} = 0;
 
     try {
-        # update columns names, data type and default value
-        $dbh->do( 'ALTER TABLE test_results MODIFY COLUMN id BIGINT AUTO_INCREMENT' );
-        $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN creation_time created_at DATETIME NOT NULL' );
-        $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN test_start_time started_at DATETIME DEFAULT NULL' );
-        $dbh->do( 'ALTER TABLE test_results CHANGE COLUMN test_end_time ended_at DATETIME DEFAULT NULL' );
-
-        $dbh->do( 'ALTER TABLE batch_jobs CHANGE COLUMN creation_time created_at DATETIME NOT NULL' );
-
-        # add table constraints
-        $dbh->do( 'ALTER TABLE test_results ADD CONSTRAINT UNIQUE (hash_id)' );
-        $dbh->do( 'ALTER TABLE users ADD CONSTRAINT UNIQUE (username)' );
-
         # normalize "domain" column
         $dbh->do(
             q[
--- a/share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
+++ b/share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
@@ -105,7 +105,7 @@ sub patch_db_postgresql {
         $dbh->do(
             q[
                 UPDATE test_results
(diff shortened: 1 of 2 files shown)


pnax:
Indeed MySQL/MariaDB does some [implicit commit](https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html). Thanks for spotting that out.

pnax:
Regarding MariaDB test, you cannot create two users with the same username [with Backend 8.1.0](
https://github.com/zonemaster/zonemaster-backend/blob/v8.1.0/lib/Zonemaster/Backend/DB.pm#L104-L105).

pnax:
> 9. Run the database migration script again to see it fail. (The tables were already altered.)

This won't work because this PR will make the constraint fail early and we install from this PR  "_4. Install this branch of Backend on top of it._". However installing Backend from develop (without this PR) will leave the database in an inconsistent state during the upgrade. This won't alter the purpose and the fix the PR provides though.

mattias-p:
> Regarding MariaDB test, you cannot create two users with the same username [with Backend 8.1.0](https://github.com/zonemaster/zonemaster-backend/blob/v8.1.0/lib/Zonemaster/Backend/DB.pm#L104-L105).

Of course, you're right. I guess I was going a bit quick when I wrote up the "how to test" steps. When I did my testing I used an SQL insert statement to emulate data created with an older version that didn't have the check. I should mention that I didn't actually check if we've ever had such a version - I just assumed that was the case.

Anyway, I updated the "how to test" steps with an "insert into".

pnax:
This was just some remark that the old testing procedure will not work exactly as you would expect I guess. I did understand your testing procedure (which was detailed enough to understand the purpose of each steps). Thanks for making it even more detailed.

What I wanted to address here is that with these changes the first command in the upgrade script is the creation of the constraints which will fail in the testing scenario. This would happen before any other update to the tables' schema. Therefore when we remove the user, we can run the upgrade script again without any error (because the table was not altered since the script failed at its first instruction). In other words we won't see the script fail in 4.iii. But this is definitely not a big issue :)

share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
```diff
@@ -105,7 +105,7 @@ sub patch_db_postgresql {
         $dbh->do(
             q[
                 UPDATE test_results
-                SET domain = RTRIM('.', domain)
+                SET domain = RTRIM(domain, '.')
```
matsduf, line 108: RTRIM is not fully correct here. "iis.se..", "iis.se." and "iis.se" will all be stored as "iis.se", which is fine for "iis.se." because it was tested as "iis.se", but "iis.se.." was not.

matsduf (approved):

share/patch/patch_db_zonemaster_backend_ver_8.2.0.pl
mattias-p: I guess the where clause could have been more restrictive. Do you know if we've had releases of Zonemaster that allowed zone names with double dots to be entered into the database?

pnax: Good point. We could use for PostgreSQL `LEFT(domain, -1)` that removes the last character from the string.

pnax: The problem is also present for MySQL, the LEFT() function does not work the same, but we could use something like `LEFT( domain, LENGTH(domain) - 1 )`.

Documentations for reference:
https://dev.mysql.com/doc/refman/5.7/en/string-functions.html
https://www.postgresql.org/docs/10/functions-string.html

pnax: see #1031

matsduf: > I guess the where clause could have been more restrictive. Do you know if we've had releases of Zonemaster that allowed zone names with double dots to be entered into the database?

I am positive that Backend before some version ago let double dots through. One test could be if there are such entries in the zonemaster.net database. Please note that I have approved this PR anyway.

Issue #1039, opened 2022-06-12
Backend does not respect disabled IPv6 in profile


matsduf:
If  IPv6 is set to false in the default profile (/usr/local/share/perl/5.26.1/auto/share/dist/Zonemaster-Engine/profile.json on Ubuntu) the setting is respected by both Backend and CLI. If we instead create a new profile file and set IPv6 off in that file and use it, CLI will respect that IPv6 is turned off, but Backend will not.

Copy the default profile to "/etc/zonemaster/profile-changed.json" and introduce the following changes:
```
--- profile.json	2022-06-12 12:00:49.729268150 +0200
+++ profile-changed.json	2022-06-12 12:00:43.165354222 +0200
@@ -8,7 +8,7 @@
     },
     "net" : {
         "ipv4" : true,
-        "ipv6" : true
+        "ipv6" : false
     },
     "no_network" : false,
     "resolver" : {
@@ -57,11 +57,11 @@
             "HAS_A_RECORDS" : "ERROR",
             "HAS_NAMESERVER_NO_WWW_A_TEST" : "INFO",
             "HAS_NAMESERVERS" : "INFO",
-            "HAS_PARENT" : "INFO",
-            "IPV4_DISABLED" : "INFO",
-            "IPV4_ENABLED" : "DEBUG",
-            "IPV6_DISABLED" : "INFO",
-            "IPV6_ENABLED" : "DEBUG",
+            "HAS_PARENT" : "ERROR",
+            "IPV4_DISABLED" : "NOTICE",
+            "IPV4_ENABLED" : "NOTICE",
+            "IPV6_DISABLED" : "NOTICE",
+            "IPV6_ENABLED" : "NOTICE",
             "NO_GLUE_PREVENTS_NAMESERVER_TESTS" : "CRITICAL",
             "NO_PARENT" : "CRITICAL",
             "NS_FAILED" : "ERROR",
```
Update "backend_config.ini" with the following:
```
[PUBLIC PROFILES]
default=/etc/zonemaster/profile-changed.json
```
Restart Backend and start a test with `zmb start_domain_test` (not with `zmtest` since it always turn IPv6 on, #1040) and fetch the result with `zmb get_test_results`. 

Note that the new profile is used because in the results you will find changed levels, but IPv6 is still enabled.
```
      {
        "message": "Parent domain 'fr' was found for the tested domain.\n",
        "level": "ERROR",
        "module": "BASIC"
      },
      {
[...]

PR #1041, opened 2022-06-12
Removes enabling IPv4 and IPv6 by default in zmtest

### LINKED ISSUE #1040: zmtest should not set IPv4 and IPv6 on by default
When `zmtest` starts a test it sets both IPv4 and IPv6 on, unless the user has explicitly turned it off. `zmtest` should just respect what is set in the profile, i.e. leave IPv4 and IPv6 unset by default.

matsduf:
## Purpose

`zmtest` enables by default both IPv4 and IPv6. It should rather let profile setting rule, unless disabled by option.

## Context

Fixes #1040

## How to test this PR

1. Disable IPv4 or IPv6 in the default policy file (see #1039 why you currently must change there).
2. Start a test with `zmtest`
3. Verify that the selected protocol remains disabled.

### DIFF
--- a/script/zmtest
+++ b/script/zmtest
@@ -15,8 +15,9 @@ usage () {
     echo >&2
     echo "Options:" >&2
     echo "  -s URL --server URL   Zonemaster Backend to query. Default is http://localhost:5000/" >&2
-    echo "  --noipv4              Run the test without ipv4." >&2
-    echo "  --noipv6              Run the test without ipv6." >&2
+    echo "  --noipv4              Run the test with IPv4 disabled." >&2
+    echo "  --noipv6              Run the test with IPv6 disabled." >&2
+    echo "                        IPv4 and IPv6 follow the profile setting unless disabled by option." >&2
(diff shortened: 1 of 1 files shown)


mattias-p (approved):
This is more limited than what I expected. But it looks fine!

matsduf:
> This is more limited than what I expected. But it looks fine!

I had forgotten that I already created it. I took a minimal approach to the issue at that time. If you want to do something more elaborate you can use `zmb`. If requested, `zmtest` could be updated in the future to have full power.

matsduf:
@pnax, is this fine to you?

pnax (approved):
lgtm after rebasing on top of develop to test with changes from #1046

tgreenx:
**Release testing 2022-2**

Works as expected.

I found an issue with script `zmtest` though, regarding options `--noipv4` and `--noipv6`. I opened https://github.com/zonemaster/zonemaster-backend/pull/1066 to fix that.

PR #1043, opened 2022-06-22
Fixes when IPv4/6 setting is overwritten by default profile

### LINKED ISSUE #1039: Backend does not respect disabled IPv6 in profile

matsduf:
## Purpose

Prevents default profile to override IPv4 or IPv6 setting in additional profile. This can happen when the `ipv4` or `ipv6` parameter keys are unset and the additional profile has a different setting from the default profile.

## Context

Resolves #1039. The resolution was proposed by @pnax in an email.

A side affect of this solution is that two tests with the same effective values can get different "fingerprints". That will happen if the IPv4 and IPv6 parameters are unset in one test and at least one of them is set to the same value as the effective profile in the other.

## How to test this PR

1. Keep the default profile with both IPv4 and IPv6 enabled ("true").
2. Create an additional profile with IPv6 disabled ("false").
3. Make the additional profile available via backend_config.ini.
4. Create a test with zmb with IPv4 and IPv6 unset and select that other profile.
5. Verfiy that the test was run with IPv6 disabled.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -617,8 +617,8 @@ sub _project_params {
     my %projection = ();
 
     $projection{domain}   = _normalize_domain( $$params{domain} // "" );
-    $projection{ipv4}     = $$params{ipv4}       // $profile->get( 'net.ipv4' );
-    $projection{ipv6}     = $$params{ipv6}       // $profile->get( 'net.ipv6' );
+    $projection{ipv4}     = $$params{ipv4};
+    $projection{ipv6}     = $$params{ipv6};
     $projection{profile}  = lc( $$params{profile} // "default" );
 
     my $array_ds_info = $$params{ds_info} // [];
--- a/t/db.t
+++ b/t/db.t
@@ -20,18 +20,18 @@ sub encode_and_fingerprint {
 subtest 'encoding and fingerprint' => sub {
 
     subtest 'missing properties' => sub {
-        my $expected_encoded_params = '{"domain":"example.com","ds_info":[],"ipv4":true,"ipv6":true,"nameservers":[],"profile":"default"}';
-
         my %params = ( domain => "example.com" );
 
+        my $expected_encoded_params = '{"domain":"example.com","ds_info":[],"ipv4":null,"ipv6":null,"nameservers":[],"profile":"default"}';
         my ( $encoded_params, $fingerprint ) = encode_and_fingerprint( \%params );
         is $encoded_params, $expected_encoded_params, 'domain only: the encoded strings should match';
         #diag ($fingerprint);
 
+        my $expected_encoded_params_v4_true = '{"domain":"example.com","ds_info":[],"ipv4":true,"ipv6":null,"nameservers":[],"profile":"default"}';
         $params{ipv4} = JSON::PP->true;
         my ( $encoded_params_ipv4, $fingerprint_ipv4 ) = encode_and_fingerprint( \%params );
-        is $encoded_params_ipv4, $expected_encoded_params, 'add ipv4: the encoded strings should match';
-        is $fingerprint_ipv4, $fingerprint, 'fingerprints should match';
+        is $encoded_params_ipv4, $expected_encoded_params_v4_true, 'add ipv4: the encoded strings should match';
+        isnt $fingerprint_ipv4, $fingerprint, 'fingerprints should not match';
     };
 
     subtest 'array properties' => sub {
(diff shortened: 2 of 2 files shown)


mattias-p:
The reason the IP options from the custom profile aren't respected is that the custom profile is never loaded by the RPCAPI.

A more robust solution would be if we loaded the custom profile in `start_domain_test`/`add_batch_test` and updated the test_params with defaults from the profile before passing it to generate_fingerprint() and friends.

But I like that the DB module is freed from the dependency on Profile. That's clean and nice.

matsduf:
> The reason the IP options from the custom profile aren't respected is that the custom profile is never loaded by the RPCAPI.
> 
> A more robust solution would be if we loaded the custom profile in `start_domain_test`/`add_batch_test` and updated the test_params with defaults from the profile before passing it to generate_fingerprint() and friends.
> 
> But I like that the DB module is freed from the dependency on Profile. That's clean and nice. Though the import itself and the fetching of the effective profile ought to be removed too.

@mattias-p, I think that is fine idea, but not for the fix release. Can we merge the PR to make Backend and CLI behave the same way (respect the custom profile)? And then make your suggestions as an issue for next release?

tgreenx:
LGTM. Tested and it works as expected.

tgreenx (approved):
LGTM. Tested and it works as expected.

PR #1053, opened 2022-09-16
Use Net::IP::XS directly

mattias-p:
## Purpose

Remove dependency on (to be deprecated) Zonemaster::Engine::Net::IP.

## Context

In zonemaster/zonemaster-engine#1119 Zonemaster::Engine::Net::IP we're dropping support for Net::IP.
The issue zonemaster/zonemaster-engine#1129 has been created to deprecate Zonemaster::Engine::Net::IP, with the intended migration path being to use Net::IP::XS directly.

## Changes

Calls into Zonemaster::Engine::Net::IP are replaced with calls into Net::IP::XS.

## How to test this PR

This change is covered by existing tests in `t/validator.t`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ addons:
       - liblocale-msgfmt-perl
       - libmail-rfc822-address-perl
       - libmodule-find-perl
-      - libnet-ip-perl
+      - libnet-ip-xs-perl
       - libpod-coverage-perl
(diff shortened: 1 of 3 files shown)


matsduf (approved):

blacksponge (approved):

mvw-afnic (approved):
Looks good to me.

Issue #790, opened 2021-06-14
RPCAPI deprecation warnings 


mattias-p:
## Purpose

It should be possible to report deprecation warnings back to RPCAPI clients. This would allow us to update the APIs in a friendlier fashion.

## Scope

This issue is about creating a channel or making room for delivery of deprecation warnings. It is not about introducing deprecation warnings.

## Design

1.  For each API method that today puts a scalar or an array in its top-level "result" property:
    1. Copy this method to a different name.
    2. Change the response object of the new method so that the top-level "result" property is an object instead of a scalar or array. Let the new object have a single property with the scalar/array data as its value.
    3. Mark the original methods as deprecated in the API documentation with a note encourage clients to migrate to the new method.
2. For each method in the API, add an optional property in the top-level "result" object. Let the new property be named "warnings" and let it have a non-empty list of human readable strings as its value.

Implementation-wise you'd probably have one method call the other and just modify the result. Documentation-wise you'd probably do something similar.

It would be nice if we could have the old methods emit deprecation warnings when you use them, encouraging clients to migrate to the new method. But that's exactly the problem with those methods. We can't. And that's exactly why they're being replaced.

## Alternatives

* Instead of adding a "warnings" property into the "result" object, we could add it to the top-level response object. However I don't think it's allowed in the JSON-RPC 2.0 specification, and I'm not sure we could coax [JSON::RPC](https://metacpan.org/pod/JSON::RPC) into doing this.
* Instead of giving immediate feedback about deprecated usage in the response object, we could introduce a new method that clients to poll to receive warnings generated in previous calls. It's not clear what mechanism we'd use to give the right deprecation warnings to the right client.
* Instead of giving immediate feedback about deprecated usage through the RPC-API, we could announce deprecations through documentation *only*. E.g. in the API documentation and release notes.

pnax:
I think this would be nice to make room for such channel. With regard to your suggested design, here are the methods that need to be updated:

| method | to update | top-level property |
| --- | :---: | :---: |
| `version_info` | | object |
| `profile_names` | yes | array |
| `get_language_tags` | yes | array |
| `get_host_by_name` | yes | array |
| `get_data_from_parent_zone` | | object |
| `start_domain_test` | yes | integer |
| `test_progress` | yes | integer |
| `get_test_results` | | object |
| `get_test_history` | yes | array |
| `get_test_params` | | object |
| `add_api_user` | yes | integer |
| `add_batch_job` | yes | integer |
[...]

Issue #791, opened 2021-06-14
Rename test id fields in RPC-API


mattias-p:
Today there are multiple fields in the API document referring to test ids. These fields are named either "test_id", "id" or "hash_id".

1. Choose a name to be given to all those fields.
2. Add replacement parameters with the new name for all those fields that don't already have the chosen name.
3. For each parameter that's gotten such a replacement, deprecate that parameter with a message encouraging migration to the replacement parameter.
4. Submitting both the old and new parameters together in the same request should probably be forbidden.

This should not be implemented before #790.

mattias-p:
Instead of renaming these parameters in existing methods, we're introducing new methods with updated parameter names. Also we're shifting our terminology from "test" to "job". See zonemaster/zonemaster#1175.

PR #1054, opened 2022-10-12
Always return an object from API methods and deprecate all existing API methods

pnax:
## Purpose

The current API methods return different object types in the "result" property. In #790 it has been discussed the need to always return an object to facilitate expanding the results. It was also noted that the API methods' names could be harmonized and improved. This first work is about providing new API methods based on the discussion in #790, and deprecate all existing API methods.

This is the first step of the process:

- [x] 1. create new methods and deprecate the old ones
- [ ] 2. update code and testcases to use the new methods (done in a follow up PR)

Mapping between old and new method names:
| current method name | current top-level property | new method name | new top-level property |
| --- | :---: | --- | :---: |
| `version_info` | object | `system_versions` | object |
| `profile_names` | array | `conf_profiles` | object |
| `get_language_tags` | array | `conf_languages` | object |
| `get_host_by_name` | array | `lookup_address_records` | object |
| `get_data_from_parent_zone` | object | `lookup_delegation_data` | object |
| `start_domain_test` | integer | `job_create` | object |
| `test_progress` | integer | `job_status` | object |
| `get_test_results`| object | `job_results` | object |
| `get_test_params` | object | `job_params` | object |
| `get_test_history` | array | `domain_history` | object |
| `add_api_user` | integer | `user_create` | object |
| `add_batch_job` | integer | `batch_create` | object |
| `get_batch_job_result` | object | `batch_status` | object |

## Context

Partially addresses #790.

## Changes

Deprecate all existing API methods, see the array in the Purpose section for the new method names.

## How to test this PR

Test should pass. At this stage the new methods are not used.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -469,7 +469,7 @@ Example response:
 ```
 
 
-##### `"result"`
+#### `"result"`
 
 An object with the following properties:
 
@@ -478,7 +478,7 @@ An object with the following properties:
 * `"zonemaster_engine"`: A string. The version number of the *Zonemaster Engine* used by the *RPC API daemon*.
 
 
-##### `"error"`
+#### `"error"`
 
 >
 > TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
@@ -514,7 +514,7 @@ Example response:
 ```
 
 
-##### `"result"`
+#### `"result"`
 
 An oject with the following property:
 
@@ -569,14 +569,14 @@ Example response:
 ```
 
 
-##### `"result"`
+#### `"result"`
 
 An oject with the following property:
 
 * `"languages"`: An array of [*language tags*][Language tag]. It is never empty.
 
 
-##### `"error"`
+#### `"error"`
 
 >
 > TODO: List all possible error codes and describe what they mean enough for
@@ -621,14 +621,14 @@ Example response:
 ```
 
 
-##### `"params"`
+#### `"params"`
 
 An object with the property:
 
 * `"hostname"`: A [*domain name*][Domain name], required. The hostname whose IP addresses are to be resolved.
 
 
-##### `"result"`
+#### `"result"`
 
 An oject with the following property:
 
@@ -643,7 +643,7 @@ An oject with the following property:
 >
 
 
-##### `"error"`
+#### `"error"`
 
 * If any parameter is invalid an error code of -32602 is returned. The `data` property contains an array of all errors, see [Validation error data].
 
@@ -724,7 +724,7 @@ Example response:
 >
 
 
-##### `"params"`
+#### `"params"`
 
 An object with the properties:
 
@@ -732,7 +732,7 @@ An object with the properties:
 * `"language"`: A [*language tag*][Language tag], optional, used for validation error messages
   translation, if not provided messages will be untranslated (in English).
 
-##### `"result"`
+#### `"result"`
 
 An object with the following properties:
 
@@ -740,7 +740,7 @@ An object with the following properties:
 * `"ds_list"`: A list of [*DS info*][DS info] objects representing delegation signer (DS record data) of the given [*domain name*][Domain name].
 
 
-##### `"error"`
+#### `"error"`
 
 * If any parameter is invalid an error code of -32602 is returned. The `data` property contains an array of all errors, see [Validation error data].
 
@@ -808,7 +808,7 @@ Example response:
 ```
 
 
-##### `"params"`
+#### `"params"`
 
 An object with the following properties:
 
@@ -827,7 +827,7 @@ An object with the following properties:
   translation, if not provided messages will be untranslated.
 
 
-##### `"result"`
+#### `"result"`
 
 An object with the following property:
 
@@ -844,7 +844,7 @@ The parameters that are compared when to determine if two requests are to be
 considered to be the same are `domain`, `ipv6`, `ipv4`, `nameservers`, `ds_info`
 and `profile`.
 
-##### `"error"`
+#### `"error"`
 
 * If any parameter is invalid an error code of -32602 is returned.
   The `data` property contains an array of all errors, see [Validation error data].
@@ -917,21 +917,21 @@ Example response:
 ```
 
 
-##### `"params"`
+#### `"params"`
 
 An object with the property:
 
 `"test_id"`: A [*test id*][Test id], required. The *test* to report on.
 
 
-##### `"result"`
+#### `"result"`
 
 An object with the following property:
 
 * `"progress"`: A [*progress percentage*][Progress percentage].
 
 
-##### `"error"`
+#### `"error"`
 
 >
 > TODO: List all possible error codes and describe what they mean enough for clients to know how react to them.
@@ -1015,7 +1015,7 @@ Example response:
 >
 
 
-##### `"params"`
+#### `"params"`
 
 An object with the following properties:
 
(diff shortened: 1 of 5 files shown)


matsduf:
I do think that the new names are better than the old ones, but is it worth the change to get better names?

#790 suggests that it should be possible to send deprecation messages with the result message, and that can sound appealing but would that really work? Who would read those messages? Do you envision that those messages are shown to the Zonemaster GUI users? Or how should those messages go to some maintainer?

pnax:
> I do think that the new names are better than the old ones, but is it worth the change to get better names?

Do we think this is a change we'd like to see at some point? Do we want to harmonize the API? Do we want to clean our codebase? I would answer yes to all.
Furthermore one of the point raised in #790 is our use of the "test" term that refers to several different things (a test case, a test of a zone…) , making it easy to mix things up.
I think it's important to keep our API and code base harmonized. Currently there is a lot of room for improvements (see #791 for example).

docs/API.md
```diff
@@ -459,22 +476,23 @@ Example response:
 ```
 
 
-#### `"result"`
+##### `"result"`
```
matsduf, line 479: I think this should be on level 4 (`####`) not 5 since it is a subheader of `### API method: system_versions` on level 3.

docs/API.md
```diff
@@ -1462,86 +1579,119 @@ If the `batch_id` is undefined the following error is returned:
...
-    }
-}
-```
 
+### API method: `profile_names`
 
-#### `"params"`
+**Deprecated**.
+Same as [`conf_profiles`][conf_profiles], but returns only the array of
+[*Profile names*][Profile name].
```
matsduf, line 1594: Unclear, what does "but returns only the array of Profile names" mean?

docs/API.md
```diff
@@ -1462,86 +1579,119 @@ If the `batch_id` is undefined the following error is returned:
 }
 ```
 
-## API method: `get_test_params`
 
-Return a normalized *params* objects of a *test*.
+## Deprecated
```
matsduf, line 1583: I think we should state when the old API names will be removed.

matsduf:
`job_progress` returns the status of ongoing job or completed job.
`batch_status` returns the status of ongoing batch or completed batch.

`job_progress` and `batch_progress` or `job_status` and `batch_status`?

pnax:
Good point, I like `*_status`, it's more generic. I'll provide an update.

docs/API.md
```diff
@@ -1282,7 +1396,9 @@ Example response:
 {
     "jsonrpc": "2.0",
     "id": 147559211348450,
-    "result": 8
+    "result": {
+      "id": 8
```
matsduf, line 1400: It would be clearer with "batch_id" instead of just "id". There are other "id". In method "batch_status" "batch_id" is used for the equivalent thing.

docs/API.md
```diff
@@ -1312,7 +1428,9 @@ The value of `"test_params"` is an object with the following properties:
 
 #### `"result"`
 
-A [*batch id*][Batch id].
+An object with the following property:
+
+* `"id"`: An integer. The [*batch id*][Batch id].
```
matsduf, line 1433: See above, "batch_id".

docs/API.md
```diff
@@ -1396,7 +1514,7 @@ Example request:
 {
     "jsonrpc": "2.0",
     "id": 147559211994909,
-    "method": "get_batch_job_result",
+    "method": "batch_status",
     "params": {"batch_id": "8"}
```
matsduf, line 1518: I suggest:
```
"params": {
    "batch_id": "8"
}
```

matsduf (changes requested):

docs/API.md
pnax: Rebased and updated to include this proposal.

docs/API.md
pnax: same

docs/API.md
pnax: done

docs/API.md
pnax: Note added.

matsduf:
E.g. method `start_domain_test` is deprecated, but it is also used in some `t` files (e.g. `t/lifecycle.t`). Other deprecated methods might be used in the `t` files. But the new methods are not added, which I think they should be. The deprecated methods could maybe be removed now, but it would be better if they are kept until they are removed.

pnax:
I planned to do this on a second stage, ie in a follow-up PR. If this is ok, could you approve this PR?

matsduf:
> I planned to do this on a second stage, ie in a follow-up PR. If this is ok, could you approve this PR?

That is also fine.

matsduf (dismissed):

docs/API.md
```diff
@@ -30,20 +30,35 @@
...
-* [API method: add_api_user](#api-method-add_api_user)
-* [API method: add_batch_job](#api-method-add_batch_job)
-* [API method: get_batch_job_result](#api-method-get_batch_job_result)
-* [API method: get_test_params](#api-method-get_test_params)
+* [API methods](#api-methods)
+  * [API method: system_versions](#api-method-system_versions)
+  * [API method: conf_profiles](#api-method-conf_profiles)
+  * [API method: conf_languages](#api-method-conf_languages)
+  * [API method: lookup_address_records](#api-method-lookup_address_records)
+  * [API method: lookup_delegatio_data](#api-method-lookup_delegation_data)
```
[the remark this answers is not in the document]
pnax: fixed

script/zonemaster_backend_rpcapi.psgi
```diff
@@ -59,76 +59,152 @@ my $handler = Zonemaster::Backend::RPCAPI->new( { config => $config } );
...
         action => "get_batch_job_result"
     };
+
+    connect "batch_status" => {
+        handler => $handler,
+        action => "batch_status"
+    };
 };
 
 if ($config->RPCAPI_enable_add_api_user) {
```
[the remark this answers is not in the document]
pnax: good point, I've created this new config parameter

script/zonemaster_backend_rpcapi.psgi
```diff
@@ -59,76 +59,152 @@ my $handler = Zonemaster::Backend::RPCAPI->new( { config => $config } );
...
         handler => $handler,
         action => "add_api_user"
     });
+    $router->connect("user_create", {
+        handler => $handler,
+        action => "user_create"
+    });
 }
 
 if ($config->RPCAPI_enable_add_batch_job) {
```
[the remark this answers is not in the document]
pnax: good point, I've created this new config parameter

lib/Zonemaster/Backend/RPCAPI.pm
```diff
@@ -271,37 +315,59 @@ sub start_domain_test {
...
 
-        $result = $self->{db}->create_new_test( $params->{domain}, $params, $self->{config}->ZONEMASTER_age_reuse_previous_test );
+        $job_id = $self->{db}->create_new_test( $params->{domain}, $params, $self->{config}->ZONEMASTER_age_reuse_previous_test );
     };
     if ($@) {
         handle_exception( $@ );
     }
 
+    my $result = {
+        id => $job_id
```
pnax, line 325: Update to `job_id => $job_id` (similar to what is done in `batch_create`)

matsduf (approved):
This is still fine.

Issue #1049, opened 2022-08-22
Update upgrade instruction with free disk space requirement


matsstralbergiis:
When upgrading the server, the instruction says that you should make a backup which in my case was 15GB. When running the patch-script the space needed seemed to be about 20 GB which was close to fill the disk.

Can we add an warning to makes sure that you should check that you have at least 150% of the size of the backup file free on the disk before proceeding with the patch-script?

mattias-p:
Hi Mats, thanks for reporting!

I couldn't find the instruction you're referring to. Do you have a link?

matsstralbergiis:
This is the instruction I followed: https://github.com/zonemaster/zonemaster-backend/blob/master/docs/upgrade/upgrade_zonemaster_backend_ver_9.0.0.md

mattias-p:
Thanks for clarifying. Yeah, this seems like a good idea.

PR #1065, opened 2022-11-30
Add note about free space before dumping database

pnax:
## Purpose

During upgrade, there is a suggestion to dump the database, depending on the database size this could require some disk space. This is about adding a note to the administrator to make sure there is sufficient free space on the destination disk.

## Context

Addresses #1049 

## Changes

Upgrade document to v9.0.0.

## How to test this PR

Documentation only.


matsduf (approved):

pnax:
Rereading the issue, I think the issue is more about the space by the database engine while altering the structure than the dump size. I need to test a database migration and look for disk usage.

matsduf:
@pnax, should this be included in v2022.2?

PR #1083, opened 2023-02-24
RPCAPI apply new methods in code, scripts and tests

pnax:
## Purpose

Second stage in the update of the RPCAPI methods. This PR updates the code and tests to use the new methods from #1054.

## Context

Follow up #1054 

## Changes

* Scripts
* Unit tests
* Deprecate RPCAPI.enable_add_api_user and RPCAPI.enable_add_batch_job properties in the INI file. Emit warning if used.
* Add unit tests
* Documentation

## How to test this PR

Unit tests should pass.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1582,77 +1582,78 @@ If the `batch_id` is undefined the following error is returned:
 ### API method: `version_info`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`system_versions`][API system_versions].
+Replaced by [`system_versions`][API system_versions].
 
 
 ### API method: `profile_names`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`conf_profiles`][API conf_profiles].
+Replaced by [`conf_profiles`][API conf_profiles].
 
 
 ### API method: `get_language_tags`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`conf_languages`][API conf_languages].
+Replaced by [`conf_languages`][API conf_languages].
 
 ### API method: `get_host_by_name`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`lookup_address_records`][API lookup_address_records].
+Replaced by [`lookup_address_records`][API lookup_address_records].
 
 
 ### API method: `get_data_from_parent_zone`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`lookup_delegation_data`][API lookup_delegation_data].
+Replaced by [`lookup_delegation_data`][API lookup_delegation_data].
 
 
 ### API method: `start_domain_test`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`job_create`][API job_create].
+Replaced by [`job_create`][API job_create].
 
 
 ### API method: `test_progress`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`job_status`][API job_status].
+Replaced by [`job_status`][API job_status].
 
 
 ### API method: `get_test_results`
 
 **Deprecated** (planned removal: v2024.1).
+Replaced by [`job_results`][API job_results].
 
 
 ### API method: `get_test_history`
 
 **Deprecated** (planned removal: v2024.1).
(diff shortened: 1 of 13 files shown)


matsduf:
Is this a "V-Minor" or "V-Major"? I guess it depends on if `zmb` is seen to be part of the API or not.

pnax:
> Is this a "V-Minor" or "V-Major"? I guess it depends on if zmb is seen to be part of the API or not.

I'd say "V-Minor". zmb explicitly states:
```
This interface is unstable and will change in a future release.
```
So I don't think it's part of the API yet.

docs/API.md
```diff
@@ -344,9 +344,9 @@ with this type, it returns the following error message:
     }
 }
 ```
-The error code is "009" (as above) if method [`start_domain_test`][API start_domain_test]
+The error code is "009" (as above) if method [`job_create`][API job_create]
```
matsduf, line 347: "[API job_create]" is not defined.

docs/API.md
```diff
@@ -1226,7 +1225,7 @@ An object with the following properties:
 
 ### API method: `user_create`
 
-In order to use the [`add_batch_job`][add_batch_job] method a
+In order to use the [`batch_create`][batch_create] method a
 [*username*][Username] and its [*api key*][Api key] must be added by this method.
 
 This method is not available if [`RPCAPI.enable_add_api_user`][RPCAPI.enable_add_api_user]
```
matsduf, line 1230: Should not this be changed to the new name?

On line 1344:
```
    "message": "Call to \"add_api_user\" method not permitted from a remote IP"                                                                                   
```

On line 1354:
```
    "message": "Procedure 'add_api_user' not found"                                                                                                               
```

On lines 1367-1369:
```
A [*username*][Username] and its [*api key*][Api key] can be added with the
[`add_api_user`][add_api_user] method. A [*username*][Username] can only have
one un-finished *batch* at a time.
```

matsduf (changes requested):

matsduf:
> I'd say "V-Minor". zmb explicitly states:
> 
> ```
> This interface is unstable and will change in a future release.
> ```

That is true.

It is time to make the interface of zmb stable (in a different PR).

docs/API.md
pnax: indeed, this should be fixed now

docs/API.md
pnax: good catch, I've fixed it

docs/API.md
```diff
@@ -1584,42 +1582,42 @@ If the `batch_id` is undefined the following error is returned:
...
 **Deprecated** (planned removal: v2024.1).
-Same as [`conf_profiles`][conf_profiles].
+Same as [`conf_profiles`][API conf_profiles].
 
 
 ### API method: `get_language_tags`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`conf_languages`][conf_languages].
+Same as [`conf_languages`][API conf_languages].
```
matsduf, line 1597: The new `get_language_tags` does not give the same result as the old `conf_languages`. Should it really say "same as" since it is not totally the same? Maybe "replaced by" or "use instead"?

matsduf (changes requested):

docs/API.md
pnax: Updated using "Replaced by".

docs/API.md
```diff
@@ -1584,42 +1582,42 @@ If the `batch_id` is undefined the following error is returned:
...
 
 
 ### API method: `test_progress`
 
 **Deprecated** (planned removal: v2024.1).
-Same as [`job_status`][job_status].
+Same as [`job_status`][API job_status].
 
 
 ### API method: `get_test_results`
```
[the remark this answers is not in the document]
pnax: Line added. Good catch.

matsduf (approved):
This should be fine now.

mattias-p:
I've tested this and found the problems also reported in #1105. With the changes in #1109 this works as advertised.

I also created zonemaster/zonemaster#1159 based on what I learned testing this.

PR #1089, opened 2023-03-03
Unit test for PO files

pnax:
## Purpose

New unit test to verify that PO files are correct regarding some rules:
* no fuzzy marks
* all PO files are tidied
* PO files msgid/msgstr format should be consistent 

## Context

Same checks as in Zonemaster-Engine and Zonemaster-CLI (https://github.com/zonemaster/zonemaster-cli/pull/314)

## Changes

new t/po-files.t

## How to test this PR

Unit test should pass.

### DIFF
--- a/share/GNUmakefile
+++ b/share/GNUmakefile
@@ -33,6 +33,9 @@ new-po: extract-pot
 		-e 's/^("Content-Type:.+charset=)CHARSET/$${1}UTF-8/;' $(POLANG).po
 	@perl -ni -e 'print unless /^#( |$$)/' $(POLANG).po
 
+# Check the msgid/msgstr pair for some inconsistencies between them in the
+# selected PO file and report on standard error any errors found. The PO file
+# is not updated.
 check-po:
 	@for f in $(POFILES) ; do msgfmt -c $$f ; done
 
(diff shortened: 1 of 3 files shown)


share/GNUmakefile
```diff
@@ -33,6 +33,9 @@ new-po: extract-pot
 		-e 's/^("Content-Type:.+charset=)CHARSET/$${1}UTF-8/;' $(POLANG).po
 	@perl -ni -e 'print unless /^#( |$$)/' $(POLANG).po
 
+check-po:
+	@for f in $(POFILES) ; do msgfmt -c $$f ; done
+
```
matsduf, line 38: Please add a comment that describes what `check-po` does. Just 1-2 sentences.

matsduf (commented):
The Zonemaster-Engine `/util/check-msg-args` is not available here, and currently there are no arguments in the msgids, but that might change.

I suggest that we have the script in all three environments where we have PO files, Engine, CLI and Backend, to make it consistent and the instructions can be the same.

I also suggest that `check-msg-args` script is put in the `share` directory to make it easier for the translator to execute. It is really only meaningful for the translation work. (Moving/putting it into `share` should apply for all three environments.)

matsduf (approved):

Issue #1084, opened 2023-02-27
Disambiguate usage of the term "test" by using "job"


ghost:
With the new API (#1054) , former methods about "tests" are now called "job_*". We can now disambiguate our usage of the "test" term which refers to a lot of thing in our Zonemaster context (a testcase, a unit test, a test of a zone). Now we can refer to a job as being a task performed by the TestAgent that tests a zone.

For instance in some places `test_id` can be replaced with `job_id` in the new methods' documentation and implementation.

## Stabilization
[Tracking issue](https://github.com/zonemaster/zonemaster/issues/1175)

### Affected interfaces
Incompatibly changed:
* job_status
* job_results
* domain_history
* batch_create
* batch_status

### Dependencies
None

### Overlapping proposals
None

mattias-p:
There are also places where we use `id`, `hash_id`, `finished_test_ids`, `test_params` and `frontend_params` for things related to jobs. These should also be taken care of at the same time.

mattias-p:
We should limit the scope of this issue to exclude the batch methods and handle those parameters separately in order to avoid conflicts with other issues.

I suggest we rename parameter names according to this table.

|Old parameter name|New parameter name|Affected methods|
|---|---|---|
|test_id|job_id|job_status, job_params|
|id|job_id|job_params|
|frontend_params|params|domain_history|

mattias-p:
Fixed by #1155.

PR #1090, opened 2023-03-13
Uniform use of job

pnax:
## Purpose

The API has been updated with new methods `job_*` (#1054, #1083) . A _test_ can now be referred as a _job_. It would then make sense to rename `test_id` to `job_id`. This PR does that for the new API methods.

## Context

#1084 

## Changes

* documentation (API.md)
* tests (t/test01.t and t/lifecycle.t)
* implementation (RPCAPI.pm, DB.pm, SQLite.pm, MySQL.pm, PostgreSQL.pm)
* deprecate (planned removal v2024.1) `test_id` API documentation

## How to test this PR

Unit tests should pass.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -27,6 +27,8 @@
   * [Severity level](#severity-level)
   * [Test id](#test-id)
   * [Test result](#test-result)
+  * [Job id](#job-id)
+  * [Job result](#job-result)
   * [Timestamp (deprecated)](#timestamp-deprecated)
   * [Timestamp](#timestamp)
   * [Username](#username)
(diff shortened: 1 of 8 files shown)


matsduf:
What is a "job" and what is a "test"? I think it is unclear in the API document. It starts by saying "Health checks are called tests in Zonemaster lingo" so what we store in the database are those "test results" or "job results"? The job is there to run a test and that would result in a "test result". If so, "test-id" is more reasonable than "job-id". The test remains when the job has completed.

I think this should be clarified in the API document.

matsduf (changes requested):
What is a "job" and what is a "test"? I think it is unclear in the API document. It starts by saying "Health checks are called tests in Zonemaster lingo" so what we store in the database are those "test results" or "job results"? The job is there to run a test and that would result in a "test result". If so, "test-id" is more reasonable than "job-id". The test remains when the job has completed.

I think this should be clarified in the API document.

pnax:
> What is a "job" and what is a "test"?

The way I see it, a job runs some tests. The tests are the one from the specification.

matsduf:
I looked through the API document, and I do not think that "job result" is a good term. A job is run and a test result comes out. If we communicate that it is a Job Result, then it is less understandable than Test Result. But it is fine to say that a job has been created.

If you look for a test result, what should the identifier be? "Test ID" or "Job ID"? Here I think Test ID is probably easier to understand, but Job ID will work.

If the outcome of a Job is a Test Result, then the method should rather be `job_outcome` or `test_result` than `job_results`.

When the user runs (orders) a domain test, a Job is created and then there will be Test Result to present to the user.

pnax:
Closing this PR. The renaming might happen once the new RPCAPI is stabilized.

PR #1092, opened 2023-03-16
New result entries table

pnax:
## Purpose

Move the `test_results.results` json array into a dedicated table `result_entries`.

## Context

Replaces #856 

## Changes

Adds a new table `result_entries`:

```
                                     Table "public.result_entries"
  Column   |          Type          | Collation | Nullable |                  Default                   
-----------+------------------------+-----------+----------+--------------------------------------------
 id        | integer                |           | not null | nextval('result_entries_id_seq'::regclass)
 hash_id   | character varying(16)  |           | not null | 
 level     | log_level              |           | not null | 
 module    | character varying(255) |           | not null | 
 testcase  | character varying(255) |           | not null | 
 tag       | character varying(255) |           | not null | 
 timestamp | real                   |           | not null | 
 args      | json                   |           | not null | 
Indexes:
    "result_entries_pkey" PRIMARY KEY, btree (id)
    "result_entries__hash_id" btree (hash_id)
    "result_entries__level" btree (level)
```

* The `get_test_result` and `get_test_history` methods are modified to use the new table
* `DB::test_result` is now only a getter, all write operations use the new database methods `add_result_entry` and `add_result_entries`

## How to test this PR

* Create a few tests
* Get the history / results
It should work the same way as it was

Credits to @blacksponge for most of the work.


matsduf:
Isn't i better to store the hash_id as number?

pnax:
> Isn't i better to store the hash_id as number?

TBH I've never thought about that. This uses the same datatype as in the `test_results` table. Maybe we could consider updating the `hash_id` datatype.

lib/Zonemaster/Backend/Config.pm
```diff
@@ -809,7 +809,7 @@ sub new_PM {
                 }
             }
         },
-        1
+        0.125
```
blacksponge, line 812: Isn't that unrelated to the purpose of this PR?

pnax: Indeed. Would you suggest to do this change in a different PR?

matsduf: Yes, and there some comment what the change is for. I saw it in #1094 too.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -761,6 +790,50 @@ sub to_iso8601 {
...
+            $json->encode( $m->args // {} ),
+        ];
+
+        push @records, $r;
+    }
+    my $query_values = join ", ", ("(?, ?, ?, ?, ?, ?, ?)") x @records;
+    my $query = "INSERT INTO result_entries (hash_id, level, module, testcase, tag, timestamp, args) VALUES $query_values";
+    my $sth = $self->dbh->prepare($query);
+    $sth = $sth->execute(map { @$_ } @records);
+}
```
matsduf, line 835: These two methods, `add_result_entry()` and `add_result_entries()` do the same thing, as far as I can see. Just that the first does it one entry and the other on multiple. Is there a reason for not using just one method, e.g. requiring "entry" to always be a hash reference even if there is only one member?

matsduf (commented):
Will the change decrease or increase the load of the database when running a batch?

marc-vanderwal (commented):
Apart from the only comment I have, it looks good to me.

lib/Zonemaster/Backend/DB.pm
pnax: Thanks for the review and suggestion. I've updated the code to make it variadic as suggested.

lib/Zonemaster/Backend/Config.pm
pnax: This has been removed from this PR.

marc-vanderwal (dismissed):
Looks good to me.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -241,6 +242,20 @@ sub test_progress {
...
+                    UPDATE test_results
+                    SET progress = ?,
+                        ended_at = ?
+                    WHERE hash_id = ?
+                ],
+                undef,
+                $progress,
+                $self->format_time( time() ),
+                $test_id,
+            );
```
matsduf, line 257: I do not feel it is correct that a test with progress=0 could be set directly to 100. What use case is that?

Secondly, there is no check that the update really was done. There should be exactly one record updated.

matsduf (dismissed):
Before we implement this change we should understand the gains and losses of the change. At least Mysql and Postgresql have support for querying into the JSON blob as if it consisted of several fields.

lib/Zonemaster/Backend/DB.pm
pnax: > I do not feel it is correct that a test with progress=0 could be set directly to 100. What use case is that?

Could you elaborate when would this occur?
I'm not sure I see the issue you are raising. However there is still a need to update the `progress` value to `100` once the job is completed and the data is stored. Now that there are 2 tables, this should be done in an independent manner.

> There should be exactly one record updated.

That's why there is a `WHERE hash_id = ?` clause. The hash_id should be unique in the database.

pnax (commented):
> Before we implement this change we should understand the gains and losses of the change. At least Mysql and Postgresql have support for querying into the JSON blob as if it consisted of several fields.

Pro: you could use SQL syntax to query in the table (no need for specific JSON support).
Con: you have to JOIN the tables.

If you see other gain/loss, please share them.

lib/Zonemaster/Backend/DB.pm
matsduf: I think the code should prevent an error where a test with progress = 0 is updated.

I also think the code should verify that a test really was updated, i.e. that the number of updated records is one not zero.

matsduf:
Both Mysql (Mariadb) and Postgresql have good support for JSON so how much gain is it? Are there any performance differences in saving the result in a table instead of in JSON? Are there any performance differences when reading the result?

matsduf:
@pnax, have estimated how time it will take to migrade the database in the zonemster.net server?

pnax:
> have estimated how time it will take to migrade the database in the zonemster.net server?

Yes but I can't find the figure again. If I recall correctly, this was around 2 to 3 hours.

lib/Zonemaster/Backend/DB.pm
pnax: This could be an idea. However, since this is not directly related to this PR, I think it would be better to make such change in an extra PR to not mix things.

matsduf: It looks like it is new code with this PR. But there are more places where the update is done without checking the result.

matsduf:
> Are there any performance differences when reading the result?

@matsstralbergiis will try to measure the difference.

@pnax, have you compared load of writing with old and new code?

pnax:
> Both Mysql (Mariadb) and Postgresql have good support for JSON so how much gain is it?

See https://github.com/zonemaster/zonemaster-backend/pull/856#issuecomment-1457814137, https://github.com/zonemaster/zonemaster-backend/pull/1092#pullrequestreview-1526628097. Another gain would be to ease the use of another database engine (for instance Clickhouse, see https://github.com/zonemaster/zonemaster-backend/pull/1094).

Regarding the performance gain, I need to redo some tests. (I thought I had done some a few months back but I can't find my results anymore.)

matsduf:
> See [#856 (comment)](https://github.com/zonemaster/zonemaster-backend/pull/856#issuecomment-1457814137) and [#1092 (review)](https://github.com/zonemaster/zonemaster-backend/pull/1092#pullrequestreview-1526628097). Another gain would be to ease the use of another database engine (for instance Clickhouse, see #1094).

The comment in #856 was mine, but today I am better informed and understand that it actually works to let the database engine look into JSON. But it still might be a good idea to use a separate table. It is conceptually simpler. And possibly better support for other engines.

> Regarding the performance gain, I need to redo some tests. (I thought I had done some a few months back but I can't find my results anymore.)

Thanks!

matsstralbergiis:
I have copied the 'test_result' table to a MySQL instance running in my Mac. The database contains 208 149 tests.

Then I created the new 'result_entries' table (I copied the create command from the MySQL.pm in this PR.) and extracted the data from the JSON blob in the 'test_result' table into the nwe table. It now contains 14 518 570 records.

After that I performed two queries that I believe gets the same result.

One to the 'old' format:
```
SELECT tr.batch_id, tr.hash_id, tr.domain, result.tag, count(result.tag), result.level, result.module, result.testcase FROM    test_results tr,   JSON_TABLE(     CAST(CONVERT(results USING utf8) AS JSON),     "$[*]" COLUMNS(        tag VARCHAR(40) PATH '$.tag' ERROR ON ERROR,       level VARCHAR(10) PATH '$.level' ERROR ON ERROR,       module VARCHAR(20) PATH '$.module' ERROR ON ERROR,       testcase VARCHAR(20) PATH '$.testcase' ERROR ON ERROR     )   ) as result WHERE batch_id = 10 GROUP BY tr.hash_id, tr.domain, result.tag, result.level, result.module, result.testcase ORDER BY tr.hash_id limit 10;
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
| batch_id | hash_id          | domain               | tag                        | count(result.tag) | level | module       | testcase       |
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | ADDRESSES_MATCH            |                 1 | INFO  | CONSISTENCY  | CONSISTENCY05  |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | ARE_AUTHORITATIVE          |                 1 | INFO  | DELEGATION   | DELEGATION04   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | AXFR_FAILURE               |                 8 | INFO  | NAMESERVER   | NAMESERVER03   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B01_CHILD_FOUND            |                 1 | INFO  | BASIC        | BASIC01        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B01_PARENT_FOUND           |                 1 | INFO  | BASIC        | BASIC01        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B02_AUTH_RESPONSE_SOA      |                 1 | INFO  | BASIC        | BASIC02        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CAN_BE_RESOLVED            |                 1 | INFO  | NAMESERVER   | NAMESERVER06   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CASE_QUERIES_RESULTS_OK    |                 1 | INFO  | NAMESERVER   | NAMESERVER09   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CHILD_DISTINCT_NS_IP       |                 1 | INFO  | DELEGATION   | DELEGATION02   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CN04_IPV4_DIFFERENT_PREFIX |                 1 | INFO  | CONNECTIVITY | CONNECTIVITY04 |
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
10 rows in set, 1 warning (52.8186 sec)
```

And one to the new format:

```
SELECT tr.batch_id, tr.hash_id, tr.domain, result.tag, count(result.tag), result.level, result.module, result.testcase FROM result_entries result LEFT JOIN test_results tr ON result.hash_id = tr.hash_id WHERE batch_id = 10 GROUP BY tr.hash_id, tr.domain, result.tag, result.level, result.module, result.testcase ORDER BY tr.hash_id limit 10;
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
| batch_id | hash_id          | domain               | tag                        | count(result.tag) | level | module       | testcase       |
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | ADDRESSES_MATCH            |                 1 | INFO  | CONSISTENCY  | CONSISTENCY05  |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | ARE_AUTHORITATIVE          |                 1 | INFO  | DELEGATION   | DELEGATION04   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | AXFR_FAILURE               |                 8 | INFO  | NAMESERVER   | NAMESERVER03   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B01_CHILD_FOUND            |                 1 | INFO  | BASIC        | BASIC01        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B01_PARENT_FOUND           |                 1 | INFO  | BASIC        | BASIC01        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | B02_AUTH_RESPONSE_SOA      |                 1 | INFO  | BASIC        | BASIC02        |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CAN_BE_RESOLVED            |                 1 | INFO  | NAMESERVER   | NAMESERVER06   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CASE_QUERIES_RESULTS_OK    |                 1 | INFO  | NAMESERVER   | NAMESERVER09   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CHILD_DISTINCT_NS_IP       |                 1 | INFO  | DELEGATION   | DELEGATION02   |
|       10 | 000081efd2d471e0 | xn--nsbyslott-v2a.se | CN04_IPV4_DIFFERENT_PREFIX |                 1 | INFO  | CONNECTIVITY | CONNECTIVITY04 |
+----------+------------------+----------------------+----------------------------+-------------------+-------+--------------+----------------+
10 rows in set (25.3378 sec)
```

The results are consistent when repeating the queries.

Maybe it is possible to optimize the queries to speed it up. Feel free to suggest improvements.

share/patch/patch_db_zonemaster_backend_ver_11.0.0.pl
```diff
@@ -0,0 +1,232 @@
...
+    my $db = Zonemaster::Backend::DB::MySQL->from_config( $config );
+    my $dbh = $db->dbh;
+
+    $dbh->{AutoCommit} = 0;
+
+    try {
+        $dbh->do(
+            "CREATE TABLE IF NOT EXISTS result_entries (
+                id integer AUTO_INCREMENT PRIMARY KEY,
+                hash_id VARCHAR(16) not null,
```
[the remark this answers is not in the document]
pnax: I didn't think about it. This could work. We could also remove the `id` field and make `hash_id` the PRIMARY KEY in `test_results`. Regarding performance gain, [this could depend on the database engine](https://stackoverflow.com/questions/39388912/does-adding-a-foreign-key-to-an-indexed-column-boost-performance).
I'll provide an update with this suggestion.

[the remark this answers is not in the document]
mattias-p: I believe joins are faster when using integer keys so it might be a good idea to keep the id field. I guess CHAR(16) is 128 bits, and maybe DBMSs would be able to optimize such joins, but I believe it's best to keep primary keys at 64 bits or smaller.

matsduf:
If databases are better at handling integer keys, wouldn't it be better to convert the hash_id into an integer instead?

marc-vanderwal:
Any x86_64 CPU with the [SSE4.2 instruction set extensions][1] (and running a DBMS that has been compiled to support these instructions) can compare strings with the same efficiency as integers (and strings can be seen as arrays of large integers, in a way). It seems to me that discussing the data type of primary keys is a case of premature optimization. There is [evidence][2] showing that, for MySQL at least, the performance penalty is quite negligible. I suggest we keep the primary key as `VARCHAR(16)` for now.

[1]: https://en.wikipedia.org/wiki/SSE4#SSE4.2
[2]: https://stackoverflow.com/questions/332300/is-there-a-real-performance-difference-between-int-and-varchar-primary-keys

pnax:
* I've rebased on top of `develop`
* I've addressed the changes discussed during F2F (2023-09)
   * remove the `id` field in `result_entries`
   * use a foreign key for `hash_id` in `result_entries`
   * add a new table `log_level`
   * use a foreign key for `level` in `result_entries`
   * change the JSON type of `args` to JSONb in PostgreSQL

Please re-review.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -241,6 +242,20 @@ sub test_progress {
                 $test_id,
             );
         }
+        elsif ( $progress == 100 ) {
```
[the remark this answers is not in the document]
pnax: As discussed during F2F (2023-09) this PR should be merged after #1121 so that it could be updated accordingly.

share/patch/patch_db_zonemaster_backend_ver_11.0.0.pl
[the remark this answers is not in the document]
pnax: I've added the foreign key referencing the `hash_id` from table `test_results`, and removed the`id` field.

matsduf:
Is field "hash_id" in table "result_entries" the same "hash_id" that identifies the test? If so, does that mean that the table has no unique identifier?

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -115,6 +115,68 @@ sub create_schema {
...
+        );
+    }
+
+    ####################################################################
+    # RESULT ENTRIES
+    ####################################################################
+    $dbh->do(
+        "CREATE TABLE IF NOT EXISTS result_entries (
+            hash_id VARCHAR(16) NOT NULL,
+            level VARCHAR(15) NOT NULL,
```
matsduf, line 154: Wouldn't it be more efficient to store level directly as an integer so that you could ask for all messages where level is, say, greater than 2? As I understand it now, you have to go via the "log_level" table.

matsduf:
Do you have an estimation of how long time it will take to patch the zonemaster.net database?

pnax:
> Is field "hash_id" in table "result_entries" the same "hash_id" that identifies the test?

Yes.

>If so, does that mean that the table has no unique identifier?

Yes, there will be several rows with the same `hash_id` in this table (one per msg_tag)

lib/Zonemaster/Backend/DB/MySQL.pm
matsduf: I think so. What do others think?

pnax:
> Do you have an estimation of how long time it will take to patch the zonemaster.net database?

I ran the migration script on a copy of the current database (zonemaster.net). It took around 4 hours 15 minutes on my setup.

Some figures on the database
date: 2023-09-27
database: PostgreSQL
number of tests: 1886151
db size (before migration): 7.3 GB
db size (after migration): 22 GB

matsduf:
> > Do you have an estimation of how long time it will take to patch the zonemaster.net database?
> 
> I ran the migration script on a copy of the current database (zonemaster.net). It took around 4 hours 15 minutes on my setup.

Thanks. That sounds manageable.

pnax:
I've rebased on top of develop to integrate the changes from #1121 and address a comment from @mattias-p. I think I should have then addressed all comments/remarks. Please re-review.

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
pnax: I've addressed this with a new method `set_test_completed()` that transition from `RUNNING` to `COMPLETED`.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -10,14 +10,18 @@ use JSON::PP;
 use Scalar::Util qw( blessed );
 use File::Slurp;
 use Locale::TextDomain qw[Zonemaster-Backend];
+use Log::Any qw( $log );
```
blacksponge, line 13: Module required but not used.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -132,7 +138,23 @@ sub run {
         }
     }
 
-    $self->{_db}->store_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
+    #Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_duration",  $callback_duration * 1000);
```
blacksponge, line 141: Remove commented code.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -132,7 +138,23 @@ sub run {
...
+    my @entries = grep { $_->numeric_level >= $numeric{INFO} } @{ Zonemaster::Engine->logger->entries };
+
+    Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_add_result_entry_grep_duration",  tv_interval($start_time_2) * 1000);
+
+    $self->{_db}->add_result_entries( $test_id, @entries);
+
+    my $callback_add_result_entry_duration = tv_interval($start_time_2);
+    Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_add_result_entry_duration", $callback_add_result_entry_duration * 1000);
+
+    #$log->debug("Callback timing for $test_id: $callback_duration / $callback_add_result_entry_duration ");
```
blacksponge, line 155: Remove commented code.

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -132,7 +138,23 @@ sub run {
...
 
-    $self->{_db}->store_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
+    #Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_duration",  $callback_duration * 1000);
+
+    my $start_time_2 = [ gettimeofday ];
+
+    # TODO: Make minimum level configurable
+    my @entries = grep { $_->numeric_level >= $numeric{INFO} } @{ Zonemaster::Engine->logger->entries };
+
+    Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_add_result_entry_grep_duration",  tv_interval($start_time_2) * 1000);
```
blacksponge, line 148: Maybe rename to `log_callback_add_result_entry_filter_duration`

lib/Zonemaster/Backend/TestAgent.pm
```diff
@@ -132,7 +138,23 @@ sub run {
         }
     }
 
-    $self->{_db}->store_results( $test_id, Zonemaster::Engine->logger->json( 'INFO' ) );
+    #Zonemaster::Backend::Metrics::timing("zonemaster.testagent.log_callback_duration",  $callback_duration * 1000);
+
+    my $start_time_2 = [ gettimeofday ];
```
blacksponge, line 143: Maybe rename to a more explicit name, like `insert_result_start_time`

lib/Zonemaster/Backend/DB.pm
```diff
@@ -775,44 +833,41 @@ sub select_unfinished_tests {
...
+        {
+            level     => "CRITICAL",
+            module    => "BACKEND_TEST_AGENT",
+            testcase  => "",
+            tag       => "TEST_DIED",
+            args      => {},
+            timestamp => $self->get_relative_start_time($hash_id)
+        }
+    );
+    #$msg = bless( $msg, "Zonemaster::Engine::Logger::Entry" );
```
blacksponge, line 869: Remove commented code.

blacksponge (commented):
Nice work :) Just a few cleanup remarks.

lib/Zonemaster/Backend/TestAgent.pm
pnax: Done. And all other comments as well.

lib/Zonemaster/Backend/TestAgent.pm
pnax: Suggestion applied.

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
```diff
@@ -0,0 +1,163 @@
...
+
+            next unless $results;
+
+            my @records;
+            my $entries = $json->decode( $results );
+
+            foreach my $m ( @$entries ) {
+                my $r = [
+                    $hash_id,
+                    $m->{level},
```
blacksponge, line 54: Testing the patch script I get a foreign key constraint error, I guess this should be `$levels{$m->{level}}` with `levels` defined as `%levels = Zonemaster::Engine::Logger::Entry->levels()`.

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
```diff
@@ -0,0 +1,163 @@
...
+
+    my ( $row_total ) = $dbh->selectrow_array( 'SELECT count(*) FROM test_results' );
+    print "count: $row_total\n";
+
+    # depending on the resources available to select all data in database
+    # update $row_count to your needs
+    my $row_count = 50000;
+    my $row_done = 0;
+    while ( $row_done < $row_total ) {
+        print "row_done/row_total: $row_done / $row_total\n";
```
blacksponge, line 40: Can you also add this line when the loop exits so we now that all records have been processed.

blacksponge (commented):
I noticed an issue in the migration script, beside it I tested creating a test and querying for the history without issue.

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
pnax: Good point. I forgot to update the script when changing the value stored. I'll provide an update.

pnax: Fixed

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
pnax: line added

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -101,6 +101,61 @@ sub create_schema {
...
+            "
+        );
+    }
+
+    ####################################################################
+    # RESULT ENTRIES
+    ####################################################################
+    $dbh->do(
+        'CREATE TABLE IF NOT EXISTS result_entries (
+            hash_id VARCHAR(16) NOT NULL,
```
mattias-p, line 138: Why not have the foreign key referring to `test_results.id` instead of `test_results.hash_id`? (see comment above)

lib/Zonemaster/Backend/DB/SQLite.pm
```diff
@@ -96,6 +99,62 @@ sub create_schema {
...
+            "
+        );
+    }
+
+    ####################################################################
+    # RESULT ENTRIES
+    ####################################################################
+    $dbh->do(
+        'CREATE TABLE IF NOT EXISTS result_entries (
+            hash_id VARCHAR(16) NOT NULL,
```
mattias-p, line 137: Why not have the foreign key referring to `test_results.id` instead of `test_results.hash_id`? (see comment above)

mattias-p (approved):
Looks really good to me! I only have a few remarks.

lib/Zonemaster/Backend/DB/MySQL.pm
```diff
@@ -115,6 +115,68 @@ sub create_schema {
...
+            "
+        );
+    }
+
+    ####################################################################
+    # RESULT ENTRIES
+    ####################################################################
+    $dbh->do(
+        "CREATE TABLE IF NOT EXISTS result_entries (
+            hash_id VARCHAR(16) NOT NULL,
```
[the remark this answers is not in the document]
pnax: I wanted to avoid relying on the `id` field because auto incremental key is not supported in Clickhouse (a distributed database) and I wanted to keep the database internal as close as possible between each engine. Having said that, I realize it might be a little bit silly not to rely on specific features from each database engine.

> I'm sure that would be more efficient.

Maybe that won't be that bad since there is an index on `hash_id`. But to be honest I have no idea which one is better, I've not performed any test of this.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -945,6 +999,33 @@ sub to_iso8601 {
...
+            $e->testcase,
+            $e->tag,
+            $e->timestamp,
+            $json->encode( $e->args // {} ),
+        ];
+
+        push @records, $r;
+    }
+    my $query_values = join ", ", ("(?, ?, ?, ?, ?, ?, ?)") x @records;
+    my $query = "INSERT INTO result_entries (hash_id, level, module, testcase, tag, timestamp, args) VALUES $query_values";
```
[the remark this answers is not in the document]
pnax: Good idea. Just a question on how to do this. I was thinking to add something like:
```perl
    if ( $current_state ne $TEST_RUNNING ) {
        die Zonemaster::Backend::Error::Internal->new( reason => 'cannot store results to a non-running test' );
    }
```
but I think it won't be sufficient. What if the test's state is updated by another process between the check and the insert?

matsduf (commented):
This looks fine to me.

lib/Zonemaster/Backend/DB/MySQL.pm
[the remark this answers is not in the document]
mattias-p: > Maybe that won't be that bad since there is an index on hash_id.

Yeah, we could fix it later if it turns out it contributes to some bottleneck. Another thing that feels sub-optimal is that we use VARCHAR instead of CHAR for the hashes. But that probably doesn't make a huge difference either.

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
mattias-p: Yeah, exactly. I made a proof of concept for how to make it atomic. But I'm not sure about its scalability. Maybe it's great, I don't know.

```perl
#!/usr/bin/env perl
use strict;
use warnings;
use 5.014;
use utf8;

use DBI;

my $dbh = DBI->connect( "dbi:SQLite:dbname=:memory:", "", "" );

sub setup {
    $dbh->do( '
        CREATE TABLE test_results (
            id       INTEGER,
            progress INTEGER
        )
    ' );
    $dbh->do( '
        CREATE TABLE result_entries (
            test_id INTEGER,
            text    VARCHAR(20)
        )
    ' );
    $dbh->do( 'INSERT INTO test_results VALUES (1234, 0)' );
    $dbh->do( 'INSERT INTO test_results VALUES (2345, 50)' );
    $dbh->do( 'INSERT INTO test_results VALUES (3456, 100)' );
}

sub insert {
    my ( $id, @messages ) = @_;

    my $prefix    = 'INSERT INTO result_entries ';
    my $separator = ' UNION ';
    my $select    = 'SELECT id, ? FROM test_results WHERE id = ? AND progress > 0 AND progress < 100';
    my $query     = $prefix . join $separator, ( $select ) x scalar @messages;
    my $sth       = $dbh->prepare( $query );

    my @params;
    for my $message ( @messages ) {
        push @params, $message, $id;
    }

    my $rv = $sth->execute( @params );
    if ( $rv && $rv == 0 ) {
        die "illegal state transition";
    }
}

sub show {
    my ( $table ) = @_;
    my $rows = $dbh->selectall_arrayref( "SELECT * FROM $table" );
    say "Dump of $table:";
    for my $row ( @$rows ) {
        say join ', ', @{$row};
    }
}

my $id       = shift @ARGV;
my @messages = @ARGV;

setup();
show( 'test_results' );

say 'Inserting...';
insert( $id, @messages );

show( 'result_entries' );
```
```sh
$ ./poc 1234 aaa bbb ccc                                                                  
Dump of test_results:
1234, 0
2345, 50
3456, 100
Inserting...
illegal state transition at ./poc line 45.
```
```sh
$ ./poc 2345 aaa bbb ccc                                                                  
Dump of test_results:
1234, 0
2345, 50
3456, 100
Inserting...
Dump of result_entries:
2345, aaa
2345, bbb
2345, ccc
```

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
blacksponge: works fine now!

blacksponge (approved):

lib/Zonemaster/Backend/DB.pm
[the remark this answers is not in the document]
pnax: Thanks for the proof of concept. I think I'll go with your suggestion to "leav[e] this as a future exercise". I'll open an issue for that.

PR #1093, opened 2023-03-16
Remove dependency to String::ShellQuote

pnax:
## Purpose

Zonemaster::Backend depends on String::ShellQuote. However it seems we don't use [its functions](https://metacpan.org/pod/String::ShellQuote#DESCRIPTION) anywhere in the module. It should be safe to remove this dependency.

## Context

Remark on the `zonemaster-users` mailing list.

## Changes

* Remove dependency to String::ShellQuote in the code and installation instructions

## How to test this PR

Test should pass and installation should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,6 @@ addons:
       - libplack-perl
       - librole-tiny-perl
       - librouter-simple-perl
-      - libstring-shellquote-perl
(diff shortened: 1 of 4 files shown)


matsduf (approved):
Looks fine, but should preferably be tested by Travis before merging.

PR #1111, opened 2023-06-07
Reinstate old RPC API methods

pnax:
## Purpose

Remove deprecation warning and comments to all old RPC API methods and mark the new methods as experimental.
What were the deprecated methods (and old stable) are again the default stable API methods.

## Context

https://github.com/zonemaster/zonemaster/issues/1159

Relates to https://github.com/zonemaster/zonemaster-backend/pull/1054 and https://github.com/zonemaster/zonemaster-backend/pull/1083.

## Changes

Revert commits where old methods were deprecated and based upon new methods. Now the experimental methods are based on the stable one.

## How to test this PR

* Unit tests should pass
* Review the code and verify that the unit tests cover the stable API
* Check that the new experimental API is working (there is currently no unit test)
* Check that changes made in Backend for this release are still working (and have not been inadvertently reverted)

### DIFF
--- a/t/test01.t
+++ b/t/test01.t
@@ -175,7 +175,7 @@ subtest 'API calls' => sub {
 
     subtest 'get_data_from_parent_zone' => sub {
         my $res = $rpcapi->get_data_from_parent_zone( { domain => "fr" } );
-        #diag explain( $res );
+        note explain( $res );
         ok( defined( $res->{ns_list} ), 'Has a list of nameservers' );
         ok( defined( $res->{ds_list} ), 'Has a list of DS records' );
 
@@ -234,7 +234,7 @@ subtest 'get_test_history' => sub {
     };
 
     $test_history = $rpcapi->get_test_history( $method_params );
-    #diag explain( $test_history );
+    note explain( $test_history );
     is( scalar( @$test_history ), 2, 'Two tests created' );
 
     foreach my $res (@$test_history) {
@@ -255,7 +255,7 @@ subtest 'get_test_history' => sub {
         ( $hash_id ) = $rpcapi->{db}->get_test_request();
 
         $test_history = $rpcapi->get_test_history( $method_params );
-        #diag explain( $test_history );
+        note explain( $test_history );
         is( scalar( @$test_history ), 2, 'Only 2 tests should be retrieved' );
 
         # now run the test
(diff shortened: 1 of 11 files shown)


mattias-p (dismissed):
I haven't tested but this looks good to me.

mattias-p:
In the psgi file there's a check for `$ini->RPCAPI_enable_user_create or $ini->RPCAPI_enable_add_batch_user`. Since RPCAPI.enable_add_batch_user is truthy by default, the only way to disable the methods is to disable RPCAPI.enable_add_batch_user (and leave RPCAPI.enable_user_create disabled), right?

Also from my point of view it would be better to forbid specifying both RPCAPI.enable_add_batch_user and RPCAPI.enable_user_create in the same config.

pnax:
I've updated the PR to be aligned with the current documentation in https://github.com/zonemaster/zonemaster/pull/1162. And as suggested, an error is now raised if both stable and experimental parameters are defined in the configuration file.

t/test01.t
```diff
@@ -254,35 +251,32 @@ subtest 'domain_history' => sub {
...
-        $rpcapi->job_create( $params );
+        $rpcapi->start_domain_test( $params );
         ( $hash_id ) = $rpcapi->{db}->get_test_request();
 
-        my $res_domain_history;
-        $res_domain_history = $rpcapi->domain_history( $method_params );
-        $domain_history = $res_domain_history->{history};
-        is( scalar( @$domain_history ), 2, 'Only 2 tests should be retrieved' );
+        $test_history = $rpcapi->get_test_history( $method_params );
+        #diag explain( $test_history );
```
mattias-p, line 258: Just a thought: Would it make sense to call `note` here instead of having a commented out call to `diag`?

mattias-p (approved):
LGTM!

t/test01.t
pnax: Good suggestion. I've updated.

pnax:
To quote comment from the documentation PR (https://github.com/zonemaster/zonemaster/pull/1162#issuecomment-1587474136):

AFAICT this PR is done. I've gotten a few reviews, addressed all comments up to this point and everybody seems ok with the implementation.

> I'm leaving it open until the implementation is done too so they can be merged at the same time.

I think we reached that point.

mattias-p (approved):

matsduf:
I think this can be merged.

PR #1118, opened 2023-07-12
Remove multiple batch restriction

matsduf:
## Purpose

There is an artificial restriction on multiple concurrent batches with the same user, i.e. the second batch cannot be created until the first has completed. It does not give any protection on overload since a batch user can create batch with a million zones, but not two concurrent batches with 1000 each. There can be good reasons to create multiple batches if the different batches should use profile or different settings.

This PR removes the restriction. It also adjusts the unit test.

The documentation is updated in https://github.com/zonemaster/zonemaster/pull/1178

## How to test this PR

Before this PR try with the following two commands, and the second will be refused (assuming batch is accepted and `USER` is a batch user with API key `KEY`):

```
zmb add_batch_job --username USER --api-key KEY --domain iis.se --domain arpa --domain com --domain info --domain se
```
Then repeat after this PR has been applied.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -398,23 +398,6 @@ sub create_new_batch_job {
     my ( $self, $username ) = @_;
 
     my $dbh = $self->dbh;
-    my ( $batch_id, $created_at ) = $dbh->selectrow_array( "
-            SELECT
-                batch_id,
-                batch_jobs.created_at AS batch_created_at
-            FROM
-                test_results
-            JOIN batch_jobs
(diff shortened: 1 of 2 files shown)


pnax (approved):
LGTM to lift this limitation.

mattias-p (approved):
Awesome!

marc-vanderwal:
Release testing report – No issues found

Rocky Linux 9.3 + PostgreSQL 13.11

Running the same `zmb add_batch_job` command as in the testing procedure (modified to match the user and API key configured on my test VM) about 3 seconds apart causes two batches to be created without error.

PR #1094, opened 2023-03-30
Clickhouse: new database engine (experimental)

pnax:
## Purpose

[Clickhouse](https://clickhouse.com/) is a vertical database engine that appears to be quite efficient when it comes to run Zonemaster with millions of domains

## Context

* Clickhouse is currently used internally at Afnic to monthly store Zonemaster reslults on the .fr zone (4 millions domains) 

## Changes

* new Clickhouse database engine
* uses the MySQL DBI to connect to the Clickhouse server

### Setup a Clickhouse server

* [install Clickhouse](https://clickhouse.com/docs/en/install)
* Generate a [double SHA1 password](https://clickhouse.com/docs/en/operations/settings/settings-users#password_double_sha1_hex) for the database user
   ```
   PASSWORD=$(base64 < /dev/urandom | head -c8); echo "$PASSWORD"; echo -n "$PASSWORD" | sha1sum | tr -d '-' | xxd -r -p | sha1sum | tr -d '-'
   ```
* Setup a `zonemaster` database and a `zonemaster` user:
   ```
   $ clickhouse-client
   :) CREATE DATABASE zonemaster;
   :) CREATE USER zonemaster IDENTIFIED WITH double_sha1_hash BY '<DOUBLE_SHA1>';
   :) GRANT CREATE TABLE, DROP TABLE, SELECT, INSERT, ALTER UPDATE ON zonemaster.* TO zonemaster;
   ```

### Configure and use Zonemaster-Backend

Adapt the `share/backend_config.ini` file and use `zmb` or `zmtest`.
You could check any results by running some SQL commands with `clickhouse-client`.
```
$ clickhouse-client
:) USE zonemaster;
:) SELECT * FROM test_results;
```

### Run the unit tests

Unit test should work when using Clickhouse.
```
TARGET=Clickhouse prove -l t/
```

PR #1121, opened 2023-07-19
Synchronized claiming of jobs for processing

mattias-p:
## Purpose

Make it possible for multiple Test Agents to process the same queue without multiple workers running the same job.

In #1115 @matsduf said:
> When a large batch is to be tested, adding several Test Agents increases performance if the server has large capacity. Several servers could also be used. One limitation is that current code only allows one Test Agent per queue. All [jobs] in a batch always belong to the same queue.

## Meta

In this PR description I'm using the term "job" instead of the traditional term "test". Speaking of *the testing of jobs* is less confusing than speaking of *the testing of tests*.

However in the code changes I used the term "test" in anticipation of "the great renaming".

## Context

* Replaces #1115.

#### Reviewing
The easiest way to review this is probably to do it commit by commit.

#### New constants
* `$TEST_{WAITING,RUNNING,COMPLETED`~`,CRASHED,LAPSED`~`}` are introduced to represent the different states of a job.

#### New methods
* `test_state()` is introduced to query the state of a job.
* `claim_test()` is introduced to transition jobs from "waiting" to "running".

#### Updated methods
* `get_test_request()` synchronizes its claiming of jobs.
* when `get_test_request()` finds a job that it then fails to claim, it immediately tries to find and claim another one.
* `test_progress()` is no longer able to update the state of a job.
* `test_progress()` is no longer able to update the progress value of a job that is not "running".
* `test_progress()` is no longer able to decrease the progress value of a job.
* `test_progress(0)` is now interpreted as a request to set the progress value to zero, rather than a request to read the current progress value.
* `store_results()` is no longer able to update a job that is not "running".

#### New test files
* `t/queue.t` is introduced to test the claiming of waiting jobs. The `t/lifecycle.t` already has some code to exercise reuse of jobs which is quite related. However I wanted to make sure the test database was clean for the claiming tests and this is what I ended up with. Maybe `t/lifecycle.t` should be reserved for testing individual jobs and `t/queue.t` could test the enqueuing and dequeuing of jobs.

## How to test this PR

* Run a large batch with one Test Agent and nothing should be changed.
* Run a large batch with two Test Agents (the same queue) and there should be no cases when the same job is run by more than one worker.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -547,6 +547,9 @@ If $queue_label is defined it must be an integer.
 If defined, only tests in the associated queue are considered.
 Otherwise tests from all queues are considered.
 
+Returns the test id and the batch id of the claimed test.
+If there are no waiting tests to claim, C<undef> is returned for both ids.
+
 Only tests in the "waiting" state are considered.
 When a test is claimed it is removed from the queue and it transitions to the
 "running" state.
@@ -561,47 +564,52 @@ Dies when an error occurs in the database interface.
 sub get_test_request {
     my ( $self, $queue_label ) = @_;
 
-    # Identify a candidate for allocation ...
-    my ( $hash_id, $batch_id );
-    if ( defined $queue_label ) {
-        ( $hash_id, $batch_id ) = $self->dbh->selectrow_array(
-            q[
-                SELECT hash_id,
-                       batch_id
-                FROM test_results
-                WHERE progress = 0
-                  AND queue = ?
-                ORDER BY priority DESC,
-                         id ASC
-                LIMIT 1
-            ],
-            undef,
-            $queue_label,
-        );
-    }
-    else {
-        ( $hash_id, $batch_id ) = $self->dbh->selectrow_array(
-            q[
-                SELECT hash_id,
-                       batch_id
-                FROM test_results
-                WHERE progress = 0
-                ORDER BY priority DESC,
-                         id ASC
-                LIMIT 1
-            ],
-        );
-    }
+    while ( 1 ) {
 
-    if ( defined $hash_id ) {
+        # Identify a candidate for allocation ...
+        my ( $hash_id, $batch_id );
+        if ( defined $queue_label ) {
+            ( $hash_id, $batch_id ) = $self->dbh->selectrow_array(
+                q[
+                    SELECT hash_id,
+                           batch_id
+                    FROM test_results
+                    WHERE progress = 0
+                      AND queue = ?
+                    ORDER BY priority DESC,
+                             id ASC
+                    LIMIT 1
+                ],
+                undef,
+                $queue_label,
+            );
+        }
+        else {
+            ( $hash_id, $batch_id ) = $self->dbh->selectrow_array(
+                q[
+                    SELECT hash_id,
+                           batch_id
+                    FROM test_results
+                    WHERE progress = 0
+                    ORDER BY priority DESC,
+                             id ASC
+                    LIMIT 1
+                ],
+            );
+        }
 
-        # ... and race to be the first to claim it ...
-        if ( $self->claim_test( $hash_id ) ) {
-            return ( $hash_id, $batch_id );
+        if ( defined $hash_id ) {
+
+            # ... and race to be the first to claim it ...
+            if ( $self->claim_test( $hash_id ) ) {
+                return ( $hash_id, $batch_id );
+            }
+        }
+        else {
+            # ... or stop trying if there are no candidates.
+            return ( undef, undef );
         }
     }
-
-    return ( undef, undef );
 }
 
 =head2 claim_test( $test_id )
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -52,11 +52,48 @@ has 'dbhandle' => (
     required => 1,
 );
 
-Readonly our $TEST_WAITING   => 'waiting';
-Readonly our $TEST_RUNNING   => 'running';
+=head2 $TEST_WAITING
+
+The test is waiting to be processed.
+
+=cut
+
+Readonly our $TEST_WAITING => 'waiting';
+
+=head2 $TEST_RUNNING
+
+The test is currently being processed.
+
+=cut
+
+Readonly our $TEST_RUNNING => 'running';
+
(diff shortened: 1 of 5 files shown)


lib/Zonemaster/Backend/DB.pm
```diff
@@ -209,60 +226,123 @@ sub recent_test_hash_id {
...
-                $self->format_time( time() ),
-                $test_id,
-            );
+    if ( defined $progress ) {
+        if ( $progress < 0 || 100 < $progress ) {
+            die Zonemaster::Backend::Error::Internal->new( reason => "progress out of range" );
+        } elsif ( $progress < 1 ) {
+            $progress = 1;
+        } elsif ( 99 < $progress ) {
+            $progress = 99;
```
matsduf, line 271: Why not simplify to
```
if ( $progress < 1 || $progress > 99) {
            die Zonemaster::Backend::Error::Internal->new( reason => "progress can only be set to 1-99" );
 ```

File: lib/Zonemaster/Backend/DB.pm [hunk 2]
```diff
@@ -452,24 +583,110 @@ sub batch_exists_in_db {
...
 
-    my ( $hash_id, $batch_id );
-    if ( defined $queue_label ) {
-        ( $hash_id, $batch_id ) = $dbh->selectrow_array( qq[ SELECT hash_id, batch_id FROM test_results WHERE progress=0 AND queue=? ORDER BY priority DESC, id ASC LIMIT 1 ], undef, $queue_label );
-    }
-    else {
-        ( $hash_id, $batch_id ) = $dbh->selectrow_array( q[ SELECT hash_id, batch_id FROM test_results WHERE progress=0 ORDER BY priority DESC, id ASC LIMIT 1 ] );
-    }
+        # Identify a candidate for allocation ...
+        my ( $hash_id, $batch_id );
```
matsduf, line 614: In `claim_test()` and `test_state()` you call it $test_id, here $hash_id. I think the code would be easier to read if it was consistent.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -50,6 +52,71 @@ has 'dbhandle' => (
...
+
+Readonly our $TEST_CRASHED => 'crashed';
+
+=head2 $TEST_LAPSED
+
+The test was already processed.
+
+=cut
+
+Readonly our $TEST_LAPSED => 'lapsed';
```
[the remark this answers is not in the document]
mattias-p: I'm quite sure I did remove it. But I'll do it again. Good catch.

lib/Zonemaster/Backend/DB.pm
mattias-p: I considered renaming it but then I'd rather rename it to $job_id. So I opted to defer the renaming until we make the move to "job".

matsduf (commented):
I think this looks good. As far as I can see it fully replaces #1115. I hope that @matsstralbergiis will be able to run a test with many domain names and at least two Test Agents soon to see that there in fact are no collisions or something else that seems to break.

matsstralbergiis:
I have tonight run a batch with 11004 zones and used three testagents with this DB.pm and it seems to work.
When examining the log files from all three testagents I get the same number of tests finished.
`grep "Test completed" /var/log/zonemaster/zm-testagent*.log |wc
  11004   77028 1243441`

matsduf:
> I have tonight run a batch with 11004 zones and used three testagents with this DB.pm and it seems to work.

@matsstralbergiis, were there any test that was started by more than one Test Agent?

matsstralbergiis:
As the sum of tests completed in the log files matches the number I submitted to test I would say no.
When running the same zones without the updated DB.pm I get 20-30% more "tests completed" in the log files.

matsduf (approved):

pnax (approved):
Very nice!
A few nitpicking remarks:
* I left a comment on a documentation line that feels unclear to me
* I realize you don't use the new `test_state()` method, but it's nice to have it.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -452,24 +567,110 @@ sub batch_exists_in_db {
...
+
+If $queue_label is defined it must be an integer.
+If defined, only tests in the associated queue are considered.
+Otherwise tests from all queues are considered.
+
+Returns the test id and the batch id of the claimed test.
+If there are no waiting tests to claim, C<undef> is returned for both ids.
+
+Only tests in the "waiting" state are considered.
+When a test is claimed it is removed from the queue and it transitions to the
```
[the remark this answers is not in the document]
matsduf: Both "removed from the queue" and "unavailable in its queue" is not very clear. Every test belongs to a queue through the queue field in the test_results table. Queue 0 is the default queue, and the test will remain in its queue forever. Rather "The test is now unavailable for other Test Agent processes."

marc-vanderwal:
I tested the combination of this PR with the experimental Clickhouse support (see #1094) and it doesn’t work well (see [my comment](https://github.com/zonemaster/zonemaster-backend/pull/1094#issuecomment-1853695426) for additional details). It seems this is due to the lack of proper transaction isolation when Clickhouse performs ALTER TABLE … UPDATE queries.

So it’s important to point out that the idea in this PR only works if Zonemaster-Engine is configured to use a proper DBMS that does have some sort of atomic UPDATE. I’m going to try with PostgreSQL now…

marc-vanderwal:
**Release testing report – No technical issues, one documentation issue**

Rocky Linux 9.3 + PostgreSQL 13.11

The best way of running more than one instance of zm-testagent on the same machine is not documented yet. I opened an issue about it (see zonemaster/zonemaster#1234). I duplicated the systemd unit file [as described here](https://github.com/zonemaster/zonemaster/issues/1234#issuecomment-1853483080) with satisfactory results.

In the testing procedure, the meaning of “large batch” was left unspecified. I picked an arbitrary value of 1 000 because I felt it was a reasonably large batch and a reasonable compromise in terms of runtime; it runs in about 50 to 60 minutes on two test agents on my test VM. The batch consists of 1 000 random subdomains under `.fr` from an internal data source.

With one test agent, I witnessed nothing out of the extraordinary in terms of behavior. I waited for the completion of about 500 tests before assuming this change does not break anything.

With two test agents, one agent found, started and completed 493 tests and the second one did 507 tests. This correctly adds up to 1 000 and one can reasonably assume that the load was spread evenly.

Issue #1032, opened 2022-06-03
Creating test with empty domain gives confusing error message


matsduf:
An empty domain is not allowed and should not be allowed, but it would be good if the error message says that the domain name must not be empty, not giving an incorrect message that the characters are not supported.

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "params": {
    "domain": ""
  },
  "method": "start_domain_test"
}
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "data": [
      {
        "path": "/domain",
        "message": "The domain name character(s) are not supported"
      }
    ],
    "message": "Invalid method parameter(s).",
    "code": "-32602"
  }
}
```

Here is a correct example of a domain with a character not supported, and the error message is the same:
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "params": {
    "domain": "zonemaster:net"
  },
  "method": "start_domain_test"
}
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "data": [
      {
        "path": "/domain",
        "message": "The domain name character(s) are not supported"
      }
    ],
    "message": "Invalid method parameter(s).",
    "code": "-32602"
  }
}
```

PR #1132, opened 2023-10-24
use name normalization method

blacksponge:
## Purpose

Use new name normalization method instead of Basic00.

## Context

https://github.com/zonemaster/zonemaster-engine/pull/1040 introduce a new method to normalize domain names before testing. Let's use it.

It also fixes https://github.com/zonemaster/zonemaster-backend/issues/1032.

## Changes

* Normalize domain name before saving to database
* Use normalized domain for history search
* Normalize domain name before fetching data from parent (fixes issue that cause that method to not work for idna domains, see https://github.com/zonemaster/zonemaster-gui/issues/426)

**Note:** With this PR, domains are normalized before being saved to database, meaning that domains that contain ulabel will be converted to alabel before saving. That might break history for the affected domains. Should we provide a migration script to normalize all the domains in the database?

## How to test this PR

Test that the domain validation is working

```
% ./script/zmb --server http://127.0.0.1:5000 start_domain_test --domain '' | jq .error.data -c
[{"message":"Domain name is empty.","path":"/domain"}]

% ./script/zmb --server http://127.0.0.1:5000 start_domain_test --domain , | jq .error.data -c
[{"path":"/domain","message":"Domain name has an ASCII label (\",\") with a character not permitted."}]
```

Test that the validation is also working for nameserver

```
% ./script/zmb --server http://127.0.0.1:5000 start_domain_test --domain example.org --nameserver ':0.0.0.0' | jq .error.data -c
[{"path":"/nameservers/0/ns","message":"Domain name is empty."}]
```

Test that fetching data from parent is working for idna domains

```
% curl 'http://localhost:5000' --data '{"jsonrpc":"2.0","id":0,"method":"get_data_from_parent_zone","params":{"domain":"café.fr"}}'
{"jsonrpc":"2.0","result":{"ds_list":[],"ns_list":[{"ns":"ns1.parkingcrew.net","ip":"13.248.158.159"},{"ns":"ns2.parkingcrew.net","ip":"76.223.21.9"}]},"id":0}
```

* Try lauching a test using a GUI, check that it is working correctly.
* Then check history, check that is is working correctly.

### DIFF
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -875,7 +875,7 @@ sub process_dead_test {
 sub _normalize_domain {
     my ( $domain ) = @_;
 
-    # Discard errors, at this point errors have already checked
+    # Discard errors, at this point errors have already been checked
     my ( $_errors, $normalized_domain ) = normalize_name( $domain );
 
     return $normalized_domain;
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -16,6 +16,7 @@ use POSIX qw( strftime );
 use Readonly;
 use Try::Tiny;
 
+use Zonemaster::Engine::Normalization;
 use Zonemaster::Backend::Errors;
 use Zonemaster::Engine::Logger::Entry;
 
@@ -874,10 +875,10 @@ sub process_dead_test {
 sub _normalize_domain {
     my ( $domain ) = @_;
 
-    $domain = lc( $domain );
-    $domain =~ s/\.$// unless $domain eq '.';
+    # Discard errors, at this point errors have already checked
+    my ( $_errors, $normalized_domain ) = normalize_name( $domain );
 
-    return $domain;
+    return $normalized_domain;
 }
 
 sub _project_params {
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -875,8 +875,11 @@ sub process_dead_test {
 sub _normalize_domain {
     my ( $domain ) = @_;
 
-    # Discard errors, at this point errors have already been checked
-    my ( $_errors, $normalized_domain ) = normalize_name( $domain );
+    my ( $errors, $normalized_domain ) = normalize_name( $domain );
+
+    if ( scalar( @{$errors} ) ) {
+        die Zonemaster::Backend::Error::Internal->new( reason => "Normalizing domain returned errors.", data => $errors );
+    }
 
     return $normalized_domain;
 }
--- a/lib/Zonemaster/Backend/DB.pm
+++ b/lib/Zonemaster/Backend/DB.pm
@@ -878,7 +878,7 @@ sub _normalize_domain {
     my ( $errors, $normalized_domain ) = normalize_name( $domain );
 
     if ( scalar( @{$errors} ) ) {
-        die Zonemaster::Backend::Error::Internal->new( reason => "Normalizing domain returned errors.", data => $errors );
+        die Zonemaster::Backend::Error::Internal->new( reason => "Normalizing domain returned errors.", data => [ map { $_->string } @{$errors} ] );
     }
 
     return $normalized_domain;
--- a/lib/Zonemaster/Backend/Validator.pm
+++ b/lib/Zonemaster/Backend/Validator.pm
@@ -284,8 +284,8 @@ sub check_domain {
 
     my ( $errors, $_domain ) = normalize_name( $domain );
 
-    if ( @{$errors} != 0 ) {
-        return @{$errors}[0]->message;
+    if ( @{$errors} ) {
+        return $errors->[0]->message;
     }
 
     return undef
--- a/lib/Zonemaster/Backend/Validator.pm
+++ b/lib/Zonemaster/Backend/Validator.pm
@@ -13,6 +13,7 @@ use Readonly;
 use Locale::TextDomain qw[Zonemaster-Backend];
 use Net::IP::XS;
 use Zonemaster::Engine::Logger::Entry;
+use Zonemaster::Engine::Normalization;
 use Zonemaster::LDNS;
 
 our @EXPORT_OK = qw(
@@ -281,35 +282,13 @@ sub check_domain {
         return N__ 'Domain name required';
     }
 
-    if ( $domain =~ m/[^[:ascii:]]+/ ) {
-        if ( Zonemaster::LDNS::has_idn() ) {
-            eval { $domain = Zonemaster::LDNS::to_idn( $domain ); };
-            if ( $@ ) {
-                return N__ 'The domain name is IDNA invalid';
-            }
-        }
-        else {
-            return N__ 'The domain name contains non-ascii characters and IDNA support is not installed';
-        }
+    my ( $errors, $_domain ) = normalize_name( $domain );
+
+    if ( @{$errors} != 0 ) {
+        return @{$errors}[0]->message;
     }
 
-    if ( $domain !~ m{^[a-z0-9_./-]+$}i ) {
-        return  N__ 'The domain name character(s) are not supported';
-    }
-
-    if ( $domain =~ m/\.\./i ) {
-        return  N__ 'The domain name contains consecutive dots';
-    }
-
-    my %levels = Zonemaster::Engine::Logger::Entry::levels();
-    my @res;
-    @res = Zonemaster::Engine::Test::Basic->basic00( $domain );
-    @res = grep { $_->numeric_level >= $levels{ERROR} } @res;
-    if ( @res != 0 ) {
-        return N__ 'The domain name or label is too long';
-    }
-
-    return undef;
+    return undef
 }
 
 =head2 check_language_tag($value, %locales)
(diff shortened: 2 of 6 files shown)


lib/Zonemaster/Backend/DB.pm
```diff
@@ -619,10 +620,10 @@ sub process_dead_test {
 sub _normalize_domain {
     my ( $domain ) = @_;
 
-    $domain = lc( $domain );
-    $domain =~ s/\.$// unless $domain eq '.';
+    # Discard errors, at this point errors have already checked
```
marc-vanderwal, line 623:
```suggestion
    # Discard errors, at this point errors have already been checked
```

lib/Zonemaster/Backend/Validator.pm
```diff
@@ -281,35 +282,13 @@ sub check_domain {
...
+    my ( $errors, $_domain ) = normalize_name( $domain );
 
-    my %levels = Zonemaster::Engine::Logger::Entry::levels();
-    my @res;
-    @res = Zonemaster::Engine::Test::Basic->basic00( $domain );
-    @res = grep { $_->numeric_level >= $levels{ERROR} } @res;
-    if ( @res != 0 ) {
-        return N__ 'The domain name or label is too long';
+    if ( @{$errors} != 0 ) {
+        return @{$errors}[0]->message;
```
marc-vanderwal, line 288: I think this won’t work as expected. Because `$errors` is an arrayref, this should:
```suggestion
    if ( @{$errors} ) {
        return $errors->[0]->message;
```

lib/Zonemaster/Backend/DB.pm
blacksponge: done.

lib/Zonemaster/Backend/Validator.pm
blacksponge: done.

blacksponge:
Waiting for #1092 to update the DB migration script.

lib/Zonemaster/Backend/DB.pm
```diff
@@ -619,10 +620,10 @@ sub process_dead_test {
 sub _normalize_domain {
     my ( $domain ) = @_;
 
-    $domain = lc( $domain );
-    $domain =~ s/\.$// unless $domain eq '.';
+    # Discard errors, at this point errors have already been checked
```
mattias-p, line 623: Maybe we should be so bold as to `die` if we get normalization errors at this point?

mattias-p (dismissed):
LGTM

lib/Zonemaster/Backend/DB.pm
marc-vanderwal: Or maybe introduce a function in `Zonemaster::Engine::Normalization` that attempts normalization and just returns the normalized name on success instead of a list.

matsduf (dismissed):
```
Press return to run 'make distcheck'...
Not in MANIFEST: t/db_ddl.t
Not in MANIFEST: t/queue.t
```

The error comes from older PRs, not this one. Fixed in #1136.

matsduf:
`Rå$ttgift.se` is captured as having errors ("Domain name has a non-ASCII label (\"Rå$ttgift\") which is not a valid U-label.") but `Rå_ttgift.se` passes without error. I assume this is an error in Engine.

matsduf:
It does not work to use `zmb` for IDN names with U-label, see #1138.

matsduf (approved):

blacksponge:
I have updated the migration script. I am currently running it against a snapshot of zonemaster.net to see how long it takes.

lib/Zonemaster/Backend/DB.pm
blacksponge: I chose to die if there is an error.

mattias-p (approved):
LGTM. I have one question though.

matsduf:
@blacksponge

* What happens if there is a domain name that cannot be normalized, e.g. containing character not allowed or being a U-label that cannot be converted to A-label? I think I would suggest keeping the domain name as-is in the database.

* Do you know why Travis breaks?

marc-vanderwal (approved):

blacksponge:
> @blacksponge, What happens if there is a domain name that cannot be normalized, e.g. containing character not allowed or being a U-label that cannot be converted to A-label? I think I would suggest keeping the domain name as-is in the database.

Good suggestion, I'll do that.

blacksponge:
> > @blacksponge, What happens if there is a domain name that cannot be normalized, e.g. containing character not allowed or being a U-label that cannot be converted to A-label? I think I would suggest keeping the domain name as-is in the database.
> 
> Good suggestion, I'll do that.

I implemented that.

```sh
% perl share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
Configured database engine: SQLite
Starting database migration

-> (1/2) Populating new result_entries table
Will update 0 rows
Progress update: 0 / 0

-> (2/2) Normalizing domain names
Will update 5 rows
Caught error while updating record, ignoring: Caught Zonemaster::Backend::Error::Internal in the `Zonemaster::Backend::DB::_normalize_domain` method: Normalizing domain returned errors. Context: ['Domain name has repeated dots.']
20%
40%
60%
80%
100%

Migration done
```

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
```diff
@@ -96,19 +96,24 @@ sub _update_data_nomalize_domains {
...
+            my $fingerprint = $db->generate_fingerprint( $raw_params );
 
-        $domain = Zonemaster::Backend::DB::_normalize_domain( $domain );
+            $domain = Zonemaster::Backend::DB::_normalize_domain( $domain );
 
-        $db->dbh->do('UPDATE test_results SET domain = ?, params = ?, fingerprint = ? where hash_id = ?', undef, $domain, $params, $fingerprint, $hash_id);
+            $db->dbh->do('UPDATE test_results SET domain = ?, params = ?, fingerprint = ? where hash_id = ?', undef, $domain, $params, $fingerprint, $hash_id);
+        };
+        if ($@) {
+            warn "Caught error while updating record, ignoring: $@\n";
```
mattias-p, line 115: Would it be helpful to explicitly include the hash id in this message?

mattias-p (approved):

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
blacksponge: I updated it to include the hash id,
```
Configured database engine: SQLite
Starting database migration

-> (1/2) Populating new result_entries table
Will update 0 rows
Progress update: 0 / 0

-> (2/2) Normalizing domain names
Will update 5 rows
Caught error while updating record with hash id b212fc97d29d51f3, ignoring: Caught Zonemaster::Backend::Error::Internal in the `Zonemaster::Backend::DB::_normalize_domain` method: Normalizing domain returned errors. Context: ['Le nom de domaine contient plusieurs points successifs.']
20%
40%
60%
80%
100%

Migration done
```

marc-vanderwal (approved):

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
```diff
@@ -77,7 +79,48 @@ sub _update_data {
...
+
+            # This has never been cleaned
+            delete $raw_params->{user_ip};
+
+            my $params = $db->encode_params( $raw_params );
+            my $fingerprint = $db->generate_fingerprint( $raw_params );
+
+            $domain = Zonemaster::Backend::DB::_normalize_domain( $domain );
+
+            $db->dbh->do('UPDATE test_results SET domain = ?, params = ?, fingerprint = ? where hash_id = ?', undef, $domain, $params, $fingerprint, $hash_id);
```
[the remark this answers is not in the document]
blacksponge: No, it stop the execution and exit the eval block as there is a die in the _normalize_domain method.

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
```diff
@@ -77,7 +79,48 @@ sub _update_data {
...
+        };
+        if ($@) {
+            warn "Caught error while updating record with hash id $hash_id, ignoring: $@\n";
+        }
```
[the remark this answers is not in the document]
blacksponge: I was thinking of decoding related errors.

pnax (approved):
lgtm

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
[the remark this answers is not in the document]
matsduf: So if there is one single domain name that cannot be normalized then no conversion can be done. I do not think that is reasonable.

matsduf (changes requested):
It must be possible to ignore a domain name that cannot be normalized. I think the script should ignore such a domain name by default, and just keep that domain name unnormalized, and proceed with the conversion.

share/patch/patch_db_zonemaster_backend_ver_11.0.3.pl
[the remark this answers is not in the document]
blacksponge: Well not at all, I am updating the domains one by one, so if one domain fail it does not affect the others.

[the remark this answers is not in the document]
blacksponge: The eval block just handle the migration of one single domain.

blacksponge:
> It must be possible to ignore a domain name that cannot be normalized. I think the script should ignore such a domain name by default, and just keep that domain name unnormalized, and proceed with the conversion.

This is exactly what it does.

matsduf (approved):

PR #1137, opened 2023-11-24
Adds unit test for MANIFEST

matsduf:
## Purpose

Adds unit test that will capture missing updates for MANIFEST. Copied from Zonemaster-Engine.

## How to test this PR

Verify that is reports when MANIFEST is not updated and not when it is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,24 +8,24 @@ stages:
   - test
 
 script:
-    - perl Makefile.PL && make test TEST_VERBOSE=1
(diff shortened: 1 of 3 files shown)


mattias-p (commented):
The changes in .travis.yml doesn't look right, but otherwise this looks good to me.

matsduf:
> The changes in .travis.yml doesn't look right, but otherwise this looks good to me.

@mattias-p: No, the changes are to try to find out why the MO files are not created. I cannot see them, and the `t/manifest.t` cannot see them. Do you know how to do it with Travis? It works in Zonemaster-Engine, but there we run Github actions.

An alternative is `make manifest`, see https://github.com/zonemaster/zonemaster-backend/pull/1136#issuecomment-1825744972

PR #1139, opened 2023-12-04
Make room for other database engines

pnax:
## Purpose

This PR removes the need to have a primary key defined within the database. For instance Clickhouse does not support it. So some specific code is needed. With this PR, such specific code could be removed and the same SQL queries could be used.

Also it appears that when running batches, the indexes are dropped and create again. But one index was not created as intended and included the `created_at` field. However with this PR, an index using the `created_at` field is necessary to keep good performance. So the initial index is updated (instead of updating the index in the batch method).

## Context

This was initially part of #1094, and moved to another PR to avoid changing internal while integrating an experimental feature.

## Changes

* compute the user existence by count (and not by returning its id)
* order the entries using the `created_at` field and not the `id` field
* use an explicit JOIN clause 

## How to test this PR

* Unit tests should pass

### DIFF
--- a/lib/Zonemaster/Backend/DB/PostgreSQL.pm
+++ b/lib/Zonemaster/Backend/DB/PostgreSQL.pm
@@ -222,7 +222,7 @@ sub add_batch_job {
 
     my $dbh = $self->dbh;
 
-    if ( $self->user_authorized( $params->{username}, $params->{api_key} ) ) {
+    if ( 1 == $self->user_authorized( $params->{username}, $params->{api_key} ) ) {
         $batch_id = $self->create_new_batch_job( $params->{username} );
 
         my $test_params = $params->{test_params};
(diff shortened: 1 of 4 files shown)


matsduf:
@pnax, this is for v2024.1, isn't it?

pnax:
> this is for v2024.1, isn't it?

lib/Zonemaster/Backend/DB/PostgreSQL.pm
```diff
@@ -81,7 +81,7 @@ sub create_schema {
     ) or die Zonemaster::Backend::Error::Internal->new( reason => "PostgreSQL error, could not create 'test_results' table", data => $dbh->errstr() );
 
     $dbh->do(
-        'CREATE INDEX IF NOT EXISTS test_results__hash_id ON test_results (hash_id)'
```
mattias-p, line 84: Don't we still need an index for just hash_id? E.g. for the `test_progress` RPCAPI method.

PR #1140, opened 2023-12-04
CI: update dependency and Perl versions

pnax:
## Purpose

This PR adds a missing dependency and updates the Perl versions for Travis CI.

## Context

https://github.com/zonemaster/zonemaster-engine/pull/1305
https://github.com/zonemaster/zonemaster-engine/pull/1287

## Changes

* Test against Perl 5.36, 5.30 and 5.26 instead of 5.30, 5.26 and 5.16;

## How to test this PR

Travis should pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ jobs:
     - env: TARGET=PostgreSQL ZONEMASTER_RECORD=0 ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_postgresql_backend_config.ini
       services: postgresql
     # Cover supported Perl versions with SQLite
-    - perl: "5.36"
-      env: TARGET=SQLite ZONEMASTER_RECORD=0 ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_sqlite_backend_config.ini
     - perl: "5.30"
       env: TARGET=SQLite ZONEMASTER_RECORD=0 ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_sqlite_backend_config.ini
     - perl: "5.26"
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ jobs:
     - perl: "5.30"
       env: TARGET=SQLite ZONEMASTER_RECORD=0 ZONEMASTER_BACKEND_CONFIG_FILE=./share/travis_sqlite_backend_config.ini
     - perl: "5.26"
-      dist: focal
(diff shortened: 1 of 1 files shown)


marc-vanderwal:
Better. But I wonder why Travis fails and why we can’t have the logs.

marc-vanderwal:
> > Better. But I wonder why Travis fails and why we can’t have the logs.
> 
> Unless I misunderstood you, you can see the logs in the usual "Checks" tab (https://github.com/zonemaster/zonemaster-backend/pull/1140/checks) and then clicking on any job number in the "Jobs and stages" section.
> 
> Right now this check is failing because Basic00 has just been removed in the develop branch of Engine and #1132 (which removes its usage in Backend) is not yet merged.

There might be an access issue, because all I see when I follow your steps (which is, incidentally, exactly what I tried) is this:
We’re sorry, but this data is not available. Please check the repository settings in Travis CI.

mattias-p:
I'm not currently logged into Travis (AFAICT) and I can see the Job log. Maybe they were having problem that's been fixed?
Anyway, here are the last few lines.
```
$ if [[ "$TARGET" == "PostgreSQL" ]]; then sudo sed -i 's/port = 5433/port = 5432/' /etc/postgresql/12/main/postgresql.conf; fi

sed: can't read /etc/postgresql/12/main/postgresql.conf: No such file or directory

The command "if [[ "$TARGET" == "PostgreSQL" ]]; then sudo sed -i 's/port = 5433/port = 5432/' /etc/postgresql/12/main/postgresql.conf; fi" failed and exited with 2 during .

Your build has been stopped.
```

marc-vanderwal:
> I'm not currently logged into Travis (AFAICT) and I can see the Job log. Maybe they were having problem that's been fixed? 

It seems unlikely, because I still can’t see those logs.

mattias-p:
Weird. I'm able to read it in a private browser window, so it seems unlikely to me that it's access related. Could it be some browser extension interfering?

marc-vanderwal:
> Weird. I'm able to read it in a private browser window, so it seems unlikely to me that it's access related. Could it be some browser extension interfering?

Oh, excellent suggestion! Using a private browser window works for me as well.

In a non-private browser window, I then deleted all local storage in `app.travis-ci.com`. I signed back in and all is well again.

matsduf:
> In a non-private browser window, I then deleted all local storage in `app.travis-ci.com`. I signed back in and all is well again.

After clearing the cookies for `travis-ci.com` I could see why Travis fails.

PR #1144, opened 2024-01-15
Avoid double UTF-8 encoding in zmb

### LINKED ISSUE #1138: zmb does not handle IDN names correctly
Zonemaster updated with https://github.com/zonemaster/zonemaster-backend/pull/1132 support IDN names to the RPC API using curl, e.g. with the following command,
```
curl 'http://localhost:5000' --data '{"jsonrpc":"2.0","id":0,"method":"get_data_from_parent_zone","params":{"domain":"Råttgift.se"}}'
```

But that does not work if `zmb` is used:
```
zmb start_domain_test --domain råttgift | jq
{
  "id": 1,
  "error": {
    "data": [
      {
        "path": "/domain",
        "message": "Domain name has a non-ASCII label (\"rÃ¥ttgift\") which is not a valid U-label."
      }
    ],
    "code": "-32602",
  },
  "jsonrpc": "2.0"
}
```

mattias-p:
## Purpose

This PR fixes a problem in how zmb handles UTF-8 data in its arguments.

## Context

Fixes #1138.

## Changes

Arguments are received as octet string. Before this change each argument octet would be interpreted as a native character (e.g. per ISO 8859-1) and encoded into UTF-8 before being injected into the JSON-RPC request. If the arguments were already UTF-8 encoded this would result in double UTF-8 encoding. After this change the argument octets are injected in verbatim into the JSON-RPC request.

## How to test this PR

You should be able to specify domain names with U-label to zmb like this:

```sh
$ zmb start_domain_test --domain råttgift
```

### DIFF
--- a/script/zmb
+++ b/script/zmb
@@ -3,12 +3,11 @@ use strict;
 use warnings;
 use feature 'say';
 
+use Encode qw( decode_utf8 FB_CROAK );
 use Getopt::Long qw( GetOptionsFromArray :config require_order );
-
-use Pod::Usage;
-
-use JSON::PP;
+use JSON::PP qw( encode_json );
 use LWP::UserAgent;
+use Pod::Usage;
 
 =head1 NAME
 
@@ -32,6 +31,8 @@ This interface is unstable and will change in a future release.
 
 sub main {
     my @argv = @_;
+
+    @argv = map { decode_utf8( $_, FB_CROAK ) } @argv;
 
     my $opt_help;
     my $opt_verbose;


marc-vanderwal (commented):
I have mixed feelings for this patch. It works (assuming your terminal uses the UTF-8 character encoding, which most do), but for the wrong reason. But a truly correct solution is much harder to implement.

The return value of `to_jsonrpc()` is a string that gets immediately fed into `to_request()`, which takes said string and plugs it into the content of an `HTTP::Request` object with the `content()` method.

The documentation for [content($bytes) in HTTP::Request](https://metacpan.org/pod/HTTP::Request#$r-%3Econtent(-$bytes-)) states that “the content should be a string of bytes”. So far, it *was* fed a string of bytes, but with this change, it is called with a Unicode string.

The real problem is that when `cmd_start_domain_test()` is called in the script, Getopt::Long doesn’t decode the bytes making up the `domain` command-line argument. That string should be interpreted according to the current terminal’s character encoding, but how to do that remains an open question…

mattias-p:
I agree that I'm assuming that the terminal is using UTF-8, or rather that the user by whatever means manages to place UTF-8 encoded bytes into `@ARGV`. An improvement could be to validate that the bytes actually conform to UTF-8.

I also agree that its difficult to know what encoding is actually used for the bytes in `@ARGV`, and if we knew that it would be nice to perform the proper conversion.

Additionally I agree that it works for the wrong reasons. I'm tricking `JSON::PP::encode` into believing that it's dealing with Unicode strings whose encoding it shouldn't touch, when in fact it is dealing with UTF-8 encoded strings whose encoding it shouldn't touch. Luckily all the JSON syntax it generates happens to be in ASCII which has identical encoding in UTF-8.

I guess the law abiding way would be to decode the UTF-8 byte strings taken from `@ARGV` before and let `JSON::PP` perform the UTF-8 encoding. Normally the law protects you from harm. But in the odd case when there's no harm in breaking the law, and when following the law entails having more code to write and maintain, more opportunity for bugs and less efficiency, I'm not sure I care so much for following the law.

Maybe I should just add some code comments to document the hack? And I just realized that I haven't thought about the UTF8 flag at all. If that gets set differently depending on the `utf8()` setting, that needs to be tweaked as well.

mattias-p:
I've updated this PR in response to @marc-vanderwal's comment as described in my last comment. Please review again.

marc-vanderwal (approved):
That should be fine. The worst case scenario is that IDN domain names can no longer be input on non-UTF-8 terminals. But this will only affect legacy systems and domains can still be input using their A-labels.

MichaelTimbert:
This PR works as expected on Ubuntu 20.04.6 and 22.04.4.
Work fine with U-labels & A-Labels.

PR #1159, opened 2024-03-19
Fix v2023.2 database migration script

tgreenx:
## Purpose

This PR proposes a fix to the database migration script of release v2023.2.

## Changes

- `List::Util::zip()` was only added in Perl v5.36. This makes use of `List::MoreUtils::zip_unflatten()` instead.

- The script filename is also updated to the proper Zonemaster-Backend version.

- The migration process for PostgreSQL has been heavily optimized in order to run successfully on production environments with a large `test_results` table.

## How to test this PR

- Run the script on (a copy of) Zonemaster’s database. The migration should yield exactly the same data (although the order of the rows in the new tables can’t be guaranteed and isn’t significant).

### DIFF
--- a/share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
+++ b/share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
@@ -191,7 +191,7 @@ sub _patch_db_postgresql_step1 {
     # definitely take out the machine it is running on!
     print("Starting up\n");
     $dbh->do(q[
-        DECLARE curs NO SCROLL CURSOR WITH HOLD FOR
+        DECLARE curs NO SCROLL CURSOR FOR
         SELECT
           test_results.hash_id,
           log_level.value AS level,
@@ -267,6 +267,7 @@ sub _patch_db_postgresql_step1 {
             die $_;
         }
     }
+    $dbh->do(q[CLOSE curs]);
     print("Done inserting ${row_inserted} rows\n");
 }
 


marc-vanderwal (approved):

matsduf (dismissed):
I have not tested, but the update looks fine to me. I assume it has been tested.

marc-vanderwal (commented):
There is more to be done, alas. The PostgreSQL code causes PostgreSQL to fill both disk and RAM on a production environment. There are more updates to come.

marc-vanderwal:
After inspecting a few results of both old and recent test cases in the database I’ve used for testing this new migration script, it seems that the migration went well. The Delegation01 message argument migration also went through successfully. Now is a good time to review this script. :)

share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
```diff
@@ -157,16 +159,118 @@ sub patch_db_mysql {
...
     } catch {
-        print( "Could not upgrade database:  " . $_ );
+        print( "\nCould not upgrade database:  " . $_ );
 
         $dbh->rollback();
     };
 }
 
+sub _patch_db_postgresql_step1 {
+    my ($dbh, $chunk_size) = @_;
```
matsduf, line 173: When called, only one parameter is included in the call. What is the purpose of the second parameter?

share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
```diff
@@ -178,61 +282,23 @@ sub patch_db_postgresql {
...
-                     json_array_elements(results) as res
-                     LEFT JOIN log_level ON (res->>'level' = log_level.level)
-            )
-        ]);
+        # Make sure the planner knows that log_level is a small table
+        # so it can optimize step 1 appropriately
+        $dbh->do(q[ANALYZE log_level]);
+
+        print( "\n-> (1/2) Populating new result_entries table\n" );
+        _patch_db_postgresql_step1( $dbh );
```
matsduf, line 290: Only one parameter in the call here. Shouldn't the default value of the second parameter be set here?

matsduf (commented):
After all testing the script will probably work for migration of Postgresql databases, but not much is common with Mysql. Are we confident that the script works for Mysql?

marc-vanderwal:
> After all testing the script will probably work for migration of Postgresql databases, but not much is common with Mysql. Are we confident that the script works for Mysql?

I’ve only conducted basic testing on MySQL (and a tiny dataset) when the first version of the script was tested for the previous release. I don’t have a large MySQL database on hand for testing, unfortunately.

share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
marc-vanderwal: The script alternates between fetching rows from the SELECT query and inserting them into the table by means of COPY, thus splitting one huge batch in many smaller batches. The `$chunk_size` parameter, by default set to 100 000, sets the number of rows fetched and inserted in each of those smaller batches.

matsduf:
I think it is fine to go now.

share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
matsduf: That is probably fine, but I think it is funny to have an input parameter to the `_patch_db_postgresql_step1()` that is never used in the call. It is a detail.

matsduf (approved):

PR #1174, opened 2024-06-25
Bump DBD::SQLite to 1.66

### LINKED ISSUE #1172: Unit test fails on Rocky 8
I ran the unit tests on Rocky 8 and test01.t blew up when I tried to install the dist tarball using sudo and cpanm.

```
make[1]: Leaving directory '/root/.cpanm/work/1719318302.1693/Zonemaster-Backend-11.1.1/share'
PERL_DL_NONLAZY=1 "/usr/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'inc', 'blib/lib', 'blib/arch')" t/*.t
t/00-load.t ................ ok
t/batches.t ................ ok
t/config.t ................. ok
t/db.t ..................... ok
t/db_ddl.t ................. ok
# not recording
t/idn.t .................... ok
t/lifecycle.t .............. ok
t/parameters_validation.t .. ok
t/queue.t .................. ok
t/rpc_validation.t ......... ok
# not recording
    # running the agent on test 6d02b1c7e0415d2e
# running the agent on test 1e74e2f11a5cf02e
DBD::SQLite::db prepare failed: too many SQL variables at /root/.cpanm/work/1719318302.1693/Zonemaster-Backend-11.1.1/blib/lib/Zonemaster/Backend/DB.pm line 1029.
DBD::SQLite::db prepare failed: too many SQL variables at /root/.cpanm/work/1719318302.1693/Zonemaster-Backend-11.1.1/blib/lib/Zonemaster/Backend/DB.pm line 1029.
# Looks like your test exited with 2 just after 3.
t/test01.t ................. 
Dubious, test returned 2 (wstat 512, 0x200)
All 3 subtests passed 
t/test_validate_syntax.t ... ok
t/translator.t ............. ok
t/validator.t .............. ok
```

mattias-p:
## Purpose

This PR ensures that zm-testagent doesn't fail at runtime trying to store larger test results in the database, and that the unit tests don't fail, because of a limitation in a dependency.

## Context

Fixes #1172.

On Rocky 8 and Ubuntu 20.04 the OS-provided DBD::SQLite is older than 1.66. On such systems cpanm will reject the OS package and instead pull the latest version from CPAN. The current installation instructions for Rocky and Ubuntu prescribes installing OS-packaged versions of DBD::SQLite.

We should consider what to do with the Rocky and Ubuntu installation instructions for Rocky. We have a few options, each with its own drawbacks.
1. Leave the installation instruction untouched. Users on Rocky 8 and Ubuntu 20.04 will get two versions of DBD::SQLite but only the newer one will be used.
2. Always install DBD::SQLite from CPAN. Quality control will be lower than needed for users on Ubuntu 22.04, 24.04 and Rocky 9.
3. Install DBD::SQLite from CPAN on Rocky 8 and Ubuntu 20.04 and from the OS on the others. This will raise the complexity of the installation instruction yet another step.

## Changes

Bumps DBD::SQLite to 1.66 in Makefile.PL.

## How to test this PR

1. Create a dist tarball including this change.
2. Install Zonemaster Engine.
3. Install dependencies for Zonemaster Backend.
4. Run `cpanm --test-only $DIST_TARBALL`.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -39,7 +39,7 @@ requires
   'Zonemaster::Engine'              => 5.000000,  # v5.0.0
   ;
 
-test_requires 'DBD::SQLite' => 1.4702;
+test_requires 'DBD::SQLite' => 1.66;
 test_requires 'Test::Differences';
 test_requires 'Test::Exception';
 test_requires 'Time::Local' => 1.26;
@@ -47,7 +47,7 @@ test_requires 'Test::NoWarnings';
 
 recommends 'DBD::mysql';
 recommends 'DBD::Pg';
-recommends 'DBD::SQLite';
+recommends 'DBD::SQLite' => 1.66;
 
 install_share;
 


mattias-p:
Regarding the installation instruction I have the least aversion against the drawbacks of option 1.

matsduf:
If two versions are installed then I guess it will depend on the path which one will be used. Do you see that there is a risk that the order is changed in the path leading to wrong version being used?

All options have drawbacks, and I do not think we should use option 2. I think I prefer option 3.

mattias-p:
> If two versions are installed then I guess it will depend on the path which one will be used.

Yeah, the OS package manager installs packages in one place and cpanm in another.

> Do you see that there is a risk that the order is changed in the path leading to wrong version being used?

The one installed by cpanm takes precedence so there should be no problem.

matsduf:
> > Do you see that there is a risk that the order is changed in the path leading to wrong version being used?
> 
> The one installed by cpanm takes precedence so there should be no problem.

The Perl path can be changed by the used, can't it? Maybe it is unlikely.

matsduf (dismissed):

matsduf (approved):

Issue #1191, opened 2024-11-29
Unit test fails in develop branch


matsduf:
I have installed develop branch of Zonemaster-LDNS (https://github.com/zonemaster/zonemaster-ldns/commit/f0b7b0a6661110d25e84b9390723404131ae128f) and Zonemaster-Engine (https://github.com/zonemaster/zonemaster-engine/commit/760f9ea9dbd5e0dbcc7ba13a48065f2646631016) and try to install a Zonemaster-Backend package from develop branch (https://github.com/zonemaster/zonemaster-backend/commit/dfb866509117055287a7e0da07caa99d0960d3db). When I run the unit tests I get the following error:
```
t/test01.t ................. # not recording
t/test01.t ................. 1/?     # running the agent on test 1ea22028f16ca7f5
t/test01.t ................. 2/? 
        #   Failed test 'Has correct DS values'
        #   at t/test01.t line 205.
        #     Structures begin differing at:
        #          $got->{keytag} = '29133'
        #     $expected->{keytag} = '51508'
        # Looks like you failed 1 test of 16.

    #   Failed test 'get_data_from_parent_zone'
    #   at t/test01.t line 206.
    # Looks like you failed 1 test of 6.

#   Failed test 'API calls'
#   at t/test01.t line 208.
# running the agent on test 4b926581a879494c
t/test01.t ................. 4/?         # running the agent on test 10222cc72c951c3f
t/test01.t ................. 5/?     # running the agent on test 83eb3b329bdc99c2
    # running the agent on test 65d77bef465d6438
    # running the agent on test 84e18259dfbabcf0
t/test01.t ................. 7/? # Looks like you failed 1 test of 8.
t/test01.t ................. Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/8 subtests 
(...)
Test Summary Report
-------------------
t/test01.t               (Wstat: 256 (exited 1) Tests: 8 Failed: 1)
  Failed test:  3
  Non-zero exit status: 1
Files=14, Tests=75, 47 wallclock secs ( 0.12 usr  0.02 sys + 12.52 cusr  1.23 csys = 13.89 CPU)
Result: FAIL
Failed 1/14 test programs. 1/75 subtests failed.
*** Error code 255
```

matsduf:
Will be fixed by #1192.

mattias-p:
I noticed that the develop branch passes CI. Is this still a problem?

matsduf:
[...]

PR #1192, opened 2024-11-30
Fixes broken unit tests

matsduf:
## Purpose

Unit test file `t/test01.t` leaks queries when network is set to off. A change in .fr NS and DS triggered the tests to break. This PR:

* Updates `t/test01.t` with correct expectations on .fr.
* Updates `t/test01.t` so that it does not leak queries.
* Updates `t/test01.data` to match updates on .fr and updated `t` file.
* Updates `t/idn.t` so that it does not leak queries.

## Context

Issue #1191

## How to test this PR

1. Unit tests must pass.
2. Review.
3. Verify that no queries are sent when network is set to off.


t/test01.t
```diff
@@ -190,16 +203,14 @@ subtest 'API calls' => sub {
...
         ok( grep( /2001:678:4c::1/, @ip_list ), 'Has "2001:678:4c::1" ip' );
-        ok( grep( /193\.176\.144\.22/, @ip_list ), 'Has "193.176.144.22" ip' ); # e.ext.nic.fr
-        ok( grep( /2a00:d78:0:102:193:176:144:22/, @ip_list ), 'Has "2a00:d78:0:102:193:176:144:22" ip' );
         ok( grep( /194\.146\.106\.46/, @ip_list ), 'Has "194.146.106.46" ip' ); # f.ext.nic.fr
         ok( grep( /2001:67c:1010:11::53/, @ip_list ), 'Has "2001:67c:1010:11::53" ip' );
 
         my $ds_value = {
             'algorithm' => 13,
-            'digest' => '1b3386864d30ccc8f4541b985bf2ca320e4f52c57c53353f6d29c9ad58a5671f',
+			'digest' => '1303e8da8fb60db500d5bea1ee5dc9a2bcc93dfe2fc43d346576658feccf5749', # must match case
```
marc-vanderwal, line 211: Best to keep the indentation as it was, I think (the rest is indented with spaces, here it’s being replaced with tabs):
```suggestion
            'digest' => '1303e8da8fb60db500d5bea1ee5dc9a2bcc93dfe2fc43d346576658feccf5749', # must match case
```

marc-vanderwal (commented):
Looks fine, except for one teeny tiny little thing.

t/test01.t
matsduf: My emacs wants to have it more indented.

marc-vanderwal (approved):

matsduf:
> After investigating, I would suggest an alternative fix:

I suggest that you create a new PR instead.

mattias-p:
It's a bit awkward, but it looks right to me. Not saying it's your fault that it's awkward.

PR #1194, opened 2024-12-04
Add missing files to manifest

tgreenx:
## Purpose

This PR adds missing files to manifest which were inadvertently not added by #1192.

## Context

#1192

## How to test this PR

Unit tests should pass, specifically `t/manifest.t`.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -87,6 +87,8 @@ t/rpc_validation.t
 t/test01.data
 t/test01.t
 t/test_profile.json
+t/test_profile_network_true.json
+t/test_profile_no_network.json
 t/test_validate_syntax.t
 t/TestUtil.pm
 t/validator.t


matsduf (approved):
Travis did not catch that error. `t/manifest.t` works on the command line. As far as I can see Travis has not run `t/manifest.t`. Is that because it is in MANIFEST.SKIP and not in MANIFEST? Do Travis and Github actions handle this differently?

mattias-p (approved):

mattias-p:
> Travis did not catch that error. `t/manifest.t` works on the command line. As far as I can see Travis has not run `t/manifest.t`. Is that because it is in MANIFEST.SKIP and not in MANIFEST?

I did have a look just now at what Travis is doing, and yes, Travis is skipping tests that are covered by MANIFEST.SKIP. Have a look at the log from https://app.travis-ci.com/github/zonemaster/zonemaster-backend/jobs/628723592#L1285 down to line 1322. I believe this is the directory where `make test` is executed. AFAICT this is Travis's standard way of dealing with Perl.

Issue #953, opened 2022-02-15
Dockerization for Backend


mattias-p:
We should have a Dockerfile for Backend. It's been suggested that we have a single Dockerfile for all of Backend, with support for running either zm-rpcapi or zm-testagent using different entry points. At the time of writing it is unclear to me how the DBMS, Backend configuration and Engine profile should be deployed. It would be nice if these could be user configurable.

matsduf:
If it is a single dockerfile then I assume that both RPC-API and Testagent will be running in that container. Possibly with SQLite with the database engine to make it light weight?

mattias-p:
I believe the idea is to run the services in separate containers based on the same image. If it's a good idea to run multiple services in the same container we could add that as an additional feature.

matsduf:
I guess it depends on what use case that should be supported. If testing backend on the laptop is the use case, then a single container is probably cheaper. If running backend for production is the use case, then running separate containers is probably better.

PR #1225, opened 2025-07-21
Dockerfile for backend

MichaelTimbert:
## Purpose

This PR propose a Dockerfile for backend.
It capable to launch 'rpcapi' or 'testagent' or both from docker run argument.
'zmb' and 'zmtest' are also available from docker run argument.
It is also capable to launch zonemaster-cli, it is needed for building an All In On docker image able to launch both CLI or GUI.
This docker image use S6 Overlay to manage multiple process running at the same time in docker.

It's possible to launch only 'rpcapi' or 'testagent' alone. It's for future use, and is not useful at the moment.

## Usage

For testing zonemaster-cli
```
$ docker run -ti --rm zonemaster/backend:local cli test.fr
```
All argument after 'cli' are forwarded to 'zonemaster-cli'

For testing zonemaster backend
```
$ docker run --rm -p 5000:5000 --name zm -d zonemaster/backend:local full
```
Container can be stopped with `docker stop zm`

You can interact with the backend with 'zmb' or 'zmtest' from the host if they are installed. Otherwise, you can use them from within Docker itself.

```
docker run -ti --rm --net host zonemaster/backend:local zmtest test.fr
testid: ed38765834e45b6e
1% done
```

```
docker run -ti --rm --net host zonemaster/backend:local zmb start_domain_test --domain test.fr
{"id":1,"result":"ed38765834e45b6e","jsonrpc":"2.0"}
docker run -ti --rm --net host zonemaster/backend:local zmb get_test_results --test-id ed38765834e45b6e --lang en
```

Note: The "--net host" option is required, otherwise the container cannot access the backend on port 5000.

## Context

Proposition for #953

## Changes

- Change Makefile.PL to add 'docker-build' target
- Add a Dockerfile based on zonemaster-cli image
- Add 'zonemaster_launch'  script for the docker entrypoint

## How to test this PR

To be able to test this PR, you need to build zonemaster-cli docker image first.
You must be able to build 'zonemaster/backend:local' image by executing 
```
perl Makefile.PL
make all dist docker-build
```

### Testing zonemaster-cli

You must be able to test zonemaster-cli with `docker run -ti --rm zonemaster/backend:local cli [options]`. Every option must be passed to zonemaster-cli without exception.

### Testing backend

launch the backend with the  following cmd:
```
$ docker run --rm -p 5000:5000 --name zm -d zonemaster/backend:local full
```
You must be able to test some domain using `zmb` from the host.
To stop the backend run:
```
docker stop zm
```

### DIFF
--- a/zonemaster_launch
+++ b/zonemaster_launch
@@ -34,9 +34,11 @@ case $1 in
     echo "'$1' is not a valid option.
     Available options:
 	- cli       : pass argument to zonemaster-cli then quit
-	- full      : launch both rpcapi & testagent
+	- full      : start both rpcapi & testagent
 	- rpcapi
 	- testagent
+	- zmb
+	- zmtest
     "
     ;;
 esac;


matsduf:
> You must be able to test some domain using `zmb` from the host.

Can `zmb` be run with

```
$ docker run -ti --rm zonemaster/backend:local zmb test.fr
```

MichaelTimbert:
> With this we have two docker images with `zonemaster-cli` (which is fine). Since this one inherits the other can it then be assumed that they two will have the same behavior?

not exactly, we need to add `cli` argument before any `zonemaster-cli` arguments:
```
docker run -ti --rm zonemaster/cli [zonemaster-cli options] test.fr
vs
docker run -ti --rm zonemaster/backend:local cli [zonemaster-cli options] test.fr
```

I don't think this is a problem if the end goal is to have one image to run all Zonemaster programs, like:
```
docker run -ti --rm zonemaster cli ...
docker run -ti --rm zonemaster zmtest ...
docker run -ti --rm zonemaster full ...
```

MichaelTimbert:
> > You must be able to test some domain using `zmb` from the host.
> 
> Can `zmb` be run with
> 
> ```
> $ docker run -ti --rm zonemaster/backend:local zmb test.fr
> ```

We can do it by reusing the same container : 
```
docker run -ti --rm --name zm  zonemaster/backend:local full
docker exec zm zmb start_domain_test --domain test.fr
docker exec zm zmb get_test_results --test-id 8cea2bea98449ea9 --lang en
```

or using another container, but in this case the container must use the `--net host` option to work.
```
docker run -ti --rm --name zm -p 5000:5000 zonemaster/backend:local full
docker run -ti --rm --net host --entrypoint zmb zonemaster/backend:local start_domain_test --domain test.fr
docker run -ti --rm --net host --entrypoint zmb zonemaster/backend:local get_test_results --test-id b1dac7ceb8b95f33 --lang en
```
If we want, we can add "zmb" and "zmtest" to the zonemaster_launch script to be able to run it, like this:
```
docker run -ti --rm --name zm -p 5000:5000 zonemaster/backend:local full
docker run -ti --rm --net host zonemaster/backend:local zmb start_domain_test --domain test.fr
docker run -ti --rm --net host zonemaster/backend:local zmtest test.fr
```

matsduf:
> > With this we have two docker images with `zonemaster-cli` (which is fine). Since this one inherits the other can it then be assumed that they two will have the same behavior?
> 
> not exactly, we need to add `cli` argument before any `zonemaster-cli` arguments:

Yes, but that is close enough to me. I should have clarified that in my question.

matsduf:
> I don't think this is a problem if the end goal is to have one image to run all Zonemaster programs, like:

I think that we should still keep the zonemaster-cli docker image for those that just want to run that. The extra argument for CLI when using the "full" image is not a problem, I think.

MichaelTimbert:
> > For testing zonemaster backend
> > ```
> > $ docker run --rm -p 5000:5000 --name zm -d zonemaster/backend:local full
> > ```
> 
> As I understand it, you can choose another ports, but what do the two "5000" mean? Some port translation?
> 
> > ```
> > docker run -ti --rm --net host zonemaster/backend:local zmb start_domain_test --domain test.fr
> > {"id":1,"result":"ed38765834e45b6e","jsonrpc":"2.0"}
> > docker run -ti --rm --net host zonemaster/backend:local zmb get_test_results --test-id ed38765834e45b6e --lang en
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note: The "--net host" option is required, otherwise the container cannot access the backend on port 5000.
> 
> If you choose another port for RPCAPI, e.g. 6000, can you specify that to `zmb` with `--server http://localhost:6000/`?

Yes `-p 5000:5000` mean some port translation: port 5000 of localhost is forwarded to port 5000 of the container.

You can change the port, for example  `-p 2304:5000` and use `--server http://localhost:2304` option on zmb.

MichaelTimbert:
> > To be able to test this PR, you need to build zonemaster-cli docker image first. You must be able to build 'zonemaster/backend:local' image by executing `make docker-build`
> 
> Just running `make docker-build` did not work. But the following seems to work:
> 
> ```
> perl Makefile.PL
> make all dist docker-build
> ```

Thanks, I've edited it.

zonemaster_launch
```diff
@@ -0,0 +1,42 @@
...
+        exec /init
+    ;;
+  *)
+    echo "'$1' is not a valid option.
+    Available options:
+	- cli       : pass argument to zonemaster-cli then quit
+	- full      : launch both rpcapi & testagent
+	- rpcapi
+	- testagent
+    "
```
tgreenx: Aren't these options missing?
```suggestion
    echo "'$1' is not a valid option.
    Available options:
	- cli       : pass argument to zonemaster-cli then quit
	- full      : start both rpcapi & testagent
	- rpcapi
	- testagent
	- zmb
	- zmtest
    "
```

Issue #1229, opened 2025-11-26
Build dist file in CI


mattias-p:
We should build the dist file in CI and save it as an artifact attached to the CI job.

It has been noted that this would be useful:
 * during release testing, and
 * during PR review.

Open questions:
 * When this should be triggered?
 * How and when are these artifacts cleaned up?

Follow-up:
 * [ ] Create issues/PRs to do this in the other repositories too.

PR #1235, opened 2026-02-24
Add artifact build to the CI

MichaelTimbert:
## Purpose

This PR adds artifact creation to CI. This artifact can be useful for release testing and PR review.
The artifact is created only if all tests pass.

## Context

Follow-up to https://github.com/zonemaster/zonemaster-backend/issues/1229

## How to test this PR

You should be able to download the artifact from the action tab and install it by following the Zonemaster documentation.
https://doc.zonemaster.net/latest/installation/README.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,7 +144,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: apt install
-        run: sudo apt-get install -y build-essential git libmodule-install-perl
+        run: sudo apt-get install -y build-essential git libmodule-install-perl gettext
 
       - name: build
         run: perl Makefile.PL && make all dist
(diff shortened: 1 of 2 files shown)


matsduf (changes requested):
The `Zonemaster-Backend.mo` files are missing from the TAR file. I think the MANIFEST unit test should captured that.

matsduf (approved):
Gives the same content as built on the command line.

matsduf (changes requested):
Can you add the same documentation as the other PRs?

I assumed this to have the v2026.1 milestone. Is that correct?

matsduf (approved):

matsduf:
I downloaded the zip file (artifact) produced when #1248 was merged, which is at commit 5980bb3, which is where develop branch points at. I took the tar file created locally on FreeBSD from the same commit. I compared the two and only found insignificant differences due to different version of inc::Module::Install being used (1.19 vs 1.21). Only the version identifier differed. This was done as release testing for v2026.1.

PR #1242, opened 2026-05-20
Docker related improvement.

MichaelTimbert:
## Purpose

Move files related to docker inside a dedicated docker directory.
Improve the Docker entrypoint script (`docker/zonemaster_launch`) with a cleaner help message.

## Changes

- **Add `exit 1` for invalid input** — the script now returns a non-zero exit code when an unknown command is provided, instead of silently exiting with 0
- **Improve help message formatting** — aligned descriptions for all options and added missing descriptions for `rpcapi`, `testagent`, `zmb`, and `zmtest`

## How to test this PR

`make docker-build` must work and be able to run docker container.


docker/zonemaster_launch
```diff
@@ -0,0 +1,67 @@
...
+# It acts as a command dispatcher that starts different Zonemaster components based
+# on the first argument passed to the container.
+#
+# How it works:
+# - The script reads the first argument ($1) and uses a case statement to determine
+#   which component to launch.
+# - For CLI tools (cli, zmb, zmtest), it passes any additional arguments to the
+#   respective command and exits after completion.
+# - For services (rpcapi, testagent), it starts the process in the foreground,
+#   keeping the container running.
```
matsduf, line 15: In the foreground? When I run `docker run --rm -p 5000:5000 --name zm -d zonemaster/backend full` it goes into the background as far as I can see.

matsduf (commented):
Besides my question it looks good.

docker/zonemaster_launch
marc-vanderwal: The process is running in the foreground from this shell script’s point of view. The `-d` simply sends the container started by `docker run` into the background from outside the container.

marc-vanderwal (dismissed):

matsduf (dismissed):

docker/zonemaster_launch
```diff
@@ -0,0 +1,67 @@
...
+	zmtest $@
+    ;;
+
+  rpcapi)
+	/usr/local/bin/starman --listen=0.0.0.0:5000 --preload-app --user=zonemaster --group=zonemaster /usr/local/bin/zonemaster_backend_rpcapi.psgi
+
+    ;;
+
+  testagent)
+	/usr/local/bin/zonemaster_backend_testagent -user=zonemaster --group=zonemaster foreground
```
marc-vanderwal: Usually, applications in containers should log to stdout/stderr instead of a file:
```suggestion
	/usr/local/bin/zonemaster_backend_testagent --user=zonemaster --group=zonemaster --logfile=- foreground
```

docker/zonemaster_launch
matsduf: Why doesn't it start the daemons in the background instead?

marc-vanderwal: Inside the container, daemons should be in the foreground, so that the container stops if the daemon inside it exits or crashes. That’s usually how daemons are meant to run in containers.

matsduf (approved):

marc-vanderwal (approved):

Issue #960, opened 2022-03-07
Formal states for test items


mattias-p:
At any given time a test item is in exactly one of several states. Today these states don't have formal names and it is a non-trivial operation to determine what state a test item is in. This issue describes a way to improve this situation.

Here is the list of possible states and the current criteria for each of them:
* `waiting` &mdash; progress == 0
* `running` &mdash; 0 < progress < 100
* `completed` &mdash; progress == 100 **and** test_results *does not contain* { "tag": "UNABLE_TO_FINISH_TEST", ... }
* `cancelled` &mdash; progress == 100 **and** test_results *contains* { "tag": "UNABLE_TO_FINISH_TEST", ... } **and** test_end_time ≈ test_start_time + max_zonemaster_execution_time
* `crashed` &mdash; progress == 100 **and** test_results *contains* { "tag": "UNABLE_TO_FINISH_TEST", ... } **and** logfile *contains* "Test died: {test_id}"

The procedure `job_status` should return an object with the following keys:
```
{
    "state": state,          // in all states
    "created_at": timestamp, // in all states
    "started_at": timestamp, // only in states "running", "completed", "cancelled" and "crashed"
    "ended_at": timestamp,   // only in states "completed", "cancelled" and "crashed"
    "progress": int          // only in states "running", "completed", "cancelled" and "crashed"
}
```

The `lifecycle.t` test file should be updated to:
* Cover all state transitions.
* When the intention is the check the state of a test, check the state name instead of proxies like the progress field.
* Check the timestamps and progress where it makes sense.

## Alternatives
* In an earlier version the `waiting` state was called `created`. A few arguments have been made against `created` and in favor of `waiting`.
  * All tests could be said to be created, even the completed ones.
  * The term `waiting` makes more sense than `created` should the test agent's dispatcher ever want to reclaim a test from an worker, e.g. because of inactivity.
* In an MVP implementation we could let the `completed` state also cover the `cancelled` and `crashed` states above, and the API could be extended to return the distinct `cancelled` and `crashed` as a followup. If we declare in the RPCAPI documentation that the set of reported states will be extended in the future, the followup would constitute an minor version bump and not a major version bump. This limited feature would be simpler to implement as it can be implemented efficiently without updating the database schema. In fact we already have the Zonemaster::Backend::DB::test_state() method detecting the required states.

## Stabilization
[Tracking issue](https://github.com/zonemaster/zonemaster/issues/1175)

### Affected interfaces
[...]

PR #1246, opened 2026-06-08
Formal states in tables

MichaelTimbert:
## Purpose

Introduce a formal state column to the test_results table, replacing the implicit state inference from the progress integer with explicit, well-defined lifecycle states.

## Context

https://github.com/zonemaster/zonemaster-backend/issues/960

Previously, the state of a test was derived from the progress value (0 = waiting, 1-99 = running, 100 = completed). This made it impossible to distinguish between different terminal states (e.g., completed normally vs. cancelled vs. crashed). This PR adds a proper state column with five formal states.

## Changes

- Database schema (MySQL, PostgreSQL, SQLite): Added state VARCHAR(20) NOT NULL DEFAULT "waiting" column with a CHECK constraint limiting values to waiting, running, completed, cancelled, crashed.

- State constants (DB.pm): Lowercased existing constants (WAITING → waiting, RUNNING → running, COMPLETED → completed) and added two new ones: $TEST_CANCELLED and $TEST_CRASHED.

## How to test this PR
- Run the existing test suite to verify all state transitions work correctly.
- Verify database schema creation includes the new state column and CHECK constraint on all three backends (MySQL, PostgreSQL, SQLite).
- Confirm that cancelled and crashed tests are properly distinguished in the state column.

### DIFF
--- a/script/zmb
+++ b/script/zmb
@@ -314,6 +314,32 @@ sub cmd_get_test_params {
     );
 }
 
+=head2 job_status
+
+ zmb [GLOBAL OPTIONS] job_status [OPTIONS]
+
+ Options:
+   --job-id JOB_ID
+
+=cut
+
+sub cmd_job_status {
+    my @opts = @_;
+
+    my $opt_job_id;
+    GetOptionsFromArray(    #
+        \@opts,
+        'job-id|t=s' => \$opt_job_id,
+    ) or pod2usage( 2 );
+
+    return to_jsonrpc(
+        id     => 1,
+        method => 'job_status',
+        params => {
+            job_id => $opt_job_id,
+        },
+    );
+}
 
 =head2 get_test_results
 
--- a/Changes
+++ b/Changes
@@ -1,5 +1,16 @@
 Release history for Zonemaster component Zonemaster-Backend
 
+v12.2.0 YYYY-MM-DD (part of Zonemaster v2026.2 release)
+
+ [Release information]
+ - Migration of the database is required by this release to add
+   the test state column. Follow the instructions in the
+   Zonemaster-Backend upgrade guide to upgrade to version 12.2.0.
+
+ [Features]
+ - Adds formal test states (waiting, running, completed, cancelled,
+   crashed) to the database (#1246)
+
 
 v12.1.1 2026-07-22 (part of Zonemaster v2026.1.1 release)
 
--- a/MANIFEST
+++ b/MANIFEST
@@ -54,6 +54,7 @@ share/locale/sl/LC_MESSAGES/Zonemaster-Backend.mo
 share/locale/sv/LC_MESSAGES/Zonemaster-Backend.mo
 share/Makefile
 share/patch/patch_db_schema_version_1.pl
+share/patch/patch_db_schema_version_2.pl
 share/patch/patch_db_zonemaster_backend_ver_11.1.0.pl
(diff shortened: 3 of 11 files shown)


matsduf:
I suggest that there is a document that defines the states outside the code. It would not be a public facing document and I think it can go into the `docs` directory in this repository.

matsduf:
I guess that an alternative is to have it well-defined in a POD section in the `DB.pm` module.

script/zmb
```diff
@@ -314,6 +314,32 @@ sub cmd_get_test_params {
     );
 }
 
+=head2 job_status
+
+ zmb [GLOBAL OPTIONS] job_status [OPTIONS]
+
+ Options:
+   --job-id JOB_ID
```
matsduf, line 322: Is this the same ID that is called "test-id" for other methods?

MichaelTimbert: yes, "job_id" is the new "test_id" based on https://github.com/zonemaster/zonemaster-backend/issues/1084

matsduf: I think it will be confusing to have different terms for the same thing.

matsduf:
If the new method has a new name "job-id" for what is called "test-id" in other methods, then we rename it for all methods. Do we really gain anything or enough from that?

script/zmb
MichaelTimbert: I agree, it's confusing. I'm going back to test-id.

matsduf:
Do you have an update of https://doc.zonemaster.net/latest/using/backend/rpcapi-reference.html too?

MichaelTimbert:
> Do you have an update of https://doc.zonemaster.net/latest/using/backend/rpcapi-reference.html too?

I don't think an update of rpcapi-reference.html is needed for this PR. The job_status method is listed as an experimental API method, and the experimental methods are only documented by name (no detailed description of parameters or return values). Since the changes I made only affect this experimental method, there is nothing to update in that document.

script/zmb
MichaelTimbert: Regarding the job_id vs test_id naming: I noticed that job_id was already used in the experimental job_status method before this PR, so the inconsistency existed before my changes. 
I think this naming question would be better addressed as part of a broader RPC API stabilization effort. I'm planning to write a proposal for this (see zonemaster/zonemaster#1175), and I'd prefer to postpone the naming decision until then, so we can have a consistent approach across all methods.

matsduf:
> I don't think an update of rpcapi-reference.html is needed for this PR. The job_status method is listed as an experimental API method, and the experimental methods are only documented by name (no detailed description of parameters or return values). Since the changes I made only affect this experimental method, there is nothing to update in that document.

Yes, the `job_status` method is listed today as experimental and it is not really documented, but today it does not add anything and it is not used. The updated method will add information and can be used, if documented.

* Why should it still be experimental? That method will be the only window to the formal states.
* Even if it is experimental is should still be documented. If it is not documented it can not be used in any useful way.
* The method is implemented in both the RPCAPI and in zmb. Without a documentation we cannot know how it should be implemented.

matsduf:
Today there are among others the following methods
* start_domain_test
* test_progress
* get_test_results

Unless those are renamed it would be confusing to introduce a method `job_status`. It would be better to call it `test_status`.

(I know that `job_status` formally exists, but it is redundant and it is never used today.)

PR #1261, opened 2026-09-16
Preset profiles configuration in Backend

tgreenx:
## Purpose

This PR proposes a script used to configure (expose) preset profiles in Backend's configuration file. The script:

```
   1. Changes its working directory to the directory supplied as its argument.
   2. Finds all .profile files in Zonemaster-Engine's shared preset profiles
      directory, using the PROFILE_DIR environment variable to locate it.
   3. Creates a key=value entry for each profile, using the filename without
      the ".profile" extension as the key.
   4. Prompts before changing an already existing profile.
   5. Updates or adds new profiles to the [PUBLIC PROFILES] section of
      Zonemaster-Backend's configuration file.
```

## Context

Documentation: https://github.com/zonemaster/zonemaster/pull/1538
First preset profile definition: https://github.com/zonemaster/zonemaster-engine/pull/1552

## Changes

- Adds `script/configure-preset-profiles.sh`

## How to test this PR

Note: I've tested it on a Debian 12 machine and it worked fine. It should be compatible (and tested) with BSD systems too.

Testing requires https://github.com/zonemaster/zonemaster-engine/pull/1552 (well not strictly, any other preset profile in the `share/profiles` directory of Zonemaster-Engine would do).
The following command can then be used to configure the preset profile(s) in Backend's configuration file (located in `/etc/zonemaster/backend_config.ini` in this example):

`sudo env PROFILE_DIR="$(perl -MFile::ShareDir=dist_dir -E 'say dist_dir("Zonemaster-Engine")')/profiles" ./script/configure-preset-profiles.sh /etc/zonemaster`

After that, the profile should have been added in the [PUBLIC SECTION] of the Backend configuration file (e.g. `/etc/zonemaster/backend_config.ini`):

```
$ cat /etc/zonemaster/backend_config.ini | grep -F ".profile"
kindns=/usr/local/share/perl5/auto/share/dist/Zonemaster-Engine/profiles/kindns.profile
```


script/configure-preset-profiles.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
```
matsduf, line 1: Bash is not installed in FreeBSD by default. What do you need from Bash that you cannot find in Bourne shell (sh)?

matsduf (changes requested):

script/configure-preset-profiles.sh
```diff
@@ -0,0 +1,141 @@
...
+        for (key in newentry)
+            if (!found[key] && decision[key] == "new")
+                print newentry[key]
+}
+' "$CONFIG" > "$tmp_config"
+
+# Replace the configuration file while preserving its existing permissions and ownership.
+install -p "$CONFIG" "$CONFIG.tmp"
+cat "$tmp_config" > "$CONFIG.tmp"
+mv "$CONFIG.tmp" "$CONFIG"
```
marc-vanderwal, line 141: I have mixed feeling about clobbering configuration files. IMHO, the best approach is to write a configuration file in another file and print the path on the terminal. That lets me `diff` the two files before deciding to overwrite the configuration file.

